PR #84, opened 2015-07-30
case-insensitive fuzzy autocomplete

bjarnagin:
Fuzzy autocomplete was not working for objects with capitalized names. Converting the collection item (e.g. db name or table name) to lower case allows case-insensitive regex matching.


amjith:
That was a nice catch. Thank you for the PR. Very much appreciated. :)

:tractor:

bjarnagin:
Thank YOU for building a great CLI!

PR #85, opened 2015-07-30
Fix README link to prompt_toolkit

adamchainz:
Looked like a little copypasta.


amjith:
Thank you for the fix. :)

:candy:

PR #92, opened 2015-07-31
upgrade prompt_toolkit

mattn:
If installed pgcli, mycli doesn't work because the version of prompt_toolkit is older than pgcli's one.


amjith:
Thank you for the version update. I didn't realize mycli was falling behind. 

Appreciate the proactive PR.

:rotating_light:

Issue #95, opened 2015-07-31
Please support --defaults-group-suffix


derobert:
With the stock MySQL command line client, you can add `--defaults-group-suffix=SOMETHING` to look for the `[mysqlSOMETHING]` section in `my.cnf`. This lets you store settings for, e.g., multiple database servers.

It'd be nice if mycli had this option as well.

amjith:
Yes. This is a good request. I'll add this in.

j-bennet:
@amjith: If a setting is specified in both .pgclirc and in a suffixed group of my.cnf file (such as "prompt"), which one should take priority? I'm leaning towards .pgclirc, but not sure.

j-bennet:
Actually there's another consideration here: since we have a default .myclirc, even if user erases the setting from his "home" version of .myclirc, we're going to read it from default config. If I prioritize it over "suffixed group" setting, as was my initial thought, it may bee to much pain for a user to force mycli into using the "suffixed group" setting. He'd have to erase the settings from both "home config" and "default config", and only then "suffixed group" setting would be used. Hmm. :-/

amjith:
Always give higher precedence to my.cnf.

amjith:
This is now available in master. You can try it out by installing it directly from master as follows: 

```
pip install https://github.com/dbcli/mycli/archive/master.zip
```

amjith:
This is now available in the latest version of mycli (1.3.0). Please upgrade.

derobert:
Thank you. Confirmed working here.

PR #116, opened 2015-08-04
Defaults group suffix

j-bennet:
This addresses issue #95.

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -64,14 +64,18 @@ class MyCli(object):
         '~/.my.cnf'
     ]
 
-    def __init__(self, force_passwd_prompt=False, sqlexecute=None, prompt=None, logfile=None):
+    defaults_suffix = None
 
+    def __init__(self, force_passwd_prompt=False, sqlexecute=None, prompt=None, logfile=None, defaults_suffix=None):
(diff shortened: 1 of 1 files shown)


mycli/main.py
```diff
@@ -166,11 +170,16 @@ def read_my_cnf_files(self, files, keys):
...
+
         def get(key):
-            try:
-                return cnf['client'][key]
-            except KeyError:
-                return None
+            result = None
+            for sect in sections:
+                if sect in cnf and key in cnf[sect]:
+                    result = cnf[sect][key]
```
amjith: Should this have an early return inside the for loop so it returns as soon as the key is found?

amjith:
This works as expected. I'm going to merge it in. 

Thank you for the PR. 

:money_with_wings:

mycli/main.py
j-bennet: Only just saw the email. No, the values from files lower on the list should overrides previous values.

j-bennet: Only just saw the email. No, the values from files lower on the list should overrides previous values.

derobert:
Thank you. Confirmed working here.

PR #117, opened 2015-08-05
Add change table format as a special command

brewneaux:


mycli/main.py
```diff
@@ -105,7 +104,19 @@ def register_special_commands(self):
                 aliases=('\\r', ))
         special.register_special_command(self.refresh_dynamic_completions, 'rehash',
                 '\\#', 'Refresh auto-completions.', arg_type=NO_QUERY, aliases=('\\#',))
-
+        special.register_special_command(self.change_table_format, 'tableformat', 
+                '\\T', 'Change Table Type.', aliases=('\\T',))
```
amjith: You need to add an extra parameter here called `case_sensitive=True`. Because as it stands right now this is overriding `\t` which turns on/off timing information. 

Also the third parameter is intended as a documentation string for the shortcut, so it should read `\\T [format_name]`.

mycli/packages/tabulate.py
```diff
@@ -648,6 +648,8 @@ def _normalize_tabular_data(tabular_data, headers):
 
     return rows, headers
 
+def get_allowed_table_formats():
```
amjith: I would have named this `table_formats()`.

amjith: I would have named this `table_formats()`.

amjith:
The change looks good. I like how you added auto-completion for it. :)

I've left some comments in the PR. But I'm going to merge this PR and make those changes myself. They're there for future considerations.

Thank you for taking the time to implement this feature. 

Cheers!

Issue #122, opened 2015-08-06
BINARY fields throw decode error


kived:
We have many BINARY(16) columns in our database. When trying to select from these tables, mycli encounters a decode error:
````
'ascii' codec can't decode byte 0xc0 in position 2: ordinal not in range(128)
````

The best approach would be to automatically `HEX()` any BINARY fields for display (like HeidiSQL does). And that would be just one more thing that mycli would have over the official mysql client.

amjith:
Yes. There is already work underway to do exactly this. #106 

Thanks to @danieljwest.

danieljwest:
Good to know hex was the right way to go!

mdsrosa:
I'm closing this issue since it has already been resolved by #132. :wink:

PR #123, opened 2015-08-07
Added encoding adaptor to lazy handle binary fields.

### LINKED ISSUE #106: sql statement contains "*" that has mysql spatial data format , i.e.  linestring,get below error.
2015-08-03 15:08:48,736 (9676/MainThread) mycli.main ERROR - traceback: 'Traceback (most recent call last):\n  File "/usr/share/python/mycli/local/lib/python2.7/site-packages/mycli/main.py", line 357, in run_cli\n    for title, cur, headers, status in res:\n  File "/usr/share/python/mycli/local/lib/python2.7/site-packages/mycli/sqlexecute.py", line 108, in run\n    yield self.execute_normal_sql(sql)\n  File "/usr/share/python/mycli/local/lib/python2.7/site-packages/mycli/sqlexecute.py", line 113, in execute_normal_sql\n    num_rows = cur.execute(split_sql)\n  File "/usr/share/python/mycli/local/lib/python2.7/site-packages/pymysql/cursors.py", line 134, in execute\n    result = self._query(query)\n  File "/usr/share/python/mycli/local/lib/python2.7/site-packages/pymysql/cursors.py", line 282, in _query\n    conn.query(q)\n  File "/usr/share/python/mycli/local/lib/python2.7/site-packages/pymysql/connections.py", line 768, in query\n    self._affected_rows = self._read_query_result(unbuffered=unbuffered)\n  File "/usr/share/python/mycli/local/lib/python2.7/site-packages/pymysql/connections.py", line 929, in _read_query_result\n    result.read()\n  File "/usr/share/python/mycli/local/lib/python2.7/site-packages/pymysql/connections.py", line 1132, in read\n    self._read_result_packet(first_packet)\n  File "/usr/share/python/mycli/local/lib/python2.7/site-packages/pymysql/connections.py", line 1182, in _read_result_packet\n    self._get_descriptions()\n  File "/usr/share/python/mycli/local/lib/python2.7/site-packages/pymysql/connections.py", line 1269, in _get_descriptions\n    assert eof_packet.is_eof_packet(), \'Protocol error, expecting EOF\'\nAssertionError: Protocol error, expecting EOF\n'

danieljwest:
So this has gotten a bit uglier. I thought in my original pass through that pymysql was passing down the charset or at least whether or not the field was binary... it's not. It is passing the field_type so we might be able to make something resembling an educated guess, but that's all I got.

Right now this code will take the text types and attempt to decode them as utf-8. Failing that it'll return the result as hex. I've tried a number of different datatypes and overall it works. Testing it with content above utf-8 it seems to work just as well as mysql's client at least. It just renders a blank spot.  Either way here's some code to discuss.

I've also opened an issue with pymysql about certain binary types causing issues.
Resolves #106 and #122

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -54,6 +54,7 @@ PACKAGE_ROOT = os.path.dirname(__file__)
 class MyCli(object):
 
     default_prompt = '\\t \\u@\\h:\\d> '
+    defaults_suffix = None
 
     # In order of being loaded. Files lower in list override earlier ones.
     cnf_files = [
@@ -63,11 +64,11 @@ class MyCli(object):
         '~/.my.cnf'
     ]
 
-    def __init__(self, force_passwd_prompt=False, sqlexecute=None, prompt=None, logfile=None):
-
+    def __init__(self, force_passwd_prompt=False, sqlexecute=None, prompt=None, logfile=None, defaults_suffix=None):
         self.force_passwd_prompt = force_passwd_prompt
         self.sqlexecute = sqlexecute
         self.logfile = logfile
+        self.defaults_suffix = defaults_suffix
 
         default_config = os.path.join(PACKAGE_ROOT, 'myclirc')
         write_default_config(default_config, '~/.myclirc')
@@ -80,11 +81,14 @@ class MyCli(object):
         self.table_format = c['main']['table_format']
         self.syntax_style = c['main']['syntax_style']
         self.wider_completion_menu = c['main'].as_bool('wider_completion_menu')
-        self.prompt_format = prompt or c['main']['prompt'] or \
-                             self.default_prompt
 
         self.logger = logging.getLogger(__name__)
         self.initialize_logging()
+
+        prompt_cnf = self.read_my_cnf_files(
+            self.cnf_files, ['prompt'])['prompt']
+        self.prompt_format = prompt or prompt_cnf or c['main']['prompt'] or \
+                             self.default_prompt
 
         self.query_history = []
 
@@ -177,11 +181,16 @@ class MyCli(object):
                 cnf.merge(e.config)
                 pass
 
+        sections = ['client']
+        if self.defaults_suffix:
+            sections.append('client{0}'.format(self.defaults_suffix))
+
         def get(key):
-            try:
-                return cnf['client'][key]
-            except KeyError:
-                return None
+            result = None
+            for sect in sections:
+                if sect in cnf and key in cnf[sect]:
+                    result = cnf[sect][key]
+            return result
 
         return dict([(x, get(x)) for x in keys])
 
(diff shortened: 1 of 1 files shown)


mycli/encodingadaptor.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def fetch(self):
+        row = self.cur.fetchone()
+        if row:
+            return [encodeField(x) for x in row]
+        else:
+            raise StopIteration
+
+
+def encodeField(field):
```
amjith: It is typical to use underscore_separated_names for function names and method names.

mycli/encodingadaptor.py
```diff
@@ -0,0 +1,29 @@
+class EncodingAdaptor:
```
amjith: New style classes are required to derive from `object`. So this should be: 

```
class EncodingAdaptor(object):
```

amjith:
I'm impressed by the use of Classes, list comprehensions and iterators. Not too shabby for a person who claimed to be a python newb. :)

I've mentioned a few minor nitpicks in the code. But I'd like to discuss the solution at a higher level. 

I feel that it might be better to do the encoding conversion right before we print the table. For example in the [`format_output()`](https://github.com/dbcli/mycli/blob/master/mycli/main.py#L581) function. The reason for delaying it is to make sure we catch all instances of results coming from various sources. For example, the special commands produce their own results (see here: https://github.com/dbcli/mycli/blob/master/mycli/packages/special/dbcommands.py#L24). 

I am a bit concerned about the performance impact of this change. Because we're encoding every field of every row, which could get expensive really quickly. I'd like to investigate the stack trace from the error and see where it is thrown. If it happens in mycli codebase I'd rather catch that exception and convert it to hex at that point. I'll do more digging.

Thank you for the PR. I'll keep this open and add more info as I find.

danieljwest:
Also apparently I failed to test it in Python 3! Sorry for that! Not used to targeting multiple versions of an interpreter.

Thanks for the feedback! Couple of things!
As for performance, the performance shouldn't be any different than when running pymysql with use_unicode=True. Unless I've read the code wrong they decode every field (except for very certain fields) using ascii or the field's charset. However, improving performance tends to be a good thing. So, without doing any prior decoding the place it will blow up is the tabulate package, but it won't just blow up on binary fields. It will blow up on anything non ascii related. This includes strings that you wouldn't want to hex. Once pymysql fixes their side, I think the optimal solution would be to capture the error at that point and hexify it there.

The placement could be better. I originally had it in the format_output section but I was having an issue with a particular test failing because it was passing a list and not a cursor. The cursor doesn't implement the next function and this was causing me all sorts of python newbie headaches. Long story short, I now realize that I could have used a generator and simplified the code quite a bit.

PR #125, opened 2015-08-08
Upgrade to prompt_toolkit 0.46

jonathanslenders:


jonathanslenders:
See also https://github.com/dbcli/pgcli/pull/328

My mycli, discuss whether or not we want to have the indentation. I see the prompt is quite large. If it's too large, it should also be possible for instance to display part of the information at the bottom of the screen, in a second toolbar.

amjith:
I'm ok with the prompt being long and still indenting the query in multi-line mode. 

While I was experimenting with different ways to shorten the prompt, I decided to use `\n` option in the prompt. 

Set the prompt as follows in `~/.myclirc` and launch mycli. 

```
prompt = "\t\n\u@\h:\d> "
```

I noticed the following behavior: 

1__10_1_python2_7_-__mycli___users_amjith_dropbox_code_python_mycli___tmux_

But in prompt_toolkit 0.45 it was showing properly right after the `>` symbol.

jonathanslenders:
Yeah. I have to think about this. Indentation works because the layout uses a ``VSplit``, which acts like a table: https://github.com/jonathanslenders/python-prompt-toolkit/blob/dd5c76f998e724f4252d4163acb5c15d74c7df91/prompt_toolkit/layout/containers.py#L134

However, that means that both on the left side and on the right side we get line wrapping, and independently. The experienced behaviour is expected, though not nice.

I will think about an easy way for this to fix, but that's for the next release.
I can update the pull request again with ``multiline=False``, to get the old layout behaviour back.

amjith:
No need to update the PR. I'll keep the multi-line for now.

amjith:
Thanks for the PR Jonathan. :)

PR #132, opened 2015-08-15
Caught and hexed certain binary fields coming from pymysql.

### LINKED ISSUE #106: sql statement contains "*" that has mysql spatial data format , i.e.  linestring,get below error.

danieljwest:
Resolves #106 and #122 

This code is dependent on the code in the following pull request getting merged in.
https://github.com/PyMySQL/PyMySQL/pull/368

Caught the exceptions in the tabulate and expand and hexxed where approriate. I also appended 0x to any hexed field for ease of inserting or updating.
Intended to just update the previous pull request, but in my attempts to clean up the history I forced it to close.


mycli/packages/expanded.py
```diff
@@ -17,7 +28,8 @@ def expanded_table(rows, headers):
     header_len += 2
 
     for row in rows:
-        row_len = max([len(_text_type(x)) for x in row])
+        row = [format_field(x) for x in row]
+        row_len = max([len(x) for x in row])
```
amjith: Can this be collapsed into a single line? 

```
row_len = max([len(format_field(x)) for x in row])
```

That way we don't keep two lists around just to find out the max length of the row.

mycli/packages/expanded.py
```diff
@@ -8,6 +9,16 @@ def get_separator(num, header_len, data_len):
...
 
+def format_field(value):
+    # Returns the field as a text type, otherwise will hexify the string
+    try:
+        if isinstance(value, bytes):
+            return _text_type(value, "ascii")
+        else:
+            return _text_type(value)
+    except UnicodeDecodeError:
+        return _text_type('0x' + (codecs.getencoder('hex_codec')(value)[0]).decode('ascii'))
```
amjith: I think we can use `binascii` to make this line shorter. This is purely aesthetics, so feel free to ignore if you don't agree. 

```
    return _text_type('0x' + binascii.hexlify(value).decode('ascii'))
```

You'll need to `import binascii` instead of `codecs`.

amjith:
Good work on shepherding PyMySQL to implement the necessary feature. 

I like the addition of `0x` prefix. Nice touch.

I'm not sure when they plan to release the next version of PyMySQL. Once that is release we can merge this and bump the minimum required version in setup.py.

mycli/packages/expanded.py
danieljwest: Ah! row is actually used directly below this code to present it to the user, not just to check the length. 
I figured I would just use row immediately before the length check and the output to hexify everything while making sure the length check. 

I could swap the list comprehension (or whatever python calls them) for a generator if you're concerned about memory usage, but I believe this would be at the cost of extra processing each time the generator is used.

mycli/packages/expanded.py
danieljwest: yeah binascii works just as well and is quite a bit prettier. I'll swap it out and run some tests to make sure it actually works the same way!

mycli/packages/expanded.py
amjith: Ah. I didn't realize you were mutating the row and assigning it back on to the same var. No need to make it into a generator.

amjith:
@danieljwest Do you know if your fix to pymysql was released? Is this ready to be merged?

danieljwest:
As far as I can tell the last release was in February. So, unfortunately, they haven't gotten this into a release yet... I'll ding them and see if they're looking to release anytime soon!

amjith:
Thank you!

danieljwest:
As of 8 days ago PyMySql has been released!

tsroten:
@danieljwest Awesome! This looks really good.

Do you want to bump the required version of PyMySql in your pull request so we can see if the Travis CI build passes?

amjith:
Nice! The tests are passing again.

@tsroten This means we either drop the dependency on 0.6.2 or keep the 0.6.2 dependency but fall back to a less ideal way of handling these errors. 

@danieljwest There was a discussion about keeping the pymysql version pinned to 0.6.2 and above. To summarize, debian packages only have pymysql 0.6.2, so the debian package for mycli had to rely on pymysql 0.6.2. So Thomas back ported one of the features from 0.6.6 to 0.6.2 so we can keep the 0.6.2 dependency. 

Full details: https://github.com/dbcli/mycli/issues/155

tsroten:
@amjith @danieljwest Not sure what we should do with this one. Restating the previously mentioned options with a little more detail:

1. We could make the testcases dependent on PyMySQL version and add better error handling for this bug. We could print a graceful error message that advises updating PyMySQL. So, Debian users would have to live with this bug until the Debian PyMySQL package is updated.
2. Overwrite the `TEXT_TYPES` constant when we import PyMySQL. This means that the fix would be "back-ported" in a sense.

Personally I don't want to drop 0.6.2 support (don't change `setup.py` PyMySQL version requirement) so as to support our Debian user-base.

What are your thoughts?

amjith:
Debian users are the worst. (Just kidding). 

I think we can keep 0.6.2 support in setup.py. 

We'll handle the error in main.py and print a specialized error message that says, `Upgrade your PyMySQL to 0.6.7. Use 'pip install -U pymysql'`. 

As soon as a new debian package is available for pymysql we'll upgrade our minimum version requirement. 

If you don't have any objections I'm going to merge this PR and then make the necessary changes to handle the errors and print custom error messages.

tsroten:
Sounds good to me :) Thanks again @danieljwest for the pull request!

amjith:
:cherries:

amjith:
@danieljwest Looks like you made someone happy. https://twitter.com/bslagter/status/665469553229570048

danieljwest:
Woohoo! At least one satisfied customer!

PR #141, opened 2015-08-28
treat DESC and EXPLAIN as DESCRIBE

spacewander:
Implement it with existed code, instead of hacking.


amjith:
Awesome!!

I appreciate that you took the time to add it to the tests as well. :)

Cheers!

:ice_cream:

PR #147, opened 2015-09-12
Disabled unicode_literals warning in click.

j-bennet:


amjith:
@j-bennet Thanks for the PR. But this is already fixed in master. 

https://github.com/dbcli/mycli/blob/master/mycli/main.py#L42

j-bennet:
Oic. My fork must be messed up. I merged the upstream but this didn't get in there.

PR #150, opened 2015-09-14
Add special command to describe

shoma:
Add a special command `\d` to describe table info with completion for
table name.

### DIFF
--- a/mycli/packages/completion_engine.py
+++ b/mycli/packages/completion_engine.py
@@ -102,7 +102,7 @@ def suggest_special(text):
     if cmd in ['\\f', '\\fs', '\\fd']:
         return [{'type': 'favoritequery'}]
 
-    if cmd in ['\\d']:
+    if cmd in ['\\dt']:
         return [
             {'type': 'table', 'schema': []},
             {'type': 'view', 'schema': []},
--- a/mycli/packages/completion_engine.py
+++ b/mycli/packages/completion_engine.py
@@ -102,6 +102,13 @@ def suggest_special(text):
     if cmd in ['\\f', '\\fs', '\\fd']:
         return [{'type': 'favoritequery'}]
 
+    if cmd in ['\\d']:
+        return [
+            {'type': 'table', 'schema': []},
+            {'type': 'view', 'schema': []},
+            {'type': 'schema'},
+        ]
+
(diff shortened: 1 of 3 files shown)


amjith:
@shoma Thanks for the PR. MySQL client already uses `\d` to set or change a delimiter. This feature is still not available in mycli, but I don't want to use `\d` for a feature that doesn't match the current behavior of MySQL client. 

I'll be happy to entertain other shortcuts if you have any suggestions. 

Also I noticed the tests are failing. It's because of a typo in the new tests you added. This line: https://github.com/dbcli/mycli/pull/150/files#diff-e02ad865ae02722472e35aa77caf2022R11 should be converted to

```
suggestions = suggest_type('\\d', '\\d ')
```

shoma:
@amjith Thanks for your comment. Sorry I fixed the test problem using `push -f`.

- dbcli/mycli has already `\dt` short cut for list tables.
- MySQL Client has `\d` command to set statement delimiter.
- PostgreSQL Client and dbcli/psql has `\d` short cut for describe tables.

unhh... `\dd` is suitable for `describe`? Do you like?

shoma:
`\dd` is easy to change verbose level, like memcached's  verbose option(`-v`, `-vv`, `-vvv`). 
So `\ddd` is `show create table`

amjith:
How about overloading `\dt` for this purpose. So `\dt` without any args will list all tables, but `\dt [tablename]` will list the columns in that table. 

What do you think? That way it preserves the old behavior and still adds new functionality. It might require a new function that checks the argument and routes the control to describe_table or list_tables appropriately. 

Would you like to take a stab at it?

shoma:
:+1: `\dt` for `List or describe tables.` is simple, no problem with compatibility.

amjith:
I'm sorry I didn't see that you had update the PR with a new commit. 

It looks great. It works as expected. One minor nitpick (which I'll fix) is that the help (\?) should say `\dt [table]` instead of just `\dt`. 

Thank you for taking the time to send me this PR.

shoma:
It's my pleasure. Thanks for accept and help change commit https://github.com/dbcli/mycli/commit/d2ad097e08b8c5e9d318968f0a6f859f03f7c07a

Issue #94, opened 2015-07-31
Feature request : Read the ~/.mylogin.cnf file at startup


rapha8l:
Hi, 
Would it be possible that mycli takes ~/.my.cnf ?
Thanks

mesirendon:
+1

amjith:
It already does. It reads in username, password, host, database, socket, port attributes from my.cnf. If you have other attributes that you'd like to be honored, can you please list them?

seagoj:
So is this documented somewhere? `mycli [servername]` doesn't seem to do it.

amjith:
@seagoj I didn't know about the mylogin.cnf file. I might be able to add that in.

Too many ways to specify options, it's hard to keep up.

tanji:
.mylogin.cnf is a windows thing. On Linux, the file used is ~/.my.cnf, as the OP said. It is conventional to use that file for user logins, so if you could add it to the list of parsed files that could be nice.

seagoj:
@tanji [.mylogin.cnf](http://dev.mysql.com/doc/refman/5.6/en/mysql-config-editor.html) is not a "Windows thing". It allows you to store connection information in an encrypted file using mysql_config_editor.

tanji:
@seagoj OK, it is MySQL 5.6 specific. If it's encrypted, how do you expect mycli to read it?

seagoj:
@tanji 
> pass in the --login-path to the mysql call that should do it.

seagoj:
Looks like pymysql doesn't support it, yet, so no such luck until they do.

mesirendon:
I have a configuration that allows me to use a specific server and DB when requiered... so my `.my.cnf` file looks like this:

```
[clientServerName]
user=root
password=MySuperSecurePassword
host=remote.server.com
database=example
[clientAnotherServer]
...
```

Now... I have a function in my `.zshrc` file that allows me to connect to any DB
```
mycnc() {
        mysql --defaults-group-suffix=$1
}
```
So... if I need to connect to `ServerName` I just type `mycnc ServerName` and that's all.

Hope this conf helps you to add an option to use the tool reading right away the .my.cnf file

P.S. Great tool. Keep it growing ;)

amjith:
@seagoj Thank you for filing the issue. 

@mesirendon There is a pending issue for supporting the `--default-group-suffix` option in mycli. Ref: #95

tsroten:
[...]

PR #152, opened 2015-09-16
Read .mylogin.cnf file

tsroten:
This addresses #94.

It adds support for reading `.mylogin.cnf`, the encrypted login path file created by `mysql_config_editor`. The code supports Windows (`%APPDATA%\MySQL\.mylogin.cnf`) and non-Windows (current user's home directory via `~/.mylogin.cnf`).

I had to make some changes to `MyCli.read_cnf_files` in order to accomplish this. Namely, I had to change the logic to support reading multiple "sections" in the cnf files (`'client'` and the user-specified login path group). Also, changed the cnf settings loop to loop through sections in the cnf object, not the list `sections`. This somewhat follows `mysql`'s behavior of treating sections that appear later in a cnf file as overriding those that appear earlier (e.g., if login path `'qa1'` is defined before `'client'`).

Most of the logic for finding and reading the `.mylogin.cnf` file is in `config.py`, which seemed like an appropriate place for it. It introduces another dependency, `pycrypto`, which has been added to `setup.py`. The decryption code might be a bit heavily commented. I did this because I find it a bit obtuse and foreign.

The command-line option `--login-path` was added. Even when this is not present, the `'client'` login path group is always read if it exists.

`.mylogin.cnf` is read even if the user specifies `--defaults-file`. This was done to match `mysql`'s behavior.

Lastly, I updated the list of arguments/options in `README.md`.

I tested this on Python 2.7 and 3.4.


tsroten:
We might want to add some error handling to the decryption code. What do you think?

mycli/main.py
```diff
@@ -207,13 +217,16 @@ def read_my_cnf_files(self, files, keys):
                 pass
 
         sections = ['client']
+        if self.login_path and self.login_path != 'client':
+            sections.append(self.login_path)
+
         if self.defaults_suffix:
-            sections.append('client{0}'.format(self.defaults_suffix))
+            sections.extend([sect + self.defaults_suffix for sect in sections])
```
amjith: Can a user specify both `login_path` and `defaults_sufix` in the command line? If they do, are we supposed to append the `defaults_suffix` to the `login_path`? That doesn't sound right to me. But I've never used login_path before so I'm merely speaking from my gut feeling.

mycli/main.py
```diff
@@ -207,13 +217,16 @@ def read_my_cnf_files(self, files, keys):
...
 
         def get(key):
             result = None
-            for sect in sections:
-                if sect in cnf and key in cnf[sect]:
+            for sect in cnf:
+                if sect in sections and key in cnf[sect]:
```
amjith: This is a good change. I like this.

mycli/main.py
```diff
@@ -111,6 +114,14 @@ def __init__(self, sqlexecute=None, prompt=None,
         # Register custom special commands.
         self.register_special_commands()
 
+        # Load .mylogin.cnf if it exists.
+        mylogin_cnf_path = get_mylogin_cnf_path()
+        if mylogin_cnf_path:
+            mylogin_cnf = get_mylogin_cnf_plaintext(mylogin_cnf_path)
```
amjith: If there is gibberish in the mylogin.cnf file how will this behave? In other words, you're right we should add some error handling to print a graceful error that says the mylogin.cnf file is corrupted or we're unable to decipher the file.

mycli/config.py
```diff
@@ -16,3 +20,60 @@ def write_default_config(source, destination, overwrite=False):
...
+        mylogin_config_dir = os.path.expanduser('~')
+    else:
+        mylogin_config_dir = os.path.join(app_data, 'MySQL')
+
+    mylogin_config_dir = os.path.abspath(mylogin_config_dir)
+    mylogin_config_path = os.path.join(mylogin_config_dir, '.mylogin.cnf')
+
+    return mylogin_config_path if exists(mylogin_config_path) else None
+
+def get_mylogin_cnf_plaintext(file_name):
```
amjith: It would be good to add a unit-test for this function.

amjith:
Thanks a lot for the quick turn around. 

The PR looks good. I haven't been able to test it since I've never used the mysql_config_editor before, so I don't have a mylogin.cnf file to test against. I'll figure out how to create one and test it out soon. 

I love the fact that you took the time to write plenty of comments to the crypto code. Comments are always welcome. :)

I'd appreciate if you could add a test for the decryption algorithm. 

I presume pycrypto uses C-extensions, which means I can't use the simple debian build script. But that can be remedied with some experimentation. 

Thanks again for the PR. :smiley:

mycli/main.py
tsroten: I agree that it seems odd, but that is how `mysql` behaves (I just tested it again to double-check). [`--defaults-group-suffix' is not file-specific](https://dev.mysql.com/doc/refman/5.0/en/option-file-options.html#option_general_defaults-group-suffix), it adds that suffix to every option group that `mysql` looks at, even a login path group.

Login path option groups are just like regular option groups with two exceptions. 1) They are encrypted on the filesystem. 2) They only permit certain options: `host`, `user`, `password`, `port` and `socket`.

mycli/main.py
tsroten: Okay, should mycli attempt to continue after printing an error message?

`mysql` isn't much help here because it doesn't even throw an error message! It just silently continues, even if it can't read `.mylogin.cnf`.

mycli/config.py
tsroten: Okay, good idea. I'll add that.

tsroten:
Glad to help :) It would be good if someone with access to Windows could test this as well. My `%APPDATA%` code should work in principle, but I haven't actually tested it.

mycli/main.py
amjith: I think it should continue. It can print an error message but ignore that file and continue. That's the current behavior with malformed `my.cnf` files.

amjith:
I've tested this in OS X. I think I still have a Windows VM, I'll see if I can check it there. But I don't think it should stop us from merging to master.

tsroten:
Okay, I've added quite a bit of error handling and logging messages. If there is a problem reading the login path file, then the user receives a simple message: `Error: Unable to read login path file.`. The program continues without the login path configuration options. The logging messages specify the specific problems that were encountered with reading and decrypting the file.

tsroten:
I'll work on the unit tests tomorrow.

amjith:
That's a lot of error handling and refactoring. Thank you!

tsroten:
Okay, unit tests are added. Let me know what you think.

amjith:
Tests look great. Thank you for taking the time to write the tests and polish this feature. 

I appreciate it very much. 

:ice_cream:

PR #158, opened 2015-09-25
Add "delete" and "truncate" as destructive commands

martijnengler:
After doing a DELETE without a WHERE (I *thought* multiline was on, but my quick glance at the bottom of the screen was wrong) I thought it might actually be nice to have DELETE and TRUNCATE give a warning.


amjith:
Looks good. Thanks @martijnengler. 

:mountain_bicyclist:

PR #159, opened 2015-09-25
Upgrade prompt_toolkit to 0.52.

amjith:
@tsroten Can you please review and merge? 

There is no hurry. I understand you're moving this weekend so no need to rush this, take your time. 

One new feature is the enabling of mouse support. You can now click on the query to move the cursor there. So if you're writing a long multi-line query you can click anywhere in the query to jump to that position.

Click and drag will select portions of the query, but they're a bit laggy.

My suggestion is to install this branch and try a lot of the features, such as keybindings (F3, F2 etc), check syntax coloring, try out the multi-line mode, prompts, different color schemes etc.

If you have access to a windows machine try it there, otherwise I'll try to get a Windows VM running today. 

/cc @jonathanslenders

### DIFF
--- a/mycli/key_bindings.py
+++ b/mycli/key_bindings.py
@@ -14,6 +14,8 @@ def mycli_bindings(get_key_bindings, set_key_bindings):
     key_binding_manager = KeyBindingManager(
             enable_open_in_editor=True,
             enable_system_bindings=True,
+            enable_search=True,
+            enable_abort_and_exit_bindings=True,
             enable_vi_mode=Condition(lambda cli: get_key_bindings() == 'vi'))
 
     @key_binding_manager.registry.add_binding(Keys.F2)
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -362,7 +362,7 @@ class MyCli(object):
                                   key_bindings_registry=key_binding_manager.registry,
                                   on_exit=AbortAction.RAISE_EXCEPTION,
                                   on_abort=AbortAction.RETRY,
-                                  mouse_support=True,
+                                  mouse_support=False,
                                   ignore_case=True)
         cli = CommandLineInterface(application=application, eventloop=create_eventloop())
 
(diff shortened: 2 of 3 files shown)


tsroten:
Nice! I like the mouse support.

One side-effect is that the entire terminal window's mouse selection is hijacked. I can't select from previous statements or results in my history at all. What do you think of that?

tsroten:
Also, I don't have access to a Windows installation, so I can't test it on Windows.

amjith:
> One side-effect is that the entire terminal window's mouse selection is hijacked. I can't select from previous statements or results in my history at all. What do you think of that?

Good catch. I'm going to disable the mouse integration until this is resolved. 

/cc @jonathanslenders

amjith:
I've disabled the mouse support. Haven't tested it on windows yet.

jonathanslenders:
There is no way to enable mouse support for only a part of the terminal. It's basically all or nothing. The terminal is not aware which parts of the output belong to which application. For the terminal app, it's basically just one I/O stream.

So, in most terminal applications, when mouse support is requested, the application still allows mouse selection, but by keeping Shift pressed. I think in the Terminal.App for mac, you have to press Option in order to send mouse events, so there it's the other way around. In windows, it doesn't make any difference. You have to right click, and select "mark" anyway.

amjith:
@jonathanslenders You're killing me. Why can't you make things just magically work? :crying_cat_face: 

I'm going to make the mouse support configurable and enable it by default. 

I'll probably also add a doc page with details on how to select things with the mouse. 

In all seriousness, mouse support is just too cool to disable. :)

jonathanslenders:
Yes, vt100 terminals are in interesting but amazing piece of history. :)

You know that we have four different protocols implemented for mouse support. There is the original mouse reporting (not sure where it's coming from), uxrvt and sgr mouse tracking. And then the Win32 protocol which --as usual-- requires kernel calls to enable/disable mouse support. :)

amjith:
@tsroten I'm closing this PR since there seems to be an issue with the latest prompt_toolkit that it not interpreting the enter key in Window. So I'm going to close this PR for now until it is resolved in the next version of prompt_toolkit.

tsroten:
@amjith Was the bug you mentioned fixed? I had trouble finding any report or change log that might indicate it was filed and/or fixed.

amjith:
It is fixed but Jonathan said he'll update the PR to utilize some of the new features of the latest version. 

I think he's a wee bit busy with some of his work stuff, if we don't hear back in a few days, we can ping him.

PR #160, opened 2015-09-25
Updated release script with a --dry-run and --confirm-steps option.

j-bennet:


release.py
```diff
@@ -4,8 +4,44 @@
...
+CONFIRM_STEPS = False
+DRY_RUN = False
+
+
+def skip_step():
+    """
+    Asks for user's response whether to run a step. Default is yes.
+    :return: boolean
+    """
+    global CONFIRM_STEPS
```
tsroten: @j-bennet I really like these two functions! Very handy.

The `global` statements aren't needed since `skip_step()` and `run_step()` are only referencing these two constants, not assigning to them.

amjith: I agree the dry run is pretty awesome. 

Since the `global` qualifier doesn't hurt anything, I'm gonna merge this in.

amjith:
:tractor:

PR #166, opened 2015-10-09
Fixes #23 -- Makes "enter" key function as the "tab" key only when the completion menu is displayed.

### LINKED ISSUE #23: confirming choice in autocomplete
When mycli suggests me some choice (for example database name), i select one of them using cursors, and then press enter. This ends query and send it to mysql, but I only confirmed my choice... Is there availabilty to configure this behaviour?

mdsrosa:
New users get frustrated when they select a SQL command, press `Enter` and the command gets executed. So this PR makes the enter key work like the tab, but only when the completion menu is visible.

Fixes #23.

### DIFF
--- a/mycli/filters.py
+++ b/mycli/filters.py
@@ -1,0 +1,12 @@
+from prompt_toolkit.filters import Filter
+
+class HasSelectedCompletion(Filter):
+    """Enable when the current buffer has a selected completion."""
+
+    def __call__(self, cli):
+        complete_state = cli.current_buffer.complete_state
+        return (complete_state is not None and
+                 complete_state.current_completion is not None)
+
+    def __repr__(self):
+        return "HasSelectedCompletion()"
--- a/mycli/key_bindings.py
+++ b/mycli/key_bindings.py
@@ -75,4 +75,19 @@ def mycli_bindings(get_key_bindings, set_key_bindings):
         else:
             event.cli.start_completion(select_first=False)
 
+
+    @key_binding_manager.registry.add_binding(Keys.ControlJ)
+    def _(event):
+        """
+        This sould only be activated if the config 'use_enter_key_as_tab'
+        is activated. The enter key's behaviour should be attributed to another
+        key, for example: C-M.
+        """
+        _logger.debug('Detected <C-J> key.')
+        b = event.cli.current_buffer
+        if b.complete_state:
+            b.complete_next()
+        else:
+            event.cli.start_completion(select_first=True)
+
     return key_binding_manager
--- a/mycli/key_bindings.py
+++ b/mycli/key_bindings.py
@@ -1,9 +1,11 @@
 import logging
 from prompt_toolkit.keys import Keys
 from prompt_toolkit.key_binding.manager import KeyBindingManager
-from prompt_toolkit.filters import Condition, HasCompletions
+from prompt_toolkit.filters import Condition
+from .filters import HasSelectedCompletion
 
 _logger = logging.getLogger(__name__)
+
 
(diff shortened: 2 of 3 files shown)


tsroten:
@mdsrosa This seems to work as expected. Good job on this. And, thanks @amjith as well.

Since we are changing an existing functionality and it's not strictly a bug, should we add an option for those users that want the "legacy" functionality or just simply prefer it as it is? I know we shouldn't add configurable options galore, so I'm wondering what you all think.

tsroten:
One downside is that, even if no entry in the autocomplete menu is selected, hitting enter twice is required to run the statement. So, in order to run `exit` you must hit enter twice, even though `exit` is the only option in the menu and I didn't tab in order to select it. Can we detect whether or not an entry in the autocomplete menu is selected?

mdsrosa:
Hi @tsroten! I'm glad you liked. Thanks!

About the configuration option, I also wanted to add that option, but while I was working on this implementation, I talked to the @amjith and we both decided to keep the settings to a minimum. Only if people start to complain, then yes we will add the configuration option.

mdsrosa:
@tsroten yes, the `hitting twice`is quite annoying! I'll check if we can detect that.

tsroten:
Ok, here we go, let's try a custom filter (the current one is `HasCompletions()`).
```python
from prompt_toolkit.filters import Filter

class HasSelectedCompletion(Filter):
    """Enable when the current buffer has a selected completion."""

    def __call__(self, cli):
        complete_state = cli.current_buffer.complete_state
        return (complete_state is not None and
                 complete_state.current_completion is not None)

    def __repr__(self):
        return "HasSelectedCompletion()"

mdsrosa:
@tsroten That worked! But when I typed `SELECT` it got executed and thrown a syntax error. Maybe we can do specifically to the `exit` command? I'm not sure if that is really worth it. What do you think?

tsroten:
Great! Well, you typed `SELECT` but didn't actually tab to select `SELECT`, right? So, even without my code, pressing enter wouldn't choose that autocomplete entry, it would have merely closed the menu.

mdsrosa:
Oh, I'm sorry. You're right. Now I typed `SELECT` then using the keyboard
arrows I selected the keyword and pressed enter and it didn't get executed.
Worked just fine. Nice! :D

--
Matheus Rosa

amjith:
Great job all around!!

I like the new custom filter. It works way better. 

@tsroten I had requested that we don't add a config option for legacy behavior. Here's why, this behavior is a more natural and a safer behavior than the old one. It will take people about 1 minute of using this to get used to the new behavior. 

But I'm starting to back pedal a little. Because both you and @mdsrosa had suggested adding a new config option to fallback to the legacy mode. So I'm wondering if it's more annoying than natural at first. 

I know that @mdsrosa is planning to get this ported to pgcli as well, so I'm going to ask both the core teams @dbcli/mycli-core and @dbcli/pgcli-core to weigh in on this matter. 

I'm perfectly happy to capitulate.

mdsrosa:
@amjith That's right! I'll be porting this to pgcli as well. I'm doing that right now. Thanks!
@tsroten Your filter works great. I'll be adding to the PR now. Thanks very much!

tsroten:
@amjith Well, I'm pretty indifferent on the matter. It's certainly natural enough. Some hardcore CLI users might want the option because of the extra keypress they could avoid. But, I think the majority of mycli users will be happy with the change.

tsroten:
@mdsrosa Glad to hear it. And, thank you for seeing this through. I think it's a definite improvement to the user-experience.

PR #173, opened 2015-10-16
Consolidate the timing information.

### LINKED ISSUE #172: Command Time 0 on MariaDB 10.0
I ran a query which took around 10s (checked with mysql cli) but the Command Time shows 0
```sql
mariadb root@localhost:vic_main> select count(1) from table1 where text like '%test%';
+------------+
|   count(1) |
|------------|
|       9670 |
+------------+
1 row in set
Command Time: 0.000s
Format Time: 0.000s
```
I verified it with some more queries, the Command Time shows 0 all the time.

Server version is 10.0.21-MariaDB on Ubuntu 12.04

amjith:
Fixes #172 

Our timing issue has been wrong for a while now. We changed the behavior of mycli to lazily consume the results of a query from a cursor. This meant the results of a query execution weren't consumed until we started to tabulate them.

So we are no longer able to separate the query time and format time. 

pgcli Ref: https://github.com/dbcli/pgcli/issues/378

Reviewer: @tsroten

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -405,7 +405,6 @@ class MyCli(object):
                     successful = False
                     start = time()
                     res = sqlexecute.run(document.text)
-                    duration = time() - start
                     successful = True
                     output = []
                     total = 0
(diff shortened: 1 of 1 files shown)


tsroten:
@amjith Looks good. Fast turnaround :)

PR #174, opened 2015-10-16
Add support for the system command

### LINKED ISSUE #130: Support 'system' command
`system` followed by a command is used to run a shell command. 

eg: 

```
mysql> system cat blah.csv
```

mdsrosa:
Fixes #130.

The `system` command is very useful. So what I did was register a "special command" to execute system commands.

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -139,6 +139,8 @@ class MyCli(object):
                 '\\T', 'Change Table Type.', aliases=('\\T',), case_sensitive=True)
         special.register_special_command(self.execute_from_file, 'source', '\\. filename',
                               'Execute commands from file.', aliases=('\\.',))
+        special.register_special_command(self.execute_system_command, 'system',
+                    '\\s', 'Execute a system command.', aliases=('\\s',))
 
     def change_table_format(self, arg, **_):
         if not arg in table_formats():
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -6,6 +6,7 @@ import os
 import sys
 import traceback
 import logging
+import subprocess
 from time import time
 from datetime import datetime
 from random import choice
@@ -169,6 +170,12 @@ class MyCli(object):
             return [(None, None, None, str(e))]
 
         return self.sqlexecute.run(query)
+
+    def execute_system_command(self, arg, **_):
+          if not arg:
+              message = 'Missing required argument: command.'
+              yield(None, None, None, message)
+          yield(None, None, None, subprocess.call(arg, shell=True))
 
     def initialize_logging(self):
 
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -6,7 +6,6 @@ import os
 import sys
 import traceback
 import logging
-import subprocess
 from time import time
 from datetime import datetime
 from random import choice
@@ -139,8 +138,6 @@ class MyCli(object):
                 '\\T', 'Change Table Type.', aliases=('\\T',), case_sensitive=True)
         special.register_special_command(self.execute_from_file, 'source', '\\. filename',
                               'Execute commands from file.', aliases=('\\.',))
-        special.register_special_command(self.execute_system_command, 'system',
-                    '\\s', 'Execute a system command.', aliases=('\\s',))
 
     def change_table_format(self, arg, **_):
         if not arg in table_formats():
@@ -172,12 +169,6 @@ class MyCli(object):
             return [(None, None, None, str(e))]
 
         return self.sqlexecute.run(query)
-
-    def execute_system_command(self, arg, **_):
-          if not arg:
-              message = 'Missing required argument: command.'
-              yield(None, None, None, message)
-          yield(None, None, None, subprocess.call(arg, shell=True))
 
     def initialize_logging(self):
 
--- a/mycli/packages/special/iocommands.py
+++ b/mycli/packages/special/iocommands.py
@@ -208,9 +208,10 @@ def execute_system_command(arg, **_):
     try:
         command = arg.strip()
         if command.startswith('cd'):
-            result, error_message = handle_cd_command(arg)
-            if not result:
+            ok, error_message = handle_cd_command(arg)
+            if not ok:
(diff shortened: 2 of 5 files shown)


mycli/main.py
```diff
@@ -170,6 +173,12 @@ def execute_from_file(self, arg, **_):
 
         return self.sqlexecute.run(query)
 
+    def execute_system_command(self, arg, **_):
+          if not arg:
+              message = 'Missing required argument: command.'
+              yield(None, None, None, message)
+          yield(None, None, None, subprocess.call(arg, shell=True))
+
```
amjith: This function shouldn't live in main.py. But instead should be transferred to https://github.com/dbcli/mycli/blob/master/mycli/packages/special/iocommands.py

I believe it should also use the @special_command() decorator instead of using the register_special_command() call.

The `register_special_command()` call is a way to write special commands that rely on things that are beyond the args and cursor, for example all the special command handlers in main.py uses the `self` variable in one way or another. Which is a good indication that we need those handlers to be part of this class. But the `execute_system_command` doesn't use the self variable that it takes in as an argument. Which is a good indication that it doesn't need to be part of this class.

mdsrosa: @amjith Hmm. Cool! Thanks for clarifying that to me! I'll move the function to iocommands.py. :)

amjith:
This is a great first pass at this feature and you're very close. 

As I mentioned in the inline comment, this special command handler should be moved inside the `mycli/packages/specials/iocommands.py` file. 

There are a couple of corner cases: 

1. `system cd <folder>` doesn't work. Because `cd` is not a shell command. This requires special handling. See https://github.com/donnemartin/saws/blob/master/saws/saws.py#L278 for a reference implementation on how to handle this. 

2. Unknown commands cause a python error in addition to shell error: 

```
mysql root@localhost:django_poll> system xyz
/bin/sh: xyz: command not found
'int' object has no attribute 'split'
```

The first error is bash error which is valid. The second error is a python error which should not happen. We should check the return status of a command if it is non-zero then we should send a custom message instead of sending the return code as message.

mdsrosa:
@amjith You're correct. I should be more careful. I'll fix this method. Thanks, Amjith! :)

mdsrosa:
@amjith I made the improvements you asked. Please let me know if I missed something.

mdsrosa:
Hmm. The build failed. Let me check that out.

j-bennet:
I think it's because help output has changed with addition of the new command.

mdsrosa:
@j-bennet Yes. I just fixed the expected line in the test that was failing (test_special_command).

mycli/packages/special/utils.py
```diff
@@ -0,0 +1,26 @@
+import os
+import subprocess
+
+def handle_cd_command(arg):
+    """Handles a `cd` shell command by calling python's os.chdir."""
+    CD_CMD = 'cd'
+    command = arg.strip()
```
amjith: This variable `command` is not used anywhere in this function.

mycli/packages/special/utils.py
```diff
@@ -0,0 +1,26 @@
...
+    directory = ''
+    error = False
+
+    tokens = arg.split(CD_CMD + ' ')
+    directory = tokens[-1]
+
+    try:
+        os.chdir(directory)
+        output = subprocess.check_output('pwd', stderr=subprocess.STDOUT, shell=True)
```
amjith: `subprocess.check_output()` was only introduced in Python 2.7. We've made a conscious decision to support Python 2.6 and above. So we can't use check_output(). But using `subprocess.call()` achieves the same results.

mycli/packages/special/utils.py
```diff
@@ -0,0 +1,26 @@
...
+    except OSError as e:
+        output, error = e.strerror, True
+
+    # formatting a nice output
+    if error:
+        output = "Error: {}".format(output)
+    else:
+        output = "Current directory: {}".format(output)
```
amjith: Let's not do this, because subprocess.call() doesn't return the output which can be stored in a variable. So this `format(output)` is only getting the return code not the stdout.

amjith:
@mdsrosa Would you mind adding a test for this? I found a bug while running it locally that would have been caught with a test.

mycli/packages/special/utils.py
mdsrosa: Oops! I'll remove the useless variable. Thanks! :smile:

mycli/packages/special/utils.py
mdsrosa: Ok, no problem. I'll make the change. :smile:

mdsrosa:
@amjith No problem. I'll add the tests after making the necessary changes. :smile:

mycli/packages/special/utils.py
mdsrosa: Okay :smile: :+1:

mycli/packages/special/iocommands.py
```diff
@@ -193,3 +195,28 @@ def delete_favorite_query(arg, **_):
...
+
+    if not arg:
+      return [(None, None, None, usage)]
+
+    try:
+        command = arg.strip()
+        if command.startswith('cd'):
+            result, error_message = handle_cd_command(arg)
+            if not result:
+                return [(None, None, None, error_message)]
```
amjith: There needs to be an `else` clause that does; 
```
return [(None, None, None, '')]
```

mdsrosa: @amjith I think I've a better idea. I'll not add a `else` statement I'll just add the `return [(None, None, None, '')]` line instead. Is that okay? :smiley:

amjith: Yup. That works too.

mdsrosa: Done. :smile:

amjith:
:+1: Thank you for the PR and the tests. 

:rocket:

mdsrosa:
@amjith No problem. I'm very happy to contribute. :smile:

PR #175, opened 2015-10-17
Handle decoding errors in older pymysql.

amjith:
@tsroten Can you review?

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -425,13 +425,15 @@ class MyCli(object):
                         end = time()
                         total += end - start
                         mutating = mutating or is_mutating(status)
-                except UnicodeDecodeError:
+                except UnicodeDecodeError as e:
                     import pymysql
-                    ver = pymysql.__version__.split('.')
-                    if ver < ('0', '6', '7'):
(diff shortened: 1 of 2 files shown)


mycli/main.py
```diff
@@ -425,6 +425,13 @@ def prompt_tokens(cli):
                         end = time()
                         total += end - start
                         mutating = mutating or is_mutating(status)
+                except UnicodeDecodeError:
+                    import pymysql
+                    ver = pymysql.__version__.split('.')
+                    if ver < ('0', '6', '7'):
+                        message = ('You are running an older version of pymysql.\n'
+                                'Please upgrade to 0.6.7 or above to view binary data.')
+                        self.output(message)
```
tsroten: What if an `UnicodeDecodeError` is raised, but the PyMySQL version is 0.6.7 or above? Shouldn't it raise the original error message?

tsroten:
@amjith I like the error message. It's very clear and helpful.

amjith:
Good point about reraising exception. I've changed the PR to do that.

PR #176, opened 2015-10-19
Amjith/background completion

amjith:
Reviewer: @tsroten

The completion data-structures are populated at the startup time and then refreshed every time we ALTER, DROP, CREATE or change the database. But this completion refresh happens synchronously which could tie up the REPL for a while if the database is huge (with hundreds of thousands of tables). So instead of refreshing the completions synchronously this PR will make it asynchronous by kicking off a background thread. 

This is a copy of the feature that was implemented in pgcli a little while back. 

See this PR for discussion: https://github.com/dbcli/pgcli/pull/345

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -6,6 +6,7 @@ import os
 import sys
 import traceback
 import logging
+import threading
 from time import time
 from datetime import datetime
 from random import choice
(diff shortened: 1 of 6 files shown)


tsroten:
@amjith This looks like a great feature! Other than my comment in the code, it looks good to me.

mycli/main.py
```diff
@@ -475,7 +484,7 @@ def prompt_tokens(cli):
 
                 # Refresh the table names and column names if necessary.
                 if need_completion_refresh(document.text):
-                    self.refresh_dynamic_completions()
+                    self.refresh_completions(reset=need_completion_reset(document.text))
```
tsroten: Should you check if the command was successful before starting a thread to refresh completions? If the create/alter/update statement wasn't successful, wouldn't that be a waste of resources, especially if the database is large?

amjith: You're right I didn't think about that. I'll see what I can do.

amjith:
I've updated the PR to do the completion refresh only if the command was a success.

PR #178, opened 2015-10-23
Fix typo on the handler's name when registering the `rehash` special command

mdsrosa:
Since there is no test for `rehash` command the travis-ci didn't get this typo. But when you run `mycli` you get a `AttributeError` exception.

I didn't find any method called `refresh_dynamic_completions`. So I've fixed the typo on the handler's name for the `rehash` command when registering it.


tsroten:
Nice catch. I have no idea why that didn't throw an error earlier when I was testing [#176](https://github.com/dbcli/mycli/pull/176).

mdsrosa:
@tsroten Thank you. I have no idea too. All I know is that it thrown the exception as soon as I run `mycli` in the terminal.

PR #180, opened 2015-10-23
Add ability to make suggestions for compound join clauses

mdsrosa:
This is a feature ported from **pgcli**. https://github.com/dbcli/pgcli/pull/371

It adds the ability to make alias suggestions for `AND`/`OR` in an `ON` clause.

From the PR:
Previously, this worked correctly, suggesting columns etc. from a
`SELECT * FROM abc a JOIN def d ON a.`
But this suggested only keywords:
`SELECT * FROM abc a JOIN def d ON a.id = d.id AND a.`

Thank you @darikg! :)


amjith:
Looks good. :roller_coaster: 

Thank you!

Issue #169, opened 2015-10-13
Make pycrypto optional


amjith:
PR #152 brought in the dependency for pycrypto. This is used to parse the mylogin.cnf file. 

This is a hard package to install on Windows. So I'd like to make pycrypto an optional dependency. When pycrypto is installed we can read mylogin.cnf, when it is not available we will skip reading mylogin.cnf but print a warning message at startup that we found a mylogin.cnf file but did not read in the details due to missing pycrypto package.

j-bennet:
You mind if I take this @amjith?

amjith:
It is now yours @j-bennet.

PR #181, opened 2015-10-25
Make pycrypto optional.

j-bennet:
This handles issue #169 about making pycrypto dependency optional.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,8 +29,10 @@ setup(
             'PyMySQL >= 0.6.2',
             'sqlparse >= 0.1.16',
             'configobj >= 5.0.6',
-            'pycrypto >= 2.6.1',
             ],
+        extras_require={
+            'parse_mylogin_cnf': ['pycrypto >= 2.6.1']
+        },
         entry_points='''
             [console_scripts]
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -29,8 +29,10 @@
             'PyMySQL >= 0.6.2',
             'sqlparse >= 0.1.16',
             'configobj >= 5.0.6',
-            'pycrypto >= 2.6.1',
             ],
+        extras_require={
+            'parse_mylogin_cnf': ['pycrypto >= 2.6.1']
+        },
```
amjith: This makes it so pycrypto is never installed by default. It is only installed when someone goes out and does `pip install pycrypto`. 

I'd prefer to take the approach of not installing pycrypto in Windows but install it everywhere else. 

See https://github.com/dbcli/pgcli/pull/397 for reference on how to do this.

amjith:
I've checked the remaining changes and they work great. Good job!

j-bennet:
@amjith Didn't see this approach before - thanks! I hope it's good now.

amjith:
Looks good to me. :+1: 

Thanks for the update. 

:soccer:

Issue #75, opened 2015-07-02
Add config option for always on logging


amjith:
Related to #45 

The logging option to log everything from queries to results with a timestamp should be available as a config option. 

/cc @mysz

PR #182, opened 2015-10-28
Add a config option for always logging

mdsrosa:
Addresses #75.

I added a new config in `myclirc` called `audit_log`. This is going to be used for always logging the queries. The user can still use the "new" `-a` or `--auditlog` argument when running `mycli`.

But since we've a new config option in `myclirc` I think we can drop the `--auditlog` argument. What do you guys think?

:)

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -71,10 +71,10 @@ class MyCli(object):
     ]
 
     def __init__(self, sqlexecute=None, prompt=None,
-            logfile=None, defaults_suffix=None, defaults_file=None,
+            auditlog=None, defaults_suffix=None, defaults_file=None,
             login_path=None):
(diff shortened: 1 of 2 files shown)


amjith:
I'd prefer not to change the existing command line option. Because that will break backward compatibility and I don't want to do that without a good reason. 

It's ok to call the config option as 'audit_log' in the config file but 'logfile' in the command line option.

mdsrosa:
@amjith I changed the option name from`logfile` to `auditlog` in order to be more semantic to its purpose. But that's okay I'll changed back to `logfile`. ;-)

amjith:
Thanks @mdsrosa. I'll check this one when I wake up and merge.

amjith:
Thank you!

:koala:

PR #185, opened 2015-10-29
Make `audit_log` config optional

mdsrosa:
Since not everyone wants to make an audit log this PR makes the `audit_log` optional. 

I was thinking about writing some tests for this feature. What do you guys think?

cc/ @amjith

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -101,7 +101,7 @@ class MyCli(object):
         self.wider_completion_menu = c['main'].as_bool('wider_completion_menu')
 
         # audit log
-        if self.logfile is None:
+        if self.logfile is None and 'audit_log' in c['main']:
             self.logfile = open(os.path.expanduser(c['main']['audit_log']), 'a')
 
         self.completion_refresher = CompletionRefresher()
--- a/mycli/myclirc
+++ b/mycli/myclirc
@@ -23,8 +23,9 @@ log_file = ~/.mycli.log
 # and "DEBUG".
 log_level = INFO
 
-# Log every query and its results to a file.
-audit_log = ~/.mycli-audit.log
+# Log every query and its results to a file. Enable this by uncommenting the
+# line below.
+# audit_log = ~/.mycli-audit.log
 
 # Timing of sql statments and table rendering.
 timing = True


amjith:
@tsroten Can you review this one? 

I'm slammed at the moment.

tsroten:
@mdsrosa This is a useful option! A couple tests would be good (one with audit_log defined and one with it commented out). Should be an easy commit to test.

tsroten:
@mdsrosa Actually, I don't see an easy way to override the `~/.myclirc` in order to test this. Do you?

tsroten:
@mdsrosa You could override the `load_config()` function once you import `mycli.main` to read from a test `myclirc`.

tsroten:
@mdsrosa Sorry for the barrage of comments. I'm not sure the test would be worth it in this case. Might be better to simply add a `try`/`except` clause when `MyCli()` opens the log file to ensure the user wrote the path correctly. If opening it fails you could print something like "Unable to open log file: {PATH}" and continue.

amjith:
Should we let the user continue in the event of wrong path? 

If it were to crash they'd be forced to evaluate why it crashed and fix the issue. Otherwise they might ignore the error message and continue working under the assumption that an audit log is being created, only to find out later that none of it was saved because they missed the error message early on.

amjith:
Maybe we can crash with a decent looking error message instead of throwing up a stack trace.

tsroten:
@amjith You have a good point. My thinking was that a log message isn't essential to mycli working, so we shouldn't quit. Think of this soon-to-be use case: the config used is `/etc/myclirc` and it is misconfigured to point to a file that the user can't append to. The user can't edit `/etc/myclirc` so we would force the user to create a `~/.myclirc` in order to run mycli (or wait for the admin to reconfigure the config file).

amjith:
You make a good point. How about we print the error message in red after each query? So the user has plenty of warning.

The reason I say this is because the original requestor of this feature @johshoff had a specific use case in mind. He wanted to enable audit logging when he's woken up in the middle of the night by a page and he had to run some commands on the production database to fix things. 

When I look at it from his point of view, I think we should take every precaution to make sure the user is aware of their mistake, especially a sleepy user. 

I might be over thinking this use case, so let me know my suggestion is too crazy. :smile:

tsroten:
That makes sense. If the audit log is configured, the user might have a really good reason why they need to log every query. I only have read-access to our production MySQL servers at work, but those with write-access would definitely need to have their history logged.

johshoff:
Thanks for thinking about sleepy users with too much DB access :sleeping:

mdsrosa:
@amjith @tsroten Thank you for your feedback.

About the tests: I don't know if is worth it too. I also didn't find anything that could be done to verify if the `audit_log` is enabled or not. Maybe we could add a `myclirc.test` file into the tests folder in order to test the `audit_log` option (or any other option we may need). But that will be difficult to maintain because if the official `myclirc` changes then we will need to change the `myclirc.test` as well. So I don't know if that is a good solution.

About the error message: Since the `audit_log` config is optional If the user enables it (it means he is really going to need it) I think we should quit the program and show the following error message: "Audit log is enabled. File path X was not found. Cannot proceed." or something like that.

amjith:
@mdsrosa If you noticed the last few comments, @tsroten made a compelling reason to not quit the program. So we settled on showing an error message before/after each query. 

That can be added as a separate PR and I don't see any reason why we should block this PR. 

But I'll leave it up to @tsroten to make the final call.

mdsrosa:
@amjith Oh yeah. That's right. Now I agree. Is better to not quit the program. 
But adding a error message before/after each query could be annoying, right? Or this is the point of that solution? :smile: hehehe

Well, either way, notifying the user about the error is the best thing we can do. :smiley: 

Okay. While we wait for @tsroten I'll start working on the error message. :wink:

PR #186, opened 2015-10-30
Add error message for auditlog's validation

mdsrosa:
As discussed in #185. This PR add the error message for the `audit_log` config file validation.

The error message will only show up if the user enabled the `audit_log` config but added a invalid path to the audit log file. So in order to help the user fix the path, the error message will show up after each query is executed.

This PR is related to the issue #75.

Please, can one of you review this?
cc/ @amjith @tsroten 

:wink:

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -431,10 +431,12 @@ class MyCli(object):
 
                 try:
                     logger.debug('sql: %r', document.text)
-                    if self.logfile:
+                    if self.logfile is not False:
                         self.logfile.write('\n# %s\n' % datetime.now())
                         self.logfile.write(document.text)
                         self.logfile.write('\n')
+                    else:
+                        self.output("Error: Unable to load the audit log file.", err=True, fg='red')
                     successful = False
                     start = time()
                     res = sqlexecute.run(document.text)
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -437,8 +437,6 @@ class MyCli(object):
                         self.logfile.write('\n# %s\n' % datetime.now())
                         self.logfile.write(document.text)
                         self.logfile.write('\n')
-                    elif self.logfile is False:
-                        self.output("Error: Unable to write to the audit log file.", err=True, fg='red')
 
                     successful = False
                     start = time()
@@ -517,7 +515,9 @@ class MyCli(object):
                     if need_completion_refresh(document.text):
                         self.refresh_completions(
                                 reset=need_completion_reset(document.text))
-
+                finally:
+                    if self.logfile is False:
+                        self.output("Warning: This query was not logged.", err=True, fg='red')
                 query = Query(document.text, successful, mutating)
                 self.query_history.append(query)
 
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -105,6 +105,7 @@ class MyCli(object):
             try:
                 self.logfile = open(os.path.expanduser(c['main']['audit_log']), 'a')
             except (IOError, OSError) as e:
+                self.output('Error: Unable to open the audit log file.', err=True, fg='red')
                 self.logfile = False
 
         self.completion_refresher = CompletionRefresher()
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -102,7 +102,10 @@ class MyCli(object):
 
         # audit log
         if self.logfile is None and 'audit_log' in c['main']:
-            self.logfile = open(os.path.expanduser(c['main']['audit_log']), 'a')
+            try:
+                self.logfile = open(os.path.expanduser(c['main']['audit_log']), 'a')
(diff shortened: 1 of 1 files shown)


mycli/main.py
```diff
@@ -428,10 +431,12 @@ def prompt_tokens(cli):
 
                 try:
                     logger.debug('sql: %r', document.text)
-                    if self.logfile:
+                    if self.logfile is not False:
```
tsroten: I would keep this as `if self.logfile`. That will catch any open file handles. Right now it will throw an error because a `self.logfile` with a value of `None` will be evaluated to `True`.

mycli/main.py
```diff
@@ -428,10 +431,12 @@ def prompt_tokens(cli):
...
                         self.logfile.write('\n# %s\n' % datetime.now())
                         self.logfile.write(document.text)
                         self.logfile.write('\n')
+                    else:
```
tsroten: Try `elif self.logfile is None` instead of `else`.

mycli/main.py
```diff
@@ -428,10 +431,12 @@ def prompt_tokens(cli):
...
+                        self.output("Error: Unable to load the audit log file.", err=True, fg='red')
```
tsroten: At this point we aren't loading the file. Maybe a more helpful error message might be `...write to the audit log file.`

mycli/main.py
```diff
@@ -102,7 +102,10 @@ def __init__(self, sqlexecute=None, prompt=None,
 
         # audit log
         if self.logfile is None and 'audit_log' in c['main']:
-            self.logfile = open(os.path.expanduser(c['main']['audit_log']), 'a')
+            try:
+                self.logfile = open(os.path.expanduser(c['main']['audit_log']), 'a')
+            except (IOError, OSError) as e:
+                self.logfile = False
```
tsroten: You might consider adding an error message here as well so the user won't blindly continue without knowing their audit log file wasn't opened.

tsroten:
@mdsrosa This is a great start! I've made some suggestions in the code for you.

mycli/main.py
mdsrosa: Yes! I thought that after I pushed the branch. :grin: Thank you!

mycli/main.py
mdsrosa: You're right. I'll fix this. Thank you, @tsroten! :)

mycli/main.py
mdsrosa: :+1:

mycli/main.py
mdsrosa: @tsroten Good idea! I'm thinking about adding this error message: `Error: Unable to open audit log file`. What do you think?

mdsrosa:
@tsroten I've made the changes you suggested. Thank you! ;-)

tsroten:
@mdsrosa I just tested this branch out. It works very well. The red error messages are easy to see and draw your attention. Good job on that.

I'm wondering if the way we word these messages is a little cryptic. For the first message, the user is alerted to what went wrong, but they might not understand the consequence of that. What if the first message was something like `Error: Unable to open audit log file. Your queries will not be logged.`

Then subsequent messages might be `Warning: This query is not being logged.`

What do you think about that?

mdsrosa:
@tsroten Thank you! 

About the messages: Yes! I agree with you. We should be more clear. I liked your suggestions they make a lot sense. I'll update the error messages then. ;-)

mdsrosa:
@tsroten Should we change the color of the "Warning" message? Or keep it red?

tsroten:
I'd say to leave it red. @amjith [originally suggested that in #185](https://github.com/dbcli/mycli/pull/185#issuecomment-152539735). It would be too easy to ignore otherwise. Most people will hopefully never even see this error message. And, not writing to the audit log could have pretty important implications, so being extra verbose here is probably warranted.

Another question is whether the message for each query should be outputted before or after the results. As it is currently, if the results are long enough, the user might never see the query. So, we should probably output the warning message after the results. For this, we'll need to add a `finally` clause after the [`else` clause in 508](https://github.com/dbcli/mycli/blob/auditlog_error_message/mycli/main.py#L508).  By putting the warning message in a `finally` clause, we'll make sure that users see it regardless of the query being successful or not since the audit file logs erroneous queries anyways. 

If we do output the warning message after the results/error messages, we should change it to the past tense: `Warning: This query was not logged.`

What do you think?

mdsrosa:
@tsroten That's a great point of view. I haven't thought of that. I believe that it will be a lot better if we show the message after the results. As you presented, it makes a lot more sense.This will not make users forget that the query hasn't been logged.

I've updated the PR. :wink:

tsroten:
@mdsrosa Okay, I tested it with Python 2.7 and 3.5. It works great! Thanks!

Issue #190, opened 2015-11-09
check type of cli arguments before using them


Knorkebrot:
Something like a `Usage: ...` would be nice. At least don't crash :)

```
root@debian-wheezy:~# mycli --user foo -P bar baz
/usr/local/lib/python2.7/dist-packages/mycli/main.py:565: Warning: Click detected the use of the unicode_literals __future__ import.  This is heavily discouraged because it can introduce subtle bugs in your code.  You should instead use explicit u"" literals for your unicode strings.  For more information see http://click.pocoo.org/python3/
  @click.argument('database', default='', nargs=1)
Traceback (most recent call last):
  File "/usr/local/bin/mycli", line 9, in <module>
    load_entry_point('mycli==1.4.0', 'console_scripts', 'mycli')()
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 700, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 680, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 873, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 508, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/mycli/main.py", line 582, in cli
    mycli.connect(database, user, password, host, port, socket)
  File "/usr/local/lib/python2.7/dist-packages/mycli/main.py", line 242, in connect
    port = int(port or cnf['port'] or 3306)
ValueError: invalid literal for int() with base 10: 'bar'
```

mdsrosa:
Nice catch, @Knorkebrot! That's really not cool. I'm looking into this right now and I'll make sure that this will be fixed before our next release. Thank you!

mdsrosa:
This issue has been fixed. Thank you for reporting it, @Knorkebrot! :smile:

Knorkebrot:
Great :+1:

PR #191, opened 2015-11-09
Add type validation for `port` argument

mdsrosa:
I added `int` as the `type` for the `-P`/`--port` argument. This fixes the problem reported in #190.

@amjith @tsroten @j-bennet 
What do you guys think about this solution? I'm open for ideas.

By the way, I think this should be merged before releasing our next version (1.5.0).

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -289,7 +289,7 @@ class MyCli(object):
             socket = socket or cnf['socket']
         user = user or cnf['user'] or os.getenv('USER')
         host = host or cnf['host'] or 'localhost'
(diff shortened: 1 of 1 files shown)


mdsrosa:
I just remember it. I'll need to test the `port` defined in the `cnf` file. I'll update this PR soon if needed.

mdsrosa:
I just tested with the `.my.cnf` file. It doesn't break anything. :wink:

amjith:
Looks good. I'll merge this in.

mdsrosa:
@amjith Thank you! :smile:

PR #193, opened 2015-11-11
Cast the value of port read from my.cnf to int.

amjith:
Reviewer: @mdsrosa


mdsrosa:
That bug was introduced by me on my bugfix on the port validation. I'm sorry. Thank you very much, @amjith.

Issue #196, opened 2015-11-13
error on startup


ian-paterson:
When I try to open the program using:
```
mycli -h 192.168.0.1 -u username -psecret
```

I'm getting this error
```
Traceback (most recent call last):
  File "/usr/local/Cellar/mycli/1.5.1/libexec/bin/mycli", line 9, in <module>
    load_entry_point('mycli==1.5.1', 'console_scripts', 'mycli')()
  File "/usr/local/Cellar/mycli/1.5.1/libexec/vendor/lib/python2.7/site-packages/click/core.py", line 700, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/Cellar/mycli/1.5.1/libexec/vendor/lib/python2.7/site-packages/click/core.py", line 680, in main
    rv = self.invoke(ctx)
  File "/usr/local/Cellar/mycli/1.5.1/libexec/vendor/lib/python2.7/site-packages/click/core.py", line 873, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/Cellar/mycli/1.5.1/libexec/vendor/lib/python2.7/site-packages/click/core.py", line 508, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/Cellar/mycli/1.5.1/libexec/lib/python2.7/site-packages/mycli/main.py", line 639, in cli
    mycli.connect(database, user, password, host, port, socket)
  File "/usr/local/Cellar/mycli/1.5.1/libexec/lib/python2.7/site-packages/mycli/main.py", line 292, in connect
    port = int(port or cnf['port']) or 3306
TypeError: int() argument must be a string or a number, not 'NoneType'
```

amjith:
Yup. That is a bug. Sorry about that. I'll release a fix for it asap.

In the meantime you can work around it by providing a port number in command line 

```
mycli -h 192.168.0.1 -u username -psecret -P 3306
```

ian-paterson:
@amjith Thanks for the quick fix!

amjith:
This is now fixed in version 1.5.2. Please upgrade: 

```
pip install -U mycli
```

PR #197, opened 2015-11-13
Amjith/fix port conversion

amjith:
Reviewer: @tsroten 

This fix is to address #196.

Once this is merged in, I'll release 1.5.2.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,11 @@
+1.5.2:
+======
+
+Bug Fixes:
+----------
+
+* Protect against port number being None when no port is specified in command line.
+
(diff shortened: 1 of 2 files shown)


tsroten:
@amjith Looks good. I'm going to merge this because the the fix needs to be released immediately.

Just wanted to note that, if we convert the type without catching `ValueError`, this can happen:

* User accidentally puts `'3306a'` in their configuration file.
* They run mycli and see a traceback with ```ValueError: invalid literal for int() with base 10: '3306a'```

They might not even realize the problem is the port number. Do you think we should handle that error?

PR #199, opened 2015-11-13
Handle invalid port validation exception.

amjith:
@tsroten You have a valid point. I've handled the ValueError in this PR. We can fix this in our next release since this is not an urgent requirement.

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -289,10 +289,12 @@ class MyCli(object):
             socket = socket or cnf['socket']
         user = user or cnf['user'] or os.getenv('USER')
         host = host or cnf['host'] or 'localhost'
+        port = port or cnf['port'] or 3306
         try:
-            port = int(port or cnf['port'] or 3306)
+            port = int(port)
         except ValueError as e:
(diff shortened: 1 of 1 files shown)


loslots:
I believe this 3306 # is from my phone I was on in the Early AM hoping to get help fixing my phone I just had turn on and like all my last 6+ phones it seems to be connected somehow to the rest. My SigvothervI believe has done this and uses them to 1 c what I do and 2 go to Chat/Dating or maybe worse (pornography)  its on my yahoo Acct as well so if I can help please let me know and maybe you might put it up on the priority list a bit please. I'm crushed to the core at what I seen today and have been seeing pieces of for going on 2 yrs.Please help and thank you

tsroten:
@amjith Nice!

Do you think `Invalid port number: '3306a'` might be more helpful than `Invalid port number. invalid literal for int() with base 10: '3306a'`?

amjith:
I couldn't find a good way extract out just the port number. If you can
think of a good way to do it, please feel free to edit.

amjith:
Thanks for making it user friendly.

 :sunflower:

PR #200, opened 2015-11-14
Adds support for system-wide config file.

### LINKED ISSUE #161: System wide myclirc
Should create a default myclirc in /etc/defaults (on Debian based distros) when can then be overriden by the users ~/.myclirc if they wish to change the behaviour.

Will make it easier for SysAdmins to control users setup.

Additional option to disable checking/parsing of ~/.myclirc in system wide config would also be great.

tsroten:
This fixes #161 by introducing system-wide config files. Right now, only `'/etc/myclirc'` is checked, but adding other locations to the code is a simple as adding an item to a list. If the last successful file read is not a system-wide config file, then mycli checks to see if it should write the default user config file (by seeing if one exists). But, if the last successful file read was a system-wide config file, then no default user config file is created.

This cleans up and refactors some of the `ConfigObj()` code to use a generic function that handles two types of errors: config parsing errors and file permission errors.

Because logging isn't initialized before we read the config files, these two outputs are possible:
```
% mycli
You don't have permission to read config file '/etc/myclirc'.
Version: 1.5.2
Chat: https://gitter.im/dbcli/mycli
Mail: https://groups.google.com/forum/#!forum/mycli-users
Home: http://mycli.net
Thanks to the contributor - Foo
mysql dev@localhost:(none)>
```

And:
```
% mycli
Error parsing config file '/etc/myclirc'.
Recovering partially parsed config values.
Version: 1.5.2
Chat: https://gitter.im/dbcli/mycli
Mail: https://groups.google.com/forum/#!forum/mycli-users
Home: http://mycli.net
Thanks to the contributor - Foo
mysql dev@localhost:(none)>
```

I'm not sure if there is a better way to deal with those messages. They aren't printed by `mycli.py` so they don't know about `click.secho()`. They are basic logging calls to the default logging handler (stderr). Can anyone think of a better way to handle this? Or are we happy with it as it is?

### DIFF
--- a/mycli/config.py
+++ b/mycli/config.py
@@ -19,29 +19,36 @@ class CryptoError(Exception):
 
 logger = logging.getLogger(__name__)
 
-def read_config_files(files, base_config=None):
-    """Read and merge a string or list of config files.
+def read_config_file(f, base_config=None):
+    """Read and merge a config file.
 
-    If a file is read successfully, the config object takes on that
-    filename.
+    If the file is read successfully, the config object takes
+    on that filename.
     """
 
-    config = base_config or ConfigObj()
-    files = [files] if isinstance(files, str) else files
+    config = ConfigObj() if base_config is None else base_config
+    try:
+        _config = ConfigObj(f, interpolation=False)
+        config.merge(_config)
+        if bool(_config) is True:
+            config.filename = f
+    except ConfigObjError as e:
+        logger.error("Error parsing config file '{0}'.".format(f))
+        logger.error('Recovering partially parsed config values.')
+        config.merge(e.config)
+    except (IOError, PermissionError) as e:
+        logger.warning("You don't have permission to read config "
+                       "file '{0}'.".format(e.filename))
+
+    return config
+
+def read_config_files(files, base_config=None):
+    """Read and merge a list of config files."""
+
+    config = ConfigObj() if base_config is None else base_config
 
     for _file in files:
-        try:
-            _config = ConfigObj(_file, interpolation=False)
-            if bool(_config) is True:
-                config.filename = _file
-            config.merge(_config)
-        except ConfigObjError as e:
-            logger.error("Error parsing config file '{0}'.".format(_file))
-            logger.error('Recovering partially parsed config values.')
-            config.merge(e.config)
-        except (IOError, PermissionError) as e:
-            logger.warning("You don't have permission to read config "
-                           "file' {0}'.".format(e.filename))
+        read_config_file(_file, base_config=config)
 
     return config
 
--- a/mycli/config.py
+++ b/mycli/config.py
@@ -27,7 +27,8 @@ def read_config_file(f):
     try:
         config = ConfigObj(f, interpolation=False)
     except ConfigObjError as e:
-        print("Error parsing config file '{0}'.".format(f), file=sys.stderr)
+        print("Error parsing line {0} of config file '{1}'.".format(
+              e.line_number, f), file=sys.stderr)
         print('Recovering partially parsed config values.', file=sys.stderr)
         return e.config
     except (IOError, OSError) as e:
--- a/mycli/config.py
+++ b/mycli/config.py
@@ -19,36 +19,32 @@ class CryptoError(Exception):
 
 logger = logging.getLogger(__name__)
 
-def read_config_file(f, base_config=None):
-    """Read and merge a config file.
+def read_config_file(f):
+    """Read a config file."""
 
-    If the file is read successfully, the config object takes
-    on that filename.
-    """
-
-    config = ConfigObj() if base_config is None else base_config
     try:
-        _config = ConfigObj(f, interpolation=False)
-        config.merge(_config)
-        if bool(_config) is True:
-            config.filename = f
+        config = ConfigObj(f, interpolation=False)
     except ConfigObjError as e:
         logger.error("Error parsing config file '{0}'.".format(f))
         logger.error('Recovering partially parsed config values.')
-        config.merge(e.config)
-    except (IOError, PermissionError) as e:
+        return e.config
+    except (IOError, OSError) as e:
         logger.warning("You don't have permission to read config "
                        "file '{0}'.".format(e.filename))
+        return None
 
     return config
 
-def read_config_files(files, base_config=None):
+def read_config_files(files):
     """Read and merge a list of config files."""
 
-    config = ConfigObj() if base_config is None else base_config
+    config = ConfigObj()
 
     for _file in files:
-        read_config_file(_file, base_config=config)
+        _config = read_config_file(_file)
+        if bool(_config) is True:
+            config.merge(_config)
+            config.filename = _file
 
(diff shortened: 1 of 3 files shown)


mycli/config.py
```diff
@@ -19,16 +19,40 @@ class CryptoError(Exception):
...
+                       "file '{0}'.".format(e.filename))
+
+    return config
+
+def read_config_files(files, base_config=None):
+    """Read and merge a list of config files."""
+
+    config = ConfigObj() if base_config is None else base_config
+
+    for _file in files:
```
amjith: Nice trick to avoid stomping on built-in `file` keyword.

mycli/config.py
```diff
@@ -19,16 +19,40 @@ class CryptoError(Exception):
...
-    return cfg
+    If the file is read successfully, the config object takes
+    on that filename.
+    """
+
+    config = ConfigObj() if base_config is None else base_config
+    try:
+        _config = ConfigObj(f, interpolation=False)
+        config.merge(_config)
+        if bool(_config) is True:
```
amjith: Thank you for being explicit here. It is tempting to do `if bool(_config):` instead of `if bool(_config) is True:`, but I like that you've explicitly called this out. Makes it clear to read.

mycli/main.py
```diff
@@ -85,12 +94,10 @@ def __init__(self, sqlexecute=None, prompt=None,
...
 
-        default_config = os.path.join(PACKAGE_ROOT, 'myclirc')
-        write_default_config(default_config, '~/.myclirc')
-
-
         # Load config.
-        c = self.config = load_config('~/.myclirc', default_config)
+        c = self.config = read_config_file(self.default_config_file)
+        read_config_files(self.system_config_files, base_config=c)
+        read_config_file(self.user_config_file, base_config=c)
```
amjith: Mutating `base_config` on each calls feels wrong. Can we collapse this to a single call? 

`self.config = read_config_files(config_files)` where `config_files = [default_config_file, system_config_files, user_config_file]`. 

I realize that `config_files` assignment is not right since it will result in nested list.

amjith:
Looks like you're targeting Python 3. There is no such thing called `PermissionError` in Python 2.7.

amjith:
I was surprised when you said it'll print the error messages at startup because I've never seen that behavior when loggers are uninitialized. 

Here's what I get when there is a parsing error in /etc/myclirc. 

```
No handlers could be found for logger "mycli.config"
Version: 1.5.2
Chat: https://gitter.im/dbcli/mycli
Mail: https://groups.google.com/forum/#!forum/mycli-users
Home: http://mycli.net
Thanks to the contributor - Chris Anderton
mysql root@localhost:django_poll>
```

mycli/main.py
tsroten: Yeah, I'll fix this. I originally had it set up to not mutate on each call. I changed, but I forgot why.

tsroten:
[It looks like](https://docs.python.org/3.5/library/logging.html#logging.lastResort), as of Python 3.2, logging writes to `sys.stderr` using a `StreamHandler` in the event that logging isn't configured. In Python 2, we'd have to add a `NullHandler` in order to silence that annoying `No handlers could be found...` message.

I think that most users expect command-line apps to write user-facing error messages to stderr. I'd hate for our Python 2 users to not see a potentially helpful error message. So, I suggest that we use the `print()` function to write to `sys.stderr` for all rc-related file reading/parsing (instead of using the logging library). @amjith What are your thoughts on that?

By the way, thanks for pointing out the `PermissionsError`. I'll fix that.

amjith:
> It looks like, as of Python 3.2, logging writes to sys.stderr using a StreamHandler in the event that logging isn't configured.

Python 3 really makes it easy to do the right thing. 

> I think that most users expect command-line apps to write user-facing error messages to stderr. I'd hate for our Python 2 users to not see a potentially helpful error message. So, I suggest that we use the print() function to write to sys.stderr for all rc-related file reading/parsing (instead of using the logging library). @amjith What are your thoughts on that?

Yes!! I completely agree. We should make choices that are most intuitive for the user. So, let's use print statements there.

mycli/config.py
```diff
@@ -19,16 +21,37 @@ class CryptoError(Exception):
...
-    cfg.filename = expanduser(usr_cfg)
+def read_config_file(f):
+    """Read a config file."""
 
-    return cfg
+    try:
+        config = ConfigObj(f, interpolation=False)
+    except ConfigObjError as e:
+        print("Error parsing line {0} of config file '{1}'.".format(
+              e.line_number, f), file=sys.stderr)
```
amjith: Nice touch adding the line numbers. We might get some complaints regarding this error message since ConfigObj will fail on some of the constructs such as `!includedir /usr/local/etc/my.cnf.d`. But I think that is inevitable.

mycli/packages/special/favoritequeries.py
```diff
@@ -57,5 +58,5 @@ def delete(self, name):
         self.config.write()
         return '%s: Deleted' % name
 
-from ...config import load_config
-favoritequeries = FavoriteQueries(load_config('~/.myclirc'))
+from ...config import read_config_file
+favoritequeries = FavoriteQueries(read_config_file(expanduser('~/.myclirc')))
```
amjith: Can we add the `expanduser()` to the read_config_file() and read_config_files() as well? I think those two functions should be more defensive. 

Also, `expanduser()` is idempotent, so it should be harmless to call it on path values multiple times.

amjith:
An interesting side effect of this change is now I'm aware of an inefficiency. I noticed that when I launched mycli I got the `Error parsing...` message three times. Further investigation revealed we're reading it once for pager, once for prompt and once for connection details.

Optimizing that should help a little bit with startup time. :)

amjith:
Changes look good and it works without a hitch. I'm a bit hesitant about the parsing error being printed at startup. ConfigObj can't parse some of the INI file extension that might be valid for mysql but not valid for ConfigObj. So people can't change those lines to be ConfigObj compliant. This means we'll be printing that error message every time they start mycli. Which seems excessive. 

Since this is only an issue when we're reading my.cnf files, can we log those messages to the log file instead of stdout? 

This means, errors in RC files will be printed to stdout, but errors in my.cnf files will be logged to the log file. 

I can't think of a better solution. Suggestions welcome.

tsroten:
@amjith I agree. I'll have `config.py` check to see if there is a parent logger (which means logging is initialized) to know whether to print to stderr or write to the logging file. That means:
* All `*.cnf`-related errors will be written to the log file, not printed to stderr.
* The first pass (which I'll work on optimizing later) on the `myclirc`s will print errors to stderr. Subsequent passes will write to the logging file.

tsroten:
@amjith Ok, I've added those changes. There is some optimizing to reading config files that could be done, but this doesn't seem like the appropriate pull request for that. Let me know what you think of the changes.

amjith:
Looks good. :+1: 

You're right the optimization are a digression to this PR. We can do it on a different PR.

PR #209, opened 2015-12-09
Fix #203 - Support \G terminator for \f queries

### LINKED ISSUE #203: Favorite Queries do not accept \G statement terminator
I have some queries that use the \G terminator for vertical output. When I try to save one to Favorite Queries I get an error:

```sql
mysql root@localhost:mydb> \f acc select created from account where id = 1 \G
(1064, u"You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '\\f acc select created from account where id = 1' at line 1")
```

If I replace the \G with a semicolon it saves fine, but the query has output that is more readable in vertical mode.

Terseus:
Try to fix the issue #203 by adding support for the syntax `\f favorite_query \G`.

I think this is a better feature than saving favorites queries with `\G` since it have the flexibility to execute a favorite query with or without `\G`.

### DIFF
--- a/mycli/sqlexecute.py
+++ b/mycli/sqlexecute.py
@@ -98,15 +98,14 @@ class SQLExecute(object):
             # and then proceed to execute the sql as normal.
             if sql.endswith('\\G'):
                 special.set_expanded_output(True)
-                yield self.execute_normal_sql(sql.rsplit('\\G', 1)[0])
-            else:
-                try:   # Special command
-                    _logger.debug('Trying a dbspecial command. sql: %r', sql)
-                    cur = self.conn.cursor()
-                    for result in special.execute(cur, sql):
-                        yield result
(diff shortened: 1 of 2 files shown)


mdsrosa:
@Terseus I've tested your fix and it works great!  Good job! :sunglasses: 
Thank you for taking the time to develop this! :smile: 

@amjith Any thoughts on this? If not, I'll be happy to merge this to master.

amjith:
This works well. Thank you for the PR.

Terseus:
You're welcome :)

As a side note, I made a small mistake: the comment in https://github.com/dbcli/mycli/blob/master/mycli/sqlexecute.py#L98 it's now obsolete (the query doesn't execute directly as normal anymore), maybe you want to modify it.

Issue #218, opened 2016-01-07
Disable pager


diversario:
It seems like mycli switches from dumping output on the screen to showing it in a pager depending on how many results are there (?). I almost never want the pager – is there a way to force output to never use it?

diversario:
Oh, and also – when the pager is used it destroys whatever there was on the screen before it.

tsroten:
@diversario You're right, it would be nice to have a way of turning the pager off, similar to the official mysql client. I've implemented this and am testing it now.

Regarding the pager destroying what's on the screen. That's annoying to me as well! I'm not sure what the solution to that is, but we can certainly look at it and see what's possible.

tsroten:
@diversario #237 should address the pager destroying what's on screen.

diversario:
Great! Is the version with this change available in `brew`?

tsroten:
@diversario I don't believe so.

@amjith Any idea when the next release will be published?

amjith:
I'd like to upgrade prompt_toolkit to 0.60 and then do a release. 

Hopefully later today.

amjith:
Released as part of 1.6.0. Please upgrade: 

```
pip install -U mycli 
```

Version is not yet available in brew, but a PR has been made and pending approval from brew maintainers. https://github.com/Homebrew/homebrew/pull/50388

PR #219, opened 2016-01-09
Adds nopager command and skip-pager config option.

tsroten:
This pull request adds support for the `nopager` and `\n` commands (that mimic MySQL's client). Issuing either of those commands will disable the pager until the pager is set again with the `pager` or `\P` command.

It also adds support for the `skip-pager` config option in user's `.my.cnf` file. Like expected, this must be specified in the `[client]` section of the file. If `skip-pager` is set, regardless of the value, the pager will be disabled. This mimics MySQL's behavior.

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -525,7 +525,10 @@ class MyCli(object):
                     self.output(str(e), err=True, fg='red')
                 else:
                     try:
-                        self.output_via_pager('\n'.join(output))
+                        if special.is_pager_enabled():
+                            self.output_via_pager('\n'.join(output))
+                        else:
+                            self.output('\n'.join(output))
                     except KeyboardInterrupt:
                         pass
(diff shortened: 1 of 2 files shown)


tsroten:
By the way, this PR is to address #218.

amjith:
:octocat:

horosgrisa:
@amjith Can you add posibility to set own pager?

timfeirg:
is it possible to add nopager into `.myclirc` config files? I didn't see any lines related to pager in http://mycli.net/config @tsroten

tsroten:
@timfeirg Hello! We are planning on adding those options to the myclirc file (when we rework our config-related code). For now, you can it to the file .my.cnf:
```
[client]
skip-pager=on
```

That will turn the pager off. I hope that helps!

Issue #221, opened 2016-01-22
[question] How do I enable local-infile?


daj0ker:
As said in the title, I've edited my.cnf to enable local-infile.

After that, <code>$mysql -u username -p --local-infile</code> does the job.

How do I configure mycli to do this?

tsroten:
@daj0ker We don't currently support the `--local-infile` flag.

It shouldn't be too hard to add, though. mycli uses PyMySQL which simply looks for a boolean argument `local_infile`.

tsroten:
@daj0ker Some initial tests look promising for this. I'll go ahead and submit a pull request for this.

amjith:
Released as part of 1.6.0. Please upgrade: 

```
pip install -U mycli 
```

PR #225, opened 2016-01-31
Adds support for alternate login path file names

tsroten:
This pull request adds support for the environment variable `MYSQL_TEST_LOGIN_FILE`. The MySQL docs [say this about it](http://dev.mysql.com/doc/refman/5.7/en/mysql-config-editor.html):

> To specify an alternate login path file name, set the `MYSQL_TEST_LOGIN_FILE` environment variable. This variable is recognized by **mysql_config_editor**, by standard MySQL clients (**mysql**, **mysqladmin**, and so forth), and by the **mysql-test-run.pl** testing utility.

Basically, it's a way to customize where the login path file is stored. By default, it is `%APPDATA%\MySQL\.mylogin.cnf` on Windows and `~/.mylogin.cnf` on Unix/Linux/OS X.

This pull request also adds tests for detecting both standard and alternate login path files.

### DIFF
--- a/tests/test_login_path.py
+++ b/tests/test_login_path.py
@@ -96,11 +96,15 @@ def test_corrupted_pad():
 
 def test_get_mylogin_cnf_path():
     """Tests that the path for .mylogin.cnf is detected."""
+    original_env = None
     if 'MYSQL_TEST_LOGIN_FILE' in os.environ:
-        del os.environ['MYSQL_TEST_LOGIN_FILE']
+        original_env = os.environ.pop('MYSQL_TEST_LOGIN_FILE')
     is_windows = sys.platform == 'win32'
 
     login_cnf_path = get_mylogin_cnf_path()
+
+    if original_env is not None:
+        os.environ['MYSQL_TEST_LOGIN_FILE'] = original_env
 
     if login_cnf_path is not None:
         assert login_cnf_path.endswith('.mylogin.cnf')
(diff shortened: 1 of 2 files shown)


tsroten:
@mdsrosa - Can you review this?

mdsrosa:
@tsroten Yes, Thomas. I'm on it. :smile:

tests/test_login_path.py
```diff
@@ -90,3 +92,42 @@ def test_corrupted_pad():
...
+            assert login_cnf_path.startswith(home_dir)
+
+
+def test_alternate_get_mylogin_cnf_path():
+    """Tests that the alternate path for .mylogin.cnf is detected."""
+    original_env = None
+    if 'MYSQL_TEST_LOGIN_FILE' in os.environ:
+        original_env = os.environ.pop('MYSQL_TEST_LOGIN_FILE')
+
+    temp_fh, temp_path = tempfile.mkstemp()
```
mdsrosa: I noticed that `temp_fh` is not being used in this test. Since only `temp_path` is needed and `tempfile.mkstemp()` will always return a `tuple`, what do you think about change this line to the following?
`temp_path = tempfile.mkstemp()[1]`

amjith:
@tsroten Will this address https://github.com/dbcli/mycli/issues/221 or is that a separate?

My initial glance seems to suggest that this PR only handles mylogin.cnf and not the my.ini files. Is my assumption right?

tests/test_login_path.py
tsroten: Good point! I've pushed a change for this.

tsroten:
@amjith You're right -- #221 is separate. This PR addresses a wish of mine to be able to move my login path file out of my home directory and into a separate `.config` directory.

tests/test_login_path.py
mdsrosa: @tsroten I think Github don't notify users when there is a update in a comment. As you can see, yesterday I've updated my comment  to `_, temp_path = temp_file.mkstemp()` because is the more _pythonic_ way. If you don't mind, I can push this small change and sorry to bother you with these things.

mdsrosa: @tsroten I've pushed the small change so we can merge your PR. :+1:

mdsrosa:
:+1:

tests/test_login_path.py
tsroten: @mdsrosa I should've paid closer attention. Thanks for pushing the change.

mdsrosa:
@tsroten I'm merging this in. Well done, Thomas! :smile:

PR #226, opened 2016-02-02
Adds support for local-infile option

tsroten:
This pull request adds the `--local-infile` command-line option and support for MySQL's `local-infile` config option.

The default value is `False`. Here is the precedence order:

1. command-line option `--local-infile`
2. config option `local-infile`
3. default value of `False`

The `--local-infile` command-line option takes any value that click accepts as a boolean. The `local-infile` config option accepts any value in this set (case-insensitive): `('true', 'on', '1', 'false', 'off', '0')`.

It is worth noting that this option only works if the server is configured to accept local data.

Also, it supports the `loose-` prefix for the config option.

This addresses #221.

Reviewer: @amjith


tsroten:
And, I broke all the tests :)

mycli/main.py
```diff
@@ -307,18 +309,27 @@ def connect(self, database='', user='', passwd='', host='', port='',
         passwd = passwd or cnf['password']
         charset = charset or cnf['default-character-set'] or 'utf8'
 
+        # Favor whichever local_infile option is set.
+        for local_infile_option in (local_infile, cnf['local-infile'], 
+                                    cnf['loose-local-infile'], False):
```
amjith: I was wondering why the default values for the `local_infile` in the function args is not False. But this logic assumes that if it is not explicitly set to false then we set it to None or empty string so it doesn't prematurely break out of this loop.

I think that's the reasoning as I understand it.

amjith:
Looks good. I haven't tested this with a local data source. But I presume you've done the due diligence. 

:tomato:

mycli/main.py
tsroten: Yep! We need to distinguish unset options from false options.

PR #227, opened 2016-02-02
Refactors pager code

tsroten:
This pull request removes some unneeded code and makes the pager functionality a bit more intuitive.

First, it removes code that "resets" the environment variables `PAGER` and `LESS`. This is not needed because Python's environment variable changes do not affect the global environment.

Second, it adjusts the behavior of the `pager` command to respect the `pager` config once started up. Imagine a `my.cnf` set to use pager `less` and an environment variable `PAGER` set to `more`.

* mycli starts up with `PAGER` set to `less`.
* User runs `nopager`, executes commands, then runs `pager`.
* `PAGER` is now set to `more`.
* User runs `pager less`. Pager now set to `less`.
* User runs `nopager`, executes commands, then runs `pager`.
* `PAGER` is now set to `more`.

This PR updates this so that the set pager does not change on the user.

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -368,8 +368,7 @@ class MyCli(object):
     def run_cli(self):
         sqlexecute = self.sqlexecute
         logger = self.logger
-        original_less_opts = self.adjust_less_opts()
-        self.set_pager_from_config()
+        self.configure_pager()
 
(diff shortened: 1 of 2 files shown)


amjith:
Looks good to me. 

:crown:

Issue #223, opened 2016-01-26
can not use `status` cmd.


jinleileiking:
```
mysql uWFL9Wg6159@10.36.253.127:yingjian_duin> status
(1064, u"You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'status' at line 1")
mysql uWFL9Wg6159@10.36.253.127:yingjian_duin> status;
(1064, u"You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'status' at line 1")
```

tsroten:
@jinleileiking Thanks for using mycli!

mycli doesn't currently support the `status` command. Let me take a look and see if it's something we could add.

jinleileiking:
:+1:

amjith:
Released as part of 1.6.0. Please upgrade: 

```
pip install -U mycli 
```

PR #229, opened 2016-02-06
Adds status command

tsroten:
This addresses #223.

The `status`/`\s` command is implemented by the official MySQL command-line client. It provides information about the server, client, and connection. The implementation offered in this pull request is extremely similar. The differences are detailed below.

Here is some example output from this command (with the IP address scrubbed).
```
mysql troten@db02:(none)> \s
--------------
mycli 1.5.2, running on CPython 3.5.0

Connection id:        9889
Current database:
Current user:         troten@XX-XX-XX-XXX.ded.swbell.net
Current pager:        System default
Server version:       5.6.28-0ubuntu0.14.04.1-log (Ubuntu)
Protocol version:     10
Connection:           db02 via TCP/IP
Server characterset:  latin1
Db characterset:      latin1
Client characterset:  utf8
Conn. characterset:   utf8
TCP port:             3306
Uptime:               9 days 12 hours 52 min 38 sec

Connections: 4  Queries: 51493578  Slow queries: 23333  Opens: 909  Flush tables: 1  Open tables: 672  Queries per second avg: 62.495
--------------

Time: 0.193s
```

A few things to note:

* The first line of the output is information about the client. The MySQL CLI client includes information about the server the client is bundled with (not the connected server). I opted to provide information about mycli and what version/flavor of Python it is running on (e.g. Jython, PyPy, etc.).
* I've left out the information about the delimiter and outfile. The delimiter is not applicable to mycli. The outfile (called audit log in mycli) is applicable, but I couldn't figure out a clean way to detect that file from the `dbcommands.py` module.
* The SSL data has not been included because, as of right now, mycli does not support this.
* The bottom line contains current statistics about the server. I've done away with "Threads" in favor of "Connections" (like MySQL Workbench) because the number of active connections is, to me, a more helpful number than the number of threads.
* I've also renamed the statistic "Questions" to "Queries" so it is less cryptic and truer to the status variable it references on the server.

Here is a detailed description of each line/statistic in the `status` output:

* Client information: mycli's version number, the python implementation, and python version
* [Connection id](http://dev.mysql.com/doc/refman/5.7/en/information-functions.html#function_connection-id): The unique connection (or thread) ID. This is the ID used in the `INFORMATION_SCHEMA.PROCESSLIST` table.
* Current database: the name of the database currently being used, or a blank string if there is none.
* [Current user](http://dev.mysql.com/doc/refman/5.7/en/information-functions.html#function_user): The user name specified when connecting to the server and the client host from which you connected.
* Current pager: The configured pager or output method. If the pager is enabled, this is the value of `os.environ['PAGER']` or `System default`. If the pager is disabled, then `stdout`.
* Server version: The version number and version comment of the server. E.g. `5.7.10 Homebrew` or `5.6.28-0ubuntu0.14.04.1-log (Ubuntu)`
* Protocol version: the version of the client/server protocol used by the server.
* Connection: whether or not the connection is over TCP/IP or a UNIX socket.
* Character sets: The character encoding used at the various stages in the client <--> server connection.
* TCP port (or UNIX socket): Information about the connection. If over TCP, then the port. If via a socket, then the specific socket being used.
* Uptime: the human-readable server uptime.
* Server statistics:
  - [Connections](http://dev.mysql.com/doc/refman/5.7/en/server-status-variables.html#statvar_Threads_connected): The number of currently open connections.
  - [Queries](http://dev.mysql.com/doc/refman/5.7/en/server-status-variables.html#statvar_Queries): The number of statements executed by the server. This includes those executed within stored programs.
  - [Slow queries](http://dev.mysql.com/doc/refman/5.7/en/server-status-variables.html#statvar_Slow_queries): The number of queries that have taken a long time (greater than `long_query_time` seconds).
  - [Opens](http://dev.mysql.com/doc/refman/5.7/en/server-status-variables.html#statvar_Opened_tables): The number of tables that have been opened.
  - [Flush tables](http://dev.mysql.com/doc/refman/5.7/en/server-status-variables.html#statvar_Flush_commands): The number of times the server flushes tables. This could be due to internal server operation or a user calling a `FLUSH TABLES` statement.
  - [Open tables](http://dev.mysql.com/doc/refman/5.7/en/server-status-variables.html#statvar_Open_tables): The number of tables that are open.
  - Queries per second avg.: The total number of statements divided by the uptime in seconds. This number is rounded to 3 decimal places.

Reviewer: @amjith

### DIFF
--- a/mycli/packages/special/utils.py
+++ b/mycli/packages/special/utils.py
@@ -14,3 +14,33 @@ def handle_cd_command(arg):
         return True, None
     except OSError as e:
         return False, e.strerror
+
+def format_uptime(uptime_in_seconds):
+    """Format number of seconds into human-readable string.
+
+    :param uptime_in_seconds: The server uptime in seconds.
+    :returns: A human-readable string representing the uptime.
+
+    >>> uptime = format_uptime('56892')
+    >>> print(uptime)
+    15 hours 48 min 12 sec
+    """
+
+    m, s = divmod(int(uptime_in_seconds), 60)
+    h, m = divmod(m, 60)
+    d, h = divmod(h, 24)
+
+    uptime_values = []
+
+    for value, unit in ((d, 'days'), (h, 'hours'), (m, 'min'), (s, 'sec')):
+        if value == 0 and not uptime_values:
+            # Don't include a value/unit if the unit isn't applicable to
+            # the uptime. E.g. don't do 0 days 0 hours 1 min 30 sec.
+            continue
+        elif value == 1 and unit.endswith('s'):
+            # Remove the "s" if the unit is singular.
+            unit = unit[:-1]
+        uptime_values.append('{} {}'.format(value, unit))
+
+    uptime = ' '.join(uptime_values)
+    return uptime
--- a/mycli/packages/special/utils.py
+++ b/mycli/packages/special/utils.py
@@ -40,7 +40,7 @@ def format_uptime(uptime_in_seconds):
         elif value == 1 and unit.endswith('s'):
             # Remove the "s" if the unit is singular.
             unit = unit[:-1]
-        uptime_values.append('{} {}'.format(value, unit))
+        uptime_values.append('{0} {1}'.format(value, unit))
 
     uptime = ' '.join(uptime_values)
     return uptime
--- a/mycli/packages/special/dbcommands.py
+++ b/mycli/packages/special/dbcommands.py
@@ -4,6 +4,7 @@ import platform
 from mycli import __version__
 from mycli.packages.tabulate import tabulate
 from mycli.packages.special import iocommands
+from mycli.packages.special.utils import format_uptime
 from .main import special_command, RAW_QUERY, PARSED_QUERY
 
 log = logging.getLogger(__name__)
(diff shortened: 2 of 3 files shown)


mycli/packages/special/utils.py
```diff
@@ -14,3 +14,33 @@ def handle_cd_command(arg):
...
+    h, m = divmod(m, 60)
+    d, h = divmod(h, 24)
+
+    uptime_values = []
+
+    for value, unit in ((d, 'days'), (h, 'hours'), (m, 'min'), (s, 'sec')):
+        if value == 0 and not uptime_values:
+            # Don't include a value/unit if the unit isn't applicable to
+            # the uptime. E.g. don't do 0 days 0 hours 1 min 30 sec.
+            continue
```
amjith: I'm not sure if mysql already has this or not, but this is a nice touch. :+1:

amjith:
I like this PR. Well thought out implementation and you've picked reasonable defaults for this feature.

My one question is, why are we printing directly from the `status()` function and returning a tuple of `None`s? Shouldn't it build the array of output rows and return that, so we can print them using `click.echo_via_pager()` in main.py?

tsroten:
@amjith Thanks! Good question about using `click.echo_via_pager()`. I had strayed from this because:

1. This command is essentially a port from MySQL, which prints this status information to stdout.
2. mycli's pager command says "Print the query results via PAGER". I interpreted this command's output as not strictly being a query result.

That being said, I am definitely open to making this output respect the user's pager settings if that's the preferred way of doing things in mycli. What do you think?

amjith:
I'm less concerned about piping the output through the pager (although we should do that anyway), my concern is these functions shouldn't be printing directly to the terminal. We should have a central place (in main.py) where we print to the terminal and everywhere else we should pass around data. We should treat the functions as pure functions without side effects wherever possible. 

The comment about printing query results is probably outdated. It should say printing any output.

tsroten:
@amjith Good point. I'll fix that when I get a chance.

tsroten:
@amjith I made the changes (but haven't committed them). Do you know if there is any way to make the pager not get rid of the on-screen history? If you scroll at all in the pager, it moves to the top of the screen and then the history is not visible anymore.

amjith:
Yes, the ruining of history is because of the -S switch for `less`. I've received multiple complaints about that switch now. Since you've moved the pager to the config file, it might be a good idea to remove the `-S` from that list.

amjith:
@tsroten Any update on this PR?

tsroten:
@amjith Okay, it should be returning the output instead of printing it directly. I'll make another PR to remove the `-S` less option.

amjith:
Looks good. :crown:

PR #230, opened 2016-02-07
Fixes the system command for Python 3

tsroten:
mycli's system command doesn't work under Python 3 because the subprocess returns the response encoded in bytes, not as a string. This results in an error like this:

```
[... Traceback shortened ...]
    self.output_via_pager('\n'.join(output))
TypeError: sequence item 0: expected str instance, bytes found
```

This pull request fixes this to ensure the response is always returned as a string.


tsroten:
And... I broke another system-command-related test. Fixing...

tsroten:
Okay, test is fixed.

amjith:
That is quite the embarrassing bug. Thanks for the fix. 

:microscope:

PR #232, opened 2016-02-15
Adds support for SSL connections

mrdeathless:


amjith:
@mrdeathless Thanks for taking the time to create this PR.

Can you give me examples of usage? I don't personally use this feature and I'd like to see how it is used in the real world.

The tests are failing because of the newly added ssl parameter to the `__init__()` method. Can you update the tests to make them pass again? 

Thanks!

mrdeathless:
> Can you give me examples of usage?
SSL is used to secure a connection to remote MySQL server. This is needed when other means, such as SSH to a server hosting the database, are unavailable. For example, the only way to connect to a [Google Cloud SQL](https://cloud.google.com/sql/) instance is using MySQL protocol and if you don't want to transmit your queries and data as plane text you have to use SSL.

amjith:
@mrdeathless I'm trying to setup a mysql instance with ssl to see if I can test this PR. I'm currently setting it up in a docker container. I'll let you know if I run into issues.

mycli/main.py
```diff
@@ -668,11 +719,20 @@ def cli(database, user, host, port, socket, password, dbname,
     # Choose which ever one has a valid value.
     database = database or dbname
 
+    ssl = {
+        'ca': ssl_ca,
+        'capath': ssl_capath,
+        'cert': ssl_cert,
+        'key': ssl_key,
+        'cipher': ssl_cipher,
```
amjith: Can you add os.path.expanduser() to all the file paths here? 

Otherwise specifying `~/server-ca.pem` will fail.

amjith:
@mrdeathless I was able to test this against a google cloud sql instance. Thanks for that tip, it saved me a ton of time trying to set something up locally. 

One issue I noticed was the pem file paths have to be expanded using the os.path.expanduser() otherwise it fails when I specify the path as ~/Downloads/server-ca.pem.

As soon as you fix that update the PR, I'll merge this in. 

Thanks again for the PR.

mrdeathless:
@amjith Thank you for taking time to review my PR. I've added support for tilde expansion as you asked.

amjith:
I'm going to merge this and fix the expand_path() issue as a separate PR.

Thanks again for taking the time to make mycli better. I appreciate it. :ice_cream:

PR #237, opened 2016-03-01
Don't truncate long lines in less by default

tsroten:
This PR removes the `-S` less option so that the history won't get messed up when scrolling in less. Those that still want the `-S` option can set their pager to `less -S` in their `.my.cnf` file.

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -310,7 +310,7 @@ class MyCli(object):
         charset = charset or cnf['default-character-set'] or 'utf8'
 
         # Favor whichever local_infile option is set.
-        for local_infile_option in (local_infile, cnf['local-infile'], 
+        for local_infile_option in (local_infile, cnf['local-infile'],
                                     cnf['loose-local-infile'], False):
             try:
                 local_infile = str_to_bool(local_infile_option)
@@ -574,7 +574,7 @@ class MyCli(object):
 
     def configure_pager(self):
         # Provide sane defaults for less.
-        os.environ['LESS'] = '-SRXF'
+        os.environ['LESS'] = '-RXF'
 
         cnf = self.read_my_cnf_files(self.cnf_files, ['pager', 'skip-pager'])
         if cnf['pager']:


mycli/main.py
```diff
@@ -310,7 +310,7 @@ def connect(self, database='', user='', passwd='', host='', port='',
         charset = charset or cnf['default-character-set'] or 'utf8'
 
         # Favor whichever local_infile option is set.
-        for local_infile_option in (local_infile, cnf['local-infile'], 
+        for local_infile_option in (local_infile, cnf['local-infile'],
```
tsroten: My editor automatically removes trailing spaces...

amjith:
We should document the availability of `pager` option in the config file.

Would you mind adding it to the default myclirc file and have it commented and set it to `less -RXF` so people know how to set it up for themselves, if needed.

amjith:
Never mind I got confused between pgcli and mycli. pgcli has a pager option in the pgclirc config file. So I got confused.

MrOutis:
Is there a way to enable this again?

amjith:
@MrOutis You can set a pager value in your `~/.my.cnf` file and that will be honored. It has to be specified in the `[client]` sub-section of the file. Eg:

```
[client]
pager = 'less -XSRF'
```

MrOutis:
It worked, @amjith, however, why do we need to use `~/.my.cnf` where `~/.myclirc` is available?

amjith:
At one point there was a discussion about this and came to the conclusion that if both ~/.my.cnf and ~/.myclirc had `pager` it would be hard to break the tie (or at least try to do the right thing) in that case. So we decided not to have conflicting options between my.cnf and .myclirc.

I hope that answers your question.

MrOutis:
Sorry, @amjith, I wasn't clear enough in my question, it was more like: why do we need two configuration files?

tsroten:
@MrOutis `.my.cnf` is a MySQL-designated file for various MySQL-related options. `.myclirc` is for mycli-specific options.

If we only used `.my.cnf`, then we'd lose the ability to expand into other configurable features (e.g. syntax highlighting, coloring, or favorite queries). If we only used `.myclirc`, then we wouldn't be able to honor the [client settings people configure in `.my.cnf`](http://dev.mysql.com/doc/refman/5.7/en/mysql-command-options.html), which is the standard place to set up MySQL clients.

MrOutis:
Thank you, guys, @tsroten, @amjith!

PR #238, opened 2016-03-03
Use expanduser to expand the tilde in the path names.

amjith:
Reviewer: @tsroten 

This is a follow up PR for #232. 

In #232 @mrdeathless added the ability to supply ssl options to mycli. A few of the options were path names to ssl cert files. The path names with tilde must be expanded otherwise it will fail to find the file. 

The original PR added the expanduser() but it ended up overwriting the paths with incorrect values. This PR fixes that issue.

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -720,22 +720,15 @@ def cli(database, user, host, port, socket, password, dbname,
     # Choose which ever one has a valid value.
     database = database or dbname
 
-    ssl_paths = {
-        'ca': ssl_ca,
-        'cert': ssl_cert,
-        'key': ssl_key,
-    }
-    ssl_params = {
-        'capath': ssl_capath,
-        'cipher': ssl_cipher,
-        'check_hostname': ssl_verify_server_cert,
-    }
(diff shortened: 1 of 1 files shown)


mrdeathless:
I apologize for making a blatant mistake with keys versus values in my last comming. Unfortunately, you fix won't work because `os.path.expanduser` will raise `AttributeError` when you pass `None` to it. And this is exactly the value which missing arguments have. This was the reason why I separated paths from other arguments. I intended to walk over supplied paths, hence `if p` check and only try to expand them, avoiding `None` values.

amjith:
@mrdeathless Good catch. I'll add a check.

amjith:
Check added.

tsroten:
@amjith Clever use of the `and` there.

I don't have a Google account and don't have access to a MySQL instance with SSL (at work we access them over SSH). But, the `expanduser()` calls and changes to the dictionary look solid to me.

I assume you've tested this with SSL on a Google Cloud instance, so I'll push it through :)

Issue #233, opened 2016-02-16
COUNT() returns invalid results


diversario:
Using mycli:

```
mysql user@host:dbname> SELECT COUNT(id) FROM dbname.tablename WHERE fieldname = "value";
+---------------+
|   row_count() |
|---------------|
|            -1 |
+---------------+
Query OK, 2499 rows affected
Time: 0.188s
```

Using mysql CLI:

```
[host dbname]> SELECT COUNT(id) FROM dbname.tablename WHERE fieldname = "value";
+-----------+
| COUNT(id) |
+-----------+
|      2499 |
+-----------+
1 row in set (0.12 sec)
```

amjith:
That is awful. Thanks for reporting it. I'll take a look.

tsroten:
@diversario Does this happen on all tables or a particular one?

diversario:
I think this happened on two tables but I'm not 100% sure.

tsroten:
@diversario This is definitely odd behavior.

It would be helpful if you could set the mycli logging level to `debug` and post the relevant portions of the log.

Could you edit your `~/.myclirc` to have `log_level = debug` under the `[main]` section? Also, either set or not what `log_file` is set to. That should make the log more verbose.

tsroten:
@diversario I submitted a related pull request about a week ago. Can you try the latest version of mycli from Github and see if the problem persists? Thanks!

amjith:
You can install directly from master using this command: 

```
pip install -U https://github.com/dbcli/mycli/archive/master.zip
```

amjith:
Fixed as part of 1.6.0. Please upgrade: 

```
pip install -U mycli 
```

PR #239, opened 2016-03-05
Simplify status message generation.

tsroten:
This removes the extra SQL call to `ROW_COUNT()` by using `cur.execute()`'s return value (which is [PEP 0249's `rowcount`](https://www.python.org/dev/peps/pep-0249/#rowcount) -- the number of affected rows) in order display the number of rows produced or affected.

It also simplifies the code that makes `row` single or plural in the status message.

This is also related to #233, where the call to `row_count()` appears to be displayed as the result. I'm not sure if it will fix that problem, though.

Reviewer: @amjith

### DIFF
--- a/mycli/sqlexecute.py
+++ b/mycli/sqlexecute.py
@@ -122,28 +122,19 @@ class SQLExecute(object):
         _logger.debug('Regular sql statement. sql: %r', split_sql)
         cur = self.conn.cursor()
         num_rows = cur.execute(split_sql)
-        title = None
-        if num_rows == 1:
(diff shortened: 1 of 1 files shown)


tsroten:
Oops, darn Python 2.6. Fixing.

amjith:
This works. I wonder why I chose to use the ROW_COUNT() instead of using the return value. Because it looks like I do check the return value in one scenario.

I really wish I had written a comment at the time. Oh well. This works for now and I like PRs that delete lines of code. :rotating_light:

PR #241, opened 2016-03-12
Fix PyMySQL version check.

tsroten:
If you use a version of PyMySQL less than 0.6.7, you get this error (and mycli crashes) when attempting to read binary data:

```
  File "/Users/troten/code/mycli/mycli/main.py", line 536, in run_cli
    if pymysql.VERSION < ('0', '6', '7'):
TypeError: unorderable types: int() < str()
```

This pull request fixes this by comparing the version using integers, not strings. It also makes the test suite skip the appropriate tests if the version doesn't support them.

Reviewer: @mdsrosa


amjith:
:shell:

PR #242, opened 2016-03-12
Remove license meta-data.

tsroten:
This is a very short/unimportant PR :)

[Per the Python docs](https://docs.python.org/3/distutils/setupscript.html#additional-meta-data), the `license` meta-data attribute shouldn't be used in mycli's case:
```The license field is a text indicating the license covering the package where the license is not a selection from the “License” Trove classifiers.```

Mycli uses the license Trove classifier, so we don't need the `license` field.


amjith:
:tractor:

PR #244, opened 2016-03-20
Amjith/upgrade pt

amjith:
This PR upgrades prompt-toolkit to 0.60. 

It also adds a continuation prompt that matches default mysql. If you're in multi-line mode, you can see the continuation prompt on the second line onwards. 

Reviewer: @tsroten


tsroten:
Nice! I like the new continuation prompt. Much clearer than the previous behavior. This looks good to me.

Issue #179, opened 2015-10-23
Unable to pipe input


alasdairdc:
Not sure if this already fixed in master or not.  But using Debian Jessie and version 1.4.0 from the repo, attempting to pipe input into mycli 

```bash
$ zcat ~/create-test-db.sql.gz | mycli 
```

complains with: 

```bash
/usr/share/python/mycli/local/lib/python2.7/site-packages/mycli/main.py:565: Warning: Click detected the use of the unicode_literals __future__ import.  This is heavily discouraged because it can introduce subtle bugs in your code.  You should instead use explicit u"" literals for your unicode strings.  For more information see http://click.pocoo.org/python3/
  @click.argument('database', default='', nargs=1)
Version: 1.4.0
Chat: https://gitter.im/dbcli/mycli
Mail: https://groups.google.com/forum/#!forum/mycli-users
Home: http://mycli.net
Thanks to the contributor - Iryna Cherniavska
Traceback (most recent call last):
  File "/usr/local/bin/mycli", line 9, in <module>
    load_entry_point('mycli==1.4.0', 'console_scripts', 'mycli')()
  File "/usr/share/python/mycli/local/lib/python2.7/site-packages/click/core.py", line 700, in __call__
    return self.main(*args, **kwargs)
  File "/usr/share/python/mycli/local/lib/python2.7/site-packages/click/core.py", line 680, in main
    rv = self.invoke(ctx)
  File "/usr/share/python/mycli/local/lib/python2.7/site-packages/click/core.py", line 873, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/share/python/mycli/local/lib/python2.7/site-packages/click/core.py", line 508, in invoke
    return callback(*args, **kwargs)
  File "/usr/share/python/mycli/local/lib/python2.7/site-packages/mycli/main.py", line 590, in cli
    mycli.run_cli()
  File "/usr/share/python/mycli/local/lib/python2.7/site-packages/mycli/main.py", line 347, in run_cli
    document = cli.run()
  File "/usr/share/python/mycli/local/lib/python2.7/site-packages/prompt_toolkit/interface.py", line 270, in run
    with self.input.raw_mode():
  File "/usr/share/python/mycli/local/lib/python2.7/site-packages/prompt_toolkit/input.py", line 65, in raw_mode
    return raw_mode(self.stdin.fileno())
[...]

PR #246, opened 2016-03-22
Add stdin batch mode.

tsroten:
This pull request adds support for reading commands from stdin. It works similar to the mysql client. If you pipe commands to mycli like this:
```bash
cat test.sql | mycli
```
or this
```bash
mycli < test.sql
```

Then, mycli will output the results in a plain, tab-separated format (no padding or status messages). If you use the `-t`/`--table` argument, then mycli will output the results in your configured table format.

After processing stdin, mycli will quit.

I can also see use for a yet-to-be-added (different PR) verbose mode that will add support for displaying the commands that are running as well as possibly status messages.

This addresses #179.

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -861,13 +861,13 @@ def is_destructive(queries):
 def confirm_destructive_query(queries):
     """Check if the query is destructive and prompts the user to confirm.
     Returns:
-    None if the query is non-destructive.
+    None if the query is non-destructive or we can't prompt the user.
     True if the query is destructive and the user wants to proceed.
     False if the query is destructive and the user doesn't want to proceed.
     """
     prompt_text = ("You're about to run a destructive command.\n"
                    "Do you want to proceed? (y/n)")
-    if is_destructive(queries):
+    if is_destructive(queries) and sys.stdin.isatty():
         return click.prompt(prompt_text, type=bool)
 
 def quit_command(sql):
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -777,13 +777,17 @@ def cli(database, user, host, port, socket, password, dbname,
         if (mycli.destructive_warning and
                 confirm_destructive_query(stdin_text) is False):
             exit(0)
-        results = mycli.sqlexecute.run(stdin_text)
-        for result in results:
-            title, cur, headers, status = result
-            table_format = mycli.table_format if table else None
-            output = format_output(title, cur, headers, None, table_format)
-            for line in output:
-                click.echo(line)
+        try:
+            results = mycli.sqlexecute.run(stdin_text)
+            for result in results:
+                title, cur, headers, status = result
+                table_format = mycli.table_format if table else None
+                output = format_output(title, cur, headers, None, table_format)
+                for line in output:
+                    click.echo(line)
+        except Exception as e:
+            click.secho(str(e), err=True, fg='red')
+            exit(1)
 
 def format_output(title, cur, headers, status, table_format, expanded=False, max_width=None):
     output = []
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -703,6 +703,8 @@ class MyCli(object):
               help='Only read default options from the given file')
 @click.option('--auto-vertical-output', is_flag=True,
               help='Automatically switch to vertical output mode if the result is wider than the terminal width.')
+@click.option('-t', '--table', is_flag=True,
+              help='Display batch output in table format.')
 @click.option('--local-infile', type=bool,
               help='Enable/disable LOAD DATA LOCAL INFILE.')
 @click.option('--login-path', type=str,
@@ -711,7 +713,8 @@ class MyCli(object):
 def cli(database, user, host, port, socket, password, dbname,
         version, prompt, logfile, defaults_group_suffix, defaults_file,
         login_path, auto_vertical_output, local_infile, ssl_ca, ssl_capath,
-        ssl_cert, ssl_key, ssl_cipher, ssl_verify_server_cert):
+        ssl_cert, ssl_key, ssl_cipher, ssl_verify_server_cert, table):
+
     if version:
         print('Version:', __version__)
         sys.exit(0)
@@ -747,6 +750,17 @@ def cli(database, user, host, port, socket, password, dbname,
             '\thost: %r'
             '\tport: %r', database, user, host, port)
 
+    stdin = click.get_text_stream('stdin')
+    if not stdin.isatty():
+        results = mycli.sqlexecute.run(stdin.read())
+        for result in results:
+            title, cur, headers, status = result
+            table_format = mycli.table_format if table else None
+            output = format_output(title, cur, headers, None, table_format)
+            for line in output:
+                click.echo(line)
+        exit(0)
+
     mycli.run_cli()
 
 def format_output(title, cur, headers, status, table_format, expanded=False, max_width=None):
@@ -757,7 +771,7 @@ def format_output(title, cur, headers, status, table_format, expanded=False, max
         headers = [utf8tounicode(x) for x in headers]
         if expanded:
             output.append(expanded_table(cur, headers))
-        else:
+        elif table_format is not None:
             rows = list(cur)
             tabulated, frows = tabulate(rows, headers, tablefmt=table_format,
                 missingval='<null>')
@@ -767,6 +781,10 @@ def format_output(title, cur, headers, status, table_format, expanded=False, max
                 output.append(expanded_table(rows, headers))
             else:
                 output.append(tabulated)
+        else:
+            output.append('\t'.join(headers))
+            for row in cur:
+                output.append('\t'.join([str(r) for r in row]))
     if status:  # Only print the status if it's not None.
         output.append(status)
     return output
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -202,6 +202,12 @@ class MyCli(object):
                 query = f.read()
         except IOError as e:
             return [(None, None, None, str(e))]
+
+        if (self.destructive_warning and
+            confirm_destructive_query(query) is False):
+            message = 'Wise choice. Command execution stopped.'
+            return [(None, None, None, message)]
+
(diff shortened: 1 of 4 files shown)


tsroten:
Looks like Python 2.6 tests are failing. I'll work on fixing those.

tsroten:
Here are examples of the default and table outputs:


```
$ mycli < ~/test.sql
count(*)
1428461
first_name      middle_initial
alsdkf  None
fhaaaa  None
haaaa   None
Kevin
haa     None
ha      None
foo     None
haaa    None
None    None
Thomas
```

```
$ mycli < ~/test.sql
+------------+
|   count(*) |
|------------|
|    1428461 |
+------------+
+--------------+------------------+
| first_name   | middle_initial   |
|--------------+------------------|
| alsdkf       | <null>           |
| fhaaaa       | <null>           |
| haaaa        | <null>           |
| Kevin        |                  |
| haa          | <null>           |
| ha           | <null>           |
| foo          | <null>           |
| haaa         | <null>           |
| <null>       | <null>           |
| Thomas       |                  |
+--------------+------------------+
```

mdsrosa:
@tsroten Well done! That's an awesome feature! I would be happy to test this out and merge this in.

tsroten:
@mdsrosa @amjith Looks like the test is failing in Python 2.6 because this is the first test that actually runs the mycli app (using click's `CliRunner`). This particular pull request doesn't break Python 2.6 support. It looks like [this commit did](https://github.com/dbcli/mycli/commit/ba96b11674ddd8c110f8d391c353d1d2d6d7199f). Our other tests don't initialize mycli, so this issue hasn't popped up before. The `main.initialize_logging` function calls `logging.captureWarnings`, which was added in 2.7.

Is there a particular reason we still advertise support for 2.6? It's been officially retired for 2.5 years now.

tsroten:
@mdsrosa By the way, thanks for the enthusiastic compliment! I'd appreciate any thoughts/advice you might have if you notice anything.

amjith:
Hi @tsroten, the reason for supporting Python 2.6 is because Centos 6.x and RHEL 6.x ships with Python 2.6.x as system defaults. Installing a newer version of Python in those systems is not straight forward. So I chose to support 2.6 unless it becomes a pain to do so. 

Can you add a conditional around that statement? Such as:

```
if hasattr(logging, 'captureWarnings'):
    logging.captureWarnings(True)
```

That should allow us to continue support 2.6

I understand it takes a bit more effort to support 2.6, but I'd like to be as accommodating as possible. In this case the fix is as simple as adding a single line, but if it takes significant amount of work, I'll be happy to drop 2.6 support.

amjith:
I should add, I'm very excited about this feature. This has been requested multiple times and I'm glad to see you taking an initiative to implement this. 

I wonder if we should take special care to turn off interactive features such as prompting the user before dropping tables.

tsroten:
@amjith That makes sense. Ok, I fixed the Python 2.6 support.

I'll take a look at disabling the interactive features as well. That's a good idea!

mdsrosa:
@tsroten Yes, I'm very excited about this feature.

Here's a thought: Would this work for remote scripts as well? For example:
`$ mycli < ssh://user@xxx.xxx.xxx:/home/script.sql`

@amjith I'm a little concerned about disabling interactive security features.
I once was saved by `mycli` when I ran a unintended `DROP TABLE` command. So, maybe we could add an parameter like `--i-am-feeling-lucky` (just kidding, we could use another name for this) to run the script without prompting.

What do you guys think?

tsroten:
@amjith @mdsrosa Actually, after looking, I see that no warnings/prompts are thrown at the user, even if the query is destructive. Same with the `source` command that was previously implemented. Both features call `sqlexecute.run()` directly.

I think the default for stdin and the source command should be to *not* prompt for destructive queries. One of the big reasons people run files to begin with, is that they've prepared the queries in advance. So, the queries aren't being run on a whim. I'm open to an option similar to [mysql's `--safe-updates`/`--i-am-a-dummy`](http://dev.mysql.com/doc/refman/5.5/en/mysql-command-options.html#option_mysql_safe-updates), but for destructive queries in stdin.

If we add this option, we'll want to scan the entire stdin for destructive queries in advance, before processing any of the queries. Otherwise the user might be prompted to halt the query after some `update` statements (or something similar) have been run. The database would be left in a state that is halfway through the script.

What are your thoughts?

mdsrosa:
@tsroten Thank you for clarifying your thoughts, you made a good point. I agree with you on not prompting the user in those cases (running `source` command  or `stdin`).

I also agree that we need to warn the user about the destructive commands. We can say like `Destructive commands were found in script.sql. Do you want to proceed? [y/n]`. And I also think that this should be the _default_ behavior, because maybe the script is being executed on a production database, so the user can get a chance to review the script before running it. If the script is going to be executed on a testing database, then the option `--ignore-warning` (or something similar) could be used. 

What do you think?

tsroten:
@mdsrosa The thing with `stdin` is that it will often be run as part of a chain of commands (through piping inputs), in cron jobs, in background threads, etc. If we make warning the default, users will have to remember to set the command-line flag for not warning on destructive queries.

All that being said, I see what you're saying and I'm happy with your idea, as long there is a config option for turning the warnings off.

`source` is a bit different, because it requires user interaction to even source a script (e.g. it's not easy to use `source` as part of a cron job), so it makes sense to include the warnings there by default. But, the same config option could be used for `source` as well.

mdsrosa:
@tsroten Yes, I now see the problem with warning on destructive queries by default on `stdin`. Thank you for clarifying that. :)

Adding a config option to turn the warnings off sounds like a good idea. I wonder how @amjith feels about this. 

By the way, I agree on the `source` command having warnings by default. ;-)

amjith:
Interesting discussion. 

I agree that we need a command line switch to turn off the warnings.

Regarding the source command, I think we should honor the current config option called [`destructive_warning`](https://github.com/dbcli/mycli/blob/master/mycli/myclirc#L17). But if you'd like to treat that as a separate issue that's fine too.

tsroten:
Ok, so to put the ideas together, how about we use `destructive_warning` for stdin, `source`, and interactive mode -- that way users know what to expect. And, that won't involve adding another config option.

As far as a command-line switch, we could try `--warn`/`--no-warn`. That way, the user could negate whatever the config file says. The switch would work for stdin, `source`, and interactive mode.

What do you think of that?

amjith:
Yup. That adds a good consistent configuration interface. I like it.

mdsrosa:
That's sound great! I'm on board! :smile:

tsroten:
@mdsrosa @amjith I can't find a clean way to read stdin from a pipe and stdin from a tty (interactive prompt). I'm open to suggestions for how to handle it. The *nix-only way is to open up `/dev/tty`, but that isn't a very clean solution.

Right now, if I try to prompt the user for confirmation about a `drop`/`delete` statement, click raises an `Abort` exception because the `stdin` already hit EOF and isn't reading from the tty.

The only thing I can think of is to use [`click.getchar()`](http://click.pocoo.org/5/api/#click.getchar). But, I don't really like that solution.

Any ideas?

tsroten:
We could attempt to prompt the user and accept input via `/dev/tty`, which would work fine for most of our user base. If it fails, we'd simply continue without the prompt (e.g. on windows). So, the prompt would be a safety feature that is only supported on certain OS's.

mycli/main.py
```diff
@@ -841,24 +841,34 @@ def is_select(status):
         return False
     return status.split(None, 1)[0].lower() == 'select'
 
+def query_starts_with(query, prefixes):
+    """Check if the query starts with any item from *prefixes*."""
+    prefixes = [prefix.lower() for prefix in prefixes]
+    formatted_sql = sqlparse.format(query.lower(), strip_comments=True)
```
tsroten: This also makes it so `confirm_destructive_query` will work even with comments. Before, if you tried this command:
```sql
# comment
drop database foo;
```
You would not have been prompted. Now, you will.

amjith: Thank you!! Nice catch.

amjith:
I'm testing the behavior right now. Here are my 2 cents on this matter. If we detect that we can't open tty, what if we refuse to run a dangerous command. This might be overly conservative so feel free to contradict.

The behavior I'm suggesting is, if someone pipes in a file and we encounter a dangerous command we're unable to prompt the user then we quit and say "Please turn off the `destructive_warning` option in your config file to proceed". 

Actually that seems a bit over the top. I'll submit the comment just because I've written it, but I think I'm leaning towards not recommending my suggestion above.

amjith:
The stdin piping works as expected. I checked the destructive command and that also works great (tested only on OS X). 

A few things that didn't work as expected:

1. If I launch mycli as usual in the interactive mode and when I want to quit using ^D or `exit` it doesn't quit right away, it is waiting for user input, I believe it is related to the latest addition of waiting for user input (https://github.com/dbcli/mycli/pull/246/files#diff-2fdbd8483579d3d1af50361d19c0b8e6R762). 

2. If I encounter an error while running commands from a file the output is a python stack trace. For example I added `SELECT abc` which is a non-sensical statement and it printed a python stack trace. 

3. If I try to drop an non-existent table it prints the python stack trace (same as above).

tsroten:
Re: number 1 - I'll fix that. For some reason I thought that `run_mycli` called `exit`.

Re: number 2 and 3 -- Those are on my todo list :)

tsroten:
This is a note that a useful addition to this feature, although it's out of the scope of this PR, would be displaying the line number of the statement that threw an error. This would require us to loop through the queries via `sqlparse.split()`, while keeping track of how many lines each query took up in the input. Plus, `sqlparse.split()` doesn't output newline(s) between statements, so that complicates things.

tsroten:
@amjith @mdsrosa Ok, I think I hit the major points from our discussion. Anything else stand out?

amjith:
Works like a charm. Thank you!

🌺

mdsrosa:
@tsroten @amjith I'm coming a little late but great work, guys! Well done! 👍

PR #261, opened 2016-05-06
Upgrade to prompt-toolkit 1.0.0

jonathanslenders:
A few changes for being compatible with the latest prompt-toolkit.

The changelog: https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/CHANGELOG

The most important change is that the `vi_state` parameter for `KeyBindingManager` isn't used anymore. (It's ignored.) Instead, we have `CommandLineInterface.editing_mode`.


amjith:
👍 

@dbcli/mycli-core Can you guys take it for a spin? 

In my testing things seem to work fine.

PR #268, opened 2016-05-29
less_chatty config option

scrappysoft:
I appreciate the contact URLs and giving credit to supporters and such, and friendly goodbye messages, but I prefer to suppress unwanted noise from appearing in my terminal window. 

The "less_chatty" option allows the user in _~/.myclirc_ to suppress these messages.

The default remains False, so that the messages will continue to appear as they do today unless explicitly suppressed.  By removing the corresponding change to the boilerplate _myclirc_ file, this could be an undocumented feature, if desired.

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -61,6 +61,11 @@ Query = namedtuple('Query', ['query', 'successful', 'mutating'])
 
 PACKAGE_ROOT = os.path.dirname(__file__)
 
+# no-op logging handler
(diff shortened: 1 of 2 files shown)


amjith:
I'm going to  reluctantly merge this PR. 

I like saying thanks to the contributors and saying Goodbye at the end adds a bit of character to the program. But I understand that not everyone is thrilled by this prospect.

🚵

PR #270, opened 2016-06-07
CSV support

gubbin:
Add --csv flag, which is like the --table flag but formats output in Excel-compatible CSV format.


amjith:
Looks like the tests are failing in Python 3. It's because of the absence of StringIO module in Python 3. 

Would you mind taking a look at that? 

Also we don't typically bump the version in `__init__.py` unless we're getting ready for a release.

caffeinewriter:
Taking a quick look at Python 3, it looks like `StringIO` has been moved to `io.StringIO`.

https://docs.python.org/3/library/io.html

mdsrosa:
@gubbin I'm closing this because there is a new PR for implementing this feature by following your idea. https://github.com/dbcli/mycli/pull/322

Thank you very much!

PR #272, opened 2016-06-09
add prompt_continuation support

scrappysoft:
Add ability to configure the continuation prompt string.

It supports the same expansion as **prompt**, so that you can have values like:
```
prompt = "\d > "
prompt_continuation = "\d*> "
```

In this example, the full prompt is repeated, adding an asterisk on continuation lines.

More importantly, it allows the prompt continuation string to be empty.  

```
prompt_continuation = ""
```

This makes it possible to copy multi-line queries to the clipboard without the injection of prompt characters, or having to resort to launching an editor just to get to the pure query text.

It defaults to "-> ", so the default behavior is unchanged.


amjith:
👍

Issue #271, opened 2016-06-08
display login-path within prompt


jmjpro:
I saved a DB connection to mysql config like this:

`mysql_config_editor set --login-path=qa --host=qa.domain.com --user=mysql_user --password`

Now I connect using mycli like this:

`mycli --login-path qa`

Can I configure mycli to use the login-path `qa` in the prompt? The issue is that the hostnames for AWS are really long.

amjith:
Can you elaborate a little bit about the request? When you say "use `qa` in the prompt", do you mean show `user@qa:dbname>` or something else? 

Also the title of the issue said something about using a nickname in the config file. I don't understand that request.

jmjpro:
@amjith, I changed the issue title. You're right, I like to display `user@qa:dbname`

amjith:
This might not happen right away, but I'll keep this in the back log to implement. Thanks!

johnksterling:
Hi - this feature has been merged and works great for me with the PR from j-bennet.  You can probably close this issue.

j-bennet:
Yes, this is available since 1.8.0. Closing.

PR #275, opened 2016-06-13
Display login-path instead of host in prompt. Connect #271.

j-bennet:
Implements request from #271.


amjith:
I was debating whether this should be a config option or a new prompt specifier such as `\L` which will use the login-path if it exists or default to the host name. But I think the config option should be fine. 

⛵

Issue #285, opened 2016-07-11
`timestamp` column shows `None` if value is `0`


georgexsh:
Original mysql command tool show `0000-00-00 00:00:00`, but mycli shows `None` (vertical mode) or `<null>` (horizontal mode).
It has confused me for a while because that column is not nullable.

mdsrosa:
@georgexsh Thank you for using mycli. But in this case I was unable to reproduce this behavior `0000-00-00 00:00:00` in mysql command tool. Could you please tell me which version of `MySQL` and `mycli` are you using?

georgexsh:
`mysql  Ver 14.14 Distrib 5.5.47` and `mycli Version: 1.7.1`

georgexsh:
```
mysql> desc ts_null;
+-------+-----------+------+-----+-------------------+-----------------------------+
| Field | Type      | Null | Key | Default           | Extra                       |
+-------+-----------+------+-----+-------------------+-----------------------------+
| c     | timestamp | NO   |     | CURRENT_TIMESTAMP | on update CURRENT_TIMESTAMP |
+-------+-----------+------+-----+-------------------+-----------------------------+
1 row in set (0.00 sec)

mysql> insert ts_null values (0);
Query OK, 1 row affected (0.01 sec)

mysql> select * from ts_null;
+---------------------+
| c                   |
+---------------------+
| 0000-00-00 00:00:00 |
+---------------------+
1 row in set (0.00 sec)
```
vs
```
mysql root@localhost:test> select * from ts_null;
+--------+
|      c |
|--------|
| <null> |
+--------+
1 row in set
Time: 0.001s

mysql root@localhost:test> select * from ts_null\G
***************************[ 1. row ]***************************
c | None

1 row in set
Time: 0.000s
```

amjith:
I'm able to reproduce it, I'll see if I can take a stab at the fix.

georgexsh:
@amjith thanks very much

amjith:
This is now fixed in version 1.8.0. Please upgrade mycli: 

```
pip install -U mycli
```

PR #290, opened 2016-07-20
Custom converters for time/date types.

amjith:
Having invalid time/date values in database causes mycli to display
their values as <null>. Having custom converters will handle that case.

Reviewer: @j-bennet 

This is to address #285

### DIFF
--- a/mycli/sqlexecute.py
+++ b/mycli/sqlexecute.py
@@ -2,6 +2,9 @@ import logging
 import pymysql
 import sqlparse
 from .packages import connection, special
+from pymysql.constants import FIELD_TYPE
+from pymysql.converters import (convert_mysql_timestamp, convert_datetime,
+        convert_timedelta, convert_date)
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


amjith:
@tsroten Can you please review?

tsroten:
@amjith Sure! Looking now.

mycli/sqlexecute.py
```diff
@@ -61,13 +64,19 @@ def connect(self, database=None, user=None, password=None, host=None,
             '\tlocal_infile: %r'
             '\tssl: %r',
             database, user, host, port, socket, charset, local_infile, ssl)
+        conv = {
+                FIELD_TYPE.TIMESTAMP: lambda obj: (convert_mysql_timestamp(obj) or '0000-00-00 00:00:00'),
```
tsroten: Depending on the MySQL server version (and how old the data is), there could be invalid dates/times such as `2007-02-31` which will return `None` from PyMySQL's converters. Instead of hard-coding the fallback value as `0000-00-00...`, how about we just return `obj` so that the original value from the database is shown?

amjith: Didn't think about that. I'll give that a shot. 

Good catch!

amjith:
Updated the PR. It works great. 😄 

Thanks for the suggestion.

amjith:
merging.

Issue #294, opened 2016-08-01
Completion engine throws exception


leunggamciu:
I ran into an issue that the completion engine throwed an exception when it runs for the first time and it stop working anymore after exception thrown。

call stack:

```
mysql root@localhost:(none)> showException in thread Thread-5:
Traceback (most recent call last):SHOW     
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()                             
  File "/usr/lib/python2.7/threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/usr/local/lib/python2.7/dist-packages/prompt_toolkit/interface.py", line 835, in run
    completions = list(buffer.completer.get_completions(document, complete_event))
  File "/usr/local/lib/python2.7/dist-packages/mycli/sqlcompleter.py", line 255, in get_completions
    suggestions = suggest_type(document.text, document.text_before_cursor)
  File "/usr/local/lib/python2.7/dist-packages/mycli/packages/completion_engine.py", line 85, in suggest_type
    full_text, identifier)                 
  File "/usr/local/lib/python2.7/dist-packages/mycli/packages/completion_engine.py", line 134, in suggest_based_on_last_token
    token_v = token.value.lower()          
AttributeError: 'tuple' object has no attribute 'value'
```

I am running version 1.7.1 on my Ubuntu 16.04.1 machine，and it was installed via pip。

lastly, thanks for binging us such an awesome tool!

leunggamciu:
refer to #291

PR #295, opened 2016-08-01
Pin sqlparse to 0.1.19

amjith:
Latest version of sqlparse (0.20.0) is breaking everything. I'm going to pin the dependency to 0.1.19 until we fix all the issues associated with the upgrade. 

/cc @mdsrosa 

Closes: #294


mdsrosa:
👍

Issue #305, opened 2016-08-08
Autocomplete double


Vakhrushev:
double only after `show`, common autocomplete work fine

![imgur-2016_08_08-14 58 21](https://cloud.githubusercontent.com/assets/692146/17473122/f879f642-5d78-11e6-82cf-877a2eccfc32.png)

`mycli -v                                                                                                                
Version: 1.8.0
`

`
Linux UWC 4.6.4-1-ARCH #1 SMP PREEMPT Mon Jul 11 19:12:32 CEST 2016 x86_64 GNU/Linux
`

amjith:
This is now fixed in master. I'll do a release soon. In the meantime you can install directly from master as follows: 

```
pip install -u https://github.com/dbcli/mycli/archive/master.zip
```

johnksterling:
Looks like this issue can be closed

tsroten:
@Vakhrushev This should be fixed in mycli 1.8.1. Thanks for reporting this!

PR #306, opened 2016-08-09
Reset the show items when completion is refreshed.

amjith:
This addresses #305 

Reviewer: @j-bennet


j-bennet:
Verified that it works.

👍

PR #312, opened 2016-09-18
Amjith/sqlparse upgrade

amjith:
@meeuw After I merged your PR I noticed that the completion engine would crash after the keyword `AS`. This PR addresses that issue. There is already a bug filed with sqlparse to handle that case correctly, so this try/except block is merely a temporary hack before it is fixed in sqlparse. 

I've pinned the version of sqlparse to 0.2.0. The reason being we've been burned in the past by specifying the dependency as `>=0.1.x`.

Reviewer: @tsroten

### DIFF
--- a/mycli/packages/completion_engine.py
+++ b/mycli/packages/completion_engine.py
@@ -28,26 +28,32 @@ def suggest_type(full_text, text_before_cursor):
 
     identifier = None
 
-    # If we've partially typed a word then word_before_cursor won't be an empty
-    # string. In that case we want to remove the partially typed string before
-    # sending it to the sqlparser. Otherwise the last token will always be the
-    # partially typed string which renders the smart completion useless because
-    # it will always return the list of keywords as completion.
-    if word_before_cursor:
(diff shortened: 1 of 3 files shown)


meeuw:
@amjith good catch, I never use "as" so I didn't noticed.

Fedora has sqlparse-0.2.1 packaged so it would be nice to have sqlparse >= 0.2.0,<0.2.2

I couldn't find the sqlparse issue because it was accidentally closed. This is the reference:
https://github.com/andialbrecht/sqlparse/issues/284

amjith:
Good idea to set `>=0.2.0, <0.2.2`, I'll make that change. Looks like they've fixed the issue in sqlparse, just waiting for a release.

tsroten:
@amjith Works great! I'll go ahead and merge this, then we can update later once `sqlparse` releases a new version.

From what you've seen, is this only present in `sqlparse` >= 2.0.0?

amjith:
> From what you've seen, is this only present in sqlparse >= 0.2.0?
 
Yes. When sqlparse released 0.2.0 they made some major changes to handle unicode and it broke compatibility.

PR #314, opened 2016-09-18
is_group() function changed to attribute

meeuw:


amjith:
Looks like the tests are failing. I'll take a look at it later today, but if you have time before that can you check the tests to see what's going on?

meeuw:
this fixes the tests but I don't understand why is_group is a method during the tests...

mycli/packages/parseutils.py
```diff
@@ -64,6 +64,9 @@ def last_word(text, include='alphanum_underscore'):
 # This code is borrowed from sqlparse example script.
 # <url>
 def is_subselect(parsed):
+    if isinstance(parsed, sqlparse.sql.Token):
+        if not parsed.is_group:
+            return False
     if not parsed.is_group():
```
tsroten: @meeuw Thanks for helping out with mycli! Is this call to `is_group()` still supposed to be in here?

meeuw: I'm not sure; I need to test it. If it's changed to is_group without () the tests fail... So somehow the unit tests are hitting this code.

mdsrosa:
@meeuw Hey, thanks for helping out mycli! We really appreciate it! 
Did you have any chance to debug the code @tsroten mentioned?

Thanks again! 😄

PR #317, opened 2016-10-14
Release as a universal wheel

adamchainz:
By releasing as a [Python wheel](http://pythonwheels.com/) as well as a source distribution, you can speed up end user’s installs. After merging this command, to release you just need to run `python setup.py clean sdist bdist_wheel upload`.


amjith:
👍

Issue #255, opened 2016-04-13
auto-vertical-output option in config file


sauloperez:
I just found the `auto_vertical_output` in the readme and it's something that it's going to be very useful for me. What I was wondering is if it's possible to add that option into my `~/.myclirc` or it's only a cli argument. According to my quick search on the codebase, looks like the latter.

amjith:
Right now this option is not available via the config file. But it should be easy to add.

sauloperez:
Cool! I'll try to take a look over the next few days

mdsrosa:
Working on it.

johnksterling:
This works for me.  Can this issue be closed?

j-bennet:
Yes, when mycli is released with the change.

tsroten:
@sauloperez 

mycli version 1.9.0 is now out which has this change in it. You can install/upgrade by running `pip install --upgrade mycli`

Let us know if you run into any other issues, thanks!

PR #321, opened 2016-10-29
Add config option for auto-vertical-output in myclirc

mdsrosa:
This PR addresses #255.

Reviewers: @amjith

### DIFF
--- a/mycli/myclirc
+++ b/mycli/myclirc
@@ -65,7 +65,8 @@ less_chatty = False
 # Use alias from --login-path instead of host name in prompt
 login_path_as_host = False
 
-# Enable auto-verical-output
+# Cause result sets to be displayed vertically if they are too wide for the current window,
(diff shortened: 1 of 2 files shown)


mycli/myclirc
```diff
@@ -65,6 +65,9 @@ less_chatty = False
 # Use alias from --login-path instead of host name in prompt
 login_path_as_host = False
 
+# Enable auto-verical-output
```
amjith: A little bit more description in terms of what this actually does would be nice. Or a link to the documentation for this option. Such as https://dev.mysql.com/doc/refman/5.5/en/mysql-command-options.html#option_mysql_auto-vertical-output

mdsrosa: @amjith Good idea! Will do it. Thanks for the reference, man! 😄

amjith:
👍

PR #322, opened 2016-10-30
Add CSV support for batch output

mdsrosa:
This PR is a follow up on the PR https://github.com/dbcli/mycli/pull/270. This PR add CSV support for batch output feature.

```
$ cat query.txt | mycli -u root test_db --csv
id,email,name
1,email1@gmail.com,Foo Bar
2,emal2@gmail.com,Bar Foo
3,null,John Doe
```

Thanks @gubbin!

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -702,15 +702,14 @@ class MyCli(object):
         string = string.replace('\\n', "\n")
         return string
 
-    def run_query(self, query, table_format=None):
+    def run_query(self, query, table_format=None, new_line=True):
         """Runs query"""
         results = self.sqlexecute.run(query)
         for result in results:
             title, cur, headers, status = result
-            table_format = self.table_format if table_format else None
             output = format_output(title, cur, headers, None, table_format)
             for line in output:
-                click.echo(line)
+                click.echo(line, nl=new_line)
 
 @click.command()
 @click.option('-h', '--host', envvar='MYSQL_HOST', help='Host address of the database.')
@@ -827,7 +826,6 @@ def cli(database, user, host, port, socket, password, dbname,
                 confirm_destructive_query(stdin_text) is False):
             exit(0)
         try:
-            results = mycli.sqlexecute.run(stdin_text)
             table_format = None
             new_line = True
 
@@ -837,12 +835,8 @@ def cli(database, user, host, port, socket, password, dbname,
             elif table:
                 table_format = mycli.table_format
 
-            for result in results:
-                title, cur, headers, status = result
-                output = format_output(title, cur, headers, None, table_format)
-                for line in output:
-                    click.echo(line, nl=new_line)
-
+            mycli.run_query(stdin_text, table_format=table_format, new_line=new_line)
+            exit(0)
         except Exception as e:
             click.secho(str(e), err=True, fg='red')
             exit(1)
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -867,11 +867,10 @@ def format_output(title, cur, headers, status, table_format, expanded=False, max
             output.append(content.getvalue())
             content.truncate(0)
             for row in cur:
-                row = ['null' if val is None else val.encode('utf-8') for val in row]
+                row = ['null' if val is None else str(val) for val in row]
                 writer.writerow(row)
                 output.append(content.getvalue())
                 content.truncate(0)
-            output.append('\n')
             content.close()
         else:
             rows = list(cur)
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -805,7 +805,12 @@ def cli(database, user, host, port, socket, password, dbname,
     #  --execute argument
     if execute:
         try:
-            mycli.run_query(execute, table_format=table)
+            table_format = None
+            if table:
+                table_format = mycli.table_format
+            elif csv:
+                table_format = 'csv'
+            mycli.run_query(execute, table_format=table_format)
             exit(0)
         except Exception as e:
             click.secho(str(e), err=True, fg='red')
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -16,7 +16,7 @@ from io import open
 
 # support StringIO for Python 2 and 3
 try:
-    from StringIO import StringIO
+    from cStringIO import StringIO
 except ImportError:
     from io import StringIO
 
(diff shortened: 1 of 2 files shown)


mdsrosa:
Working on the tests right now.

mdsrosa:
Okay. Tests fixed. 

Could you please review it? @amjith @tsroten

tests/test_main.py
```diff
@@ -88,6 +88,22 @@ def test_batch_mode_table(executor):
...
+    run(executor, '''create table test(a text, b text)''')
+    run(executor, '''insert into test (a, b) values('abc', 'def'), ('ghi', 'jkl')''')
+
+    sql = 'select * from test;'
+
+    runner = CliRunner()
+    result = runner.invoke(cli, args=CLI_ARGS + ['--csv'], input=sql)
+
+    expected = 'a,b\nabc,def\nghi,jkl\n'
+    result_output = result.output.replace('\x00', '') # python 3
```
mdsrosa: @amjith @tsroten I'm not very happy with this "hacky" solution for the test. Do you have any idea on this?

amjith: I'm not sure why the `\x00` is introduced. I wonder if this is a bug in click's CliRunner.

mycli/main.py
```diff
@@ -832,7 +856,23 @@ def format_output(title, cur, headers, status, table_format, expanded=False, max
         headers = [utf8tounicode(x) for x in headers]
         if expanded:
             output.append(expanded_table(cur, headers))
-        elif table_format is not None:
+        elif table_format is None:
+            output.append('\t'.join(headers))
+            for row in cur:
+                output.append('\t'.join([str(r) for r in row]))
```
tsroten: I think you can use our tabulate library for this. It has a `tsv` format. It looks like if you were to set `table_format` to `'tsv'` if it is `None`, then the `else:` statement below will take care of this code.

mycli/main.py
```diff
@@ -13,6 +14,12 @@
 from random import choice
 from io import open
 
+# support StringIO for Python 2 and 3
+try:
+    from StringIO import StringIO
```
tsroten: For Python2, we can get better performance if we do `from cStringIO import StringIO`.
https://docs.python.org/2.6/library/stringio.html#module-cStringIO

tsroten:
@mdsrosa I'm glad we're finally getting CSV support!!

For the actual CSV generation -- I think the various truncate/getvalue calls are leading to an exponential amount of data in memory (it keeps overwriting previous rows with null bytes, which still take up memory). Instead, we could get by with a single `output.append(content.getvalue())` after the loop that appends *both* header and data to the output. Then we can drop the `truncate` commands.

mycli/main.py
```diff
@@ -818,7 +827,22 @@ def cli(database, user, host, port, socket, password, dbname,
                 confirm_destructive_query(stdin_text) is False):
             exit(0)
         try:
-            mycli.run_query(stdin_text, table_format=table)
```
tsroten: We now have two sets of code for running a query -- one that continues to use `run_query` and one that uses this version. It looks like the big difference is the newline handling. Is there any way we could continue using the `run_query` function, but also remove the extra newline? Maybe a helper function that looks at the settings to determine if we should output a newline or something?

mdsrosa:
```
@mdsrosa I'm glad we're finally getting CSV support!!

For the actual CSV generation -- I think the various truncate/getvalue calls are leading to an exponential amount of data in memory (it keeps overwriting previous rows with null bytes, which still take up memory). Instead, we could get by with a single output.append(content.getvalue()) after the loop that appends both header and data to the output. Then we can drop the truncate commands.
```

@tsroten Oh didn't know that it is how it behaves. Thank you for that knowlodge. I will try removing those extra calls to `truncate` and `getvalue()`.

mycli/main.py
mdsrosa: @tsroten You're right. I'll refactor this out. Maybe adding a new parameter called `new_line` to the `run_query` function?

tests/test_main.py
mdsrosa: @tsroten You were right. I removed the `truncate` calls and the `\x00` took vacation. 😄 
Thanks, man!

mycli/main.py
mdsrosa: Done.

mycli/main.py
mdsrosa: @tsroten Yeah, turns out you're were on this one too. I've set the default value as `tsv` for the `table_format`. Thanks, man!

tsroten:
Looking good! I think that this might have affected the `--execute` command's output. When I run it, it doesn't use my `table_format`.

mdsrosa:
@tsroten I think you found a bug and I think I know the fix for it.

https://github.com/dbcli/mycli/blob/124ca067713a4896a398e25e2367989fbc192cc7/mycli/main.py#L808

```
if execute:
        try:
            mycli.run_query(execute, table_format=table)
```

Here it must be `mycli.run_query(execute, table_format=mycli.table_format`.  Correct?

tsroten:
@mdsrosa I think that's the right spot. I believe that would ignore the absence of the `--table` argument then. It should only use `mycli.table_format` if the `table` argument is True. I think that's why the previous code passed in `table` -- then if `table` was true it used `mycli.table_format`.

mdsrosa:
@tsroten Okay. So I should add a check for `--table` and if it's `True` I'll use `mycli.table_format` otherwise `table_format=None`.

tsroten:
If we move it down to the `csv` logic we risk running into problems with the `isatty` and `stdin` code. I wonder if we could just do:
```python
if execute:
    try:
      mycli.run_query(execute, table_format=table_format if table else None)
```

mdsrosa:
@tsroten Hmm. Sorry, could you please elaborate on the risks? I'm asking this so I can learn about more them. By the way, I ran the tests and they're passing. 😄

tsroten:
It looks like the way you did it was fine. :)

My comment about it being risky was referring to bringing the `--execute` code down to around line 843 where the CSV code was. I didn't realize you meant you were going to also add CSV support to the `--execute` command.

*If* the `--execute` code was brought down below the `isatty` check, then in most scenarios, you could never use `--execute` because mycli would run like normal. But, I think you know that since you were the one to add the `--execute` code in the correct spot to begin with ;)

mdsrosa:
@tsroten Oh, sorry for being totally clear about what I meant to do there.
So yeah, the `--execute` command is safe. Do you think this is ready to merge? 🙂

Issue #300, opened 2016-08-03
prompt setting in .my.cnf ignored


dbolser-ebi:
I set my prompt like this:
prompt=\n\u@\h:\p (\d)\n>\_

What I really like about that is the final \n that starts my query on a new line. Otherwise a big prompt makes your SQL craped up on the right.

Can you honor the prompt setting in ~/.my.cnf?

Version: 1.5.2

amjith:
mycli will only honor the [client] section in my.cnf. So if you have the prompt listed in the [client] section it will do the right thing.

dbolser-ebi:
Thanks for this tip, I had my prompt settings under the [mysql] section of ~/.my.cnf. It is now being picked up (and I'm now on version 1.8.0 of mycli), however, it isn't being formatted like in the regular client. This is what I have set currently:
```
[client]
prompt=\n\D\n\u@\h:\p (\d)\n>\_
```

and it gets rendered like this in the regular client (mysql  Ver 14.14 Distrib 5.1.73, for redhat-linux-gnu (x86_64) using readline 5.1):
```
Thu Aug  4 09:01:24 2016
ensro@mysql-eg-prod-1.ebi.ac.uk:4238 (test)
> SELECT ...
```

which is great, but it looks like this in mycli (1.8.0):
```
\D
ensro@mysql-eg-prod-1.ebi.ac.uk:\p (test)
>\_SELECT ...
```
which isn't ideal!

Obviously I don't need such a fancy prompt, but I would have thought mycli just relies on whatever mysql uses, rather than trying to do something else with the prompt?


Many thanks,
Dan.

amjith:
@dbolser-ebi mycli is a completely new client written from ground up. It does not share code with the mysql's default client. So we had to recreate the behavior of mysql wherever it made sense. But you have to remember that mysql is at least 20 years old and we started less than a year ago. So there is a long way to go. 

The difference I see between your mysql prompt and mycli's prompt is that we don't support `\D`, `\p` and `\_`. These are easy things to add support for, in fact you can add them in yourself and send us a pull request. 

Here's the code that deals with prompts: 
https://github.com/dbcli/mycli/blob/master/mycli/main.py#L688-L696

I'll be happy to provide you with any guidance you might need to add support for the missing features. 😄 

Cheers!

dbolser-ebi:
Thanks man!

mdsrosa:
@dbolser-ebi https://github.com/dbcli/mycli/pull/323 😉

[...]

PR #323, opened 2016-10-30
Add prompt support for \D, \p and \_ options

mdsrosa:
This PR address the prompt support issues related in https://github.com/dbcli/mycli/issues/300.

I only add those 3 for now because those are the ones mentioned in the https://github.com/dbcli/mycli/issues/300. If the need for other options appears, I'll be happy to add them (or invite the user to add them).

Reviewers: @tsroten @j-bennet

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -6,6 +6,7 @@ import os
 import os.path
 import sys
 import traceback
+import socket
 import logging
 import threading
 from time import time
@@ -693,6 +694,9 @@ class MyCli(object):
         string = string.replace('\\d', sqlexecute.dbname or '(none)')
         string = string.replace('\\t', sqlexecute.server_type()[0] or 'mycli')
         string = string.replace('\\n', "\n")
+        string = string.replace('\\D', datetime.now().strftime('%a %b %d %H:%M:%S %Y'))
+        string = string.replace('\\p', socket.gethostbyname(socket.gethostname()))
+        string = string.replace('\\_', ' ')
         return string
 
     def run_query(self, query, table_format=None):
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -695,7 +695,7 @@ class MyCli(object):
         string = string.replace('\\t', sqlexecute.server_type()[0] or 'mycli')
         string = string.replace('\\n', "\n")
         string = string.replace('\\D', datetime.now().strftime('%a %b %d %H:%M:%S %Y'))
-        string = string.replace('\\p', socket.gethostbyname(socket.gethostname()))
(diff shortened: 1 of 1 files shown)


dbolser-ebi:
Awesome, many thanks for this!

May I ask that the prompt setting be read from the [mysql] section of the config, rather than the [client] section of the config (or even a [mycli] section?), as it messes with other client utilities such as mysqlcheck and friends.

Should this be a different ticket?

mycli/main.py
```diff
@@ -693,6 +694,9 @@ def get_prompt(self, string):
         string = string.replace('\\d', sqlexecute.dbname or '(none)')
         string = string.replace('\\t', sqlexecute.server_type()[0] or 'mycli')
         string = string.replace('\\n', "\n")
+        string = string.replace('\\D', datetime.now().strftime('%a %b %d %H:%M:%S %Y'))
+        string = string.replace('\\p', socket.gethostbyname(socket.gethostname()))
```
mdsrosa: @amjith You're correct. I misread the MySQL documentation. Stopped at "The current TCP/IP" 😆 . Thanks! Will fix this.

mdsrosa:
@dbolser-ebi 

Sorry if I misunderstood this, but do you mean that the `prompt` config should be read from the `my.cnf` instead of the `myclirc` file? If so, that sounds reasonable. And yes that would be another ticket. Would you mind creating one requesting that change, please?

tsroten:
@dbolser-ebi Reading the prompt from `my.cnf` is definitely doable, but we wouldn't want to read it from the `[mysql]` section. That section is supposed to only be for the official MySQL command-line client's use. Third party clients, like mycli, are supposed to use the `[client]` section. Mycli-specific options generally live in `.myclirc`.

amjith:
We currently read the ~/.my.cnf file but only honor the `[client]` section. If the prompt is defined in both ~/.my.cnf as well as ~/.myclirc we give preference to ~/.my.cnf. 

I think we should only honor the [client] sub-section because that's what clients are supposed to use.

mdsrosa:
I'm merging this PR. 

@dbolser-ebi Do you have any other questions about this? If so, please, feel free to send a message to [our Gitter channel](https://gitter.im/dbcli/mycli). Thank you!

dbolser-ebi:
Many thanks for this @mdsrosa ! How do I get the master version via pip?

j-bennet:
`pip install -U https://github.com/dbcli/mycli/archive/master.zip`

dbolser-ebi:
Hey! Speedy reply! Thanks so much!

dbolser-ebi:
Please close #300 now if you like?

Issue #302, opened 2016-08-05
Support python-sqlparse 0.2.0


lvht:
mycli 1.8.0 could not work with python-sqlparse 0.2.0.

``` python
Traceback (most recent call last):
  File "/usr/lib64/python3.5/threading.py", line 914, in _bootstrap_inner
    self.run()
  File "/usr/lib64/python3.5/threading.py", line 862, in run
    self._target(*self._args, **self._kwargs)
  File ".../lib/python3.5/site-packages/prompt_toolkit/interface.py", line 836, in run
    completions = list(buffer.completer.get_completions(document, complete_event))
  File ".../mycli/mycli/sqlcompleter.py", line 246, in get_completions
    suggestions = suggest_type(document.text, document.text_before_cursor)
  File ".../mycli/mycli/packages/completion_engine.py", line 85, in suggest_type
    full_text, identifier)
  File ".../mycli/mycli/packages/completion_engine.py", line 134, in suggest_based_on_last_token
    token_v = token.value.lower()
AttributeError: 'tuple' object has no attribute 'value'
```

meeuw:
I have the same error on Fedora 24, the tuple contains an integer.
dnf downgrade python3-sqlparse-0.1.18-5.fc24
works around this.

meeuw:
I've bisected sqlparse and found this commit breaks mycli
https://github.com/andialbrecht/sqlparse/commit/c4954af15bbcb93a61d111a218c2d9ba2aa7ea3a

amjith:
@meeuw Thank you for taking the time to bisect and find the error. The latest release of sqlparse (version 0.2.0) made some significant changes that are breaking mycli. So I've chosen to pin the version of required sqlparse to 0.1.19. 

I haven't found the time to upgrade mycli to adapt to the changes introduced in 0.2.0. 

Sorry for the inconvenience. I hope to get it compatible eventually, but it's just taking a while.

tsroten:
Closing this issue since it was released in version 1.8.1.

PR #324, opened 2016-11-05
Bump sqlparse version

darikg:
So mycli and pgcli can be installed together. #302

I don't have a mysql server handy to test this one, but the tests pass.


amjith:
:joystick:

Issue #284, opened 2016-07-11
Auto reconnect when connection is lost


georgexsh:
Original mysql command line tool will try to reconnect automatically if connection is lost, but mycli will not.
What if offer an option to auto reconnect?

Version: 1.7.1

jialongl:
https://github.com/jialongl/mycli/commit/05a13b7bb327c379d1c43d932ac6ed67426ba255

diversario:
I'm on 1.8.1 and I see no auto reconnect. Is it need to be enabled somehow?

amjith:
This feature is not yet released. If you'd like to use it (and provide some feedback) you can install it directly from master

```pip install -U https://github.com/dbcli/mycli/archive/master.zip```

diversario:
I installed it, so far works well.

tsroten:
@georgexsh @diversario @jialongl

mycli version 1.9.0 is now out which should fix this. You can install/upgrade by running `pip install --upgrade mycli`

Let us know if you run into any other issues, thanks!

PR #331, opened 2016-12-20
Fixes #202 and #284

### LINKED ISSUE #202: Boring Connection reset.
Again, me     :)

vipcenter> select \* from member;
Connection reset. Reconnect (Y/n): Y
Reconnected!
Try the command again.

Why my connection is always be reset, should I change some mysql setting?

jialongl:


amjith:
@jialongl Thanks for the PR. I'm checking it out now.

amjith:
This works great. Thank you for the contribution. I'm still reading through the code to make sure it's all good. I'll merge it in once I'm satisfied. It should be later today.

jialongl:
Thanks.

In summary, the commit
- moved a big code block into a function `one_iteration` (thus changing some `continue`s to `returns`) with one single argument `document` (defaults to `None`);
- moved the initilizations of `document` var into an `if` block when `document` is `None`.
- calls `one_iteration(document)` (recursion) when the need to reconnect arises. A `return` is added after the recursion call.

amjith:
I like it. 👍 

Thank you! Thank you! 🚀

jialongl:
Glad that my quick hack worked in your framework.
Thanks for taking time to review and merge. :)

amjith:
> Thanks for taking time to review and merge. :)

My pleasure. In fact, I made a couple of false starts on the reconnect issue, but my approaches were quite complicated and convoluted. So I'm glad you found a way to do this in a simple yet elegant manner.

PR #335, opened 2017-01-24
Honor the smart_completion setting in the config file.

### LINKED ISSUE #334: "smart_completion = False" doesn't work
Having `smart_completion = False` in `~/.myclirc` and launch `mycli`, the text `[F2] Smart Completion: OFF` in greyed-out style shows at the bottom of the terminal --- normal.

However, at the end of the line it shows `Refreshing Completions ...` and when the message disappears, greyed-out `[F2] Smart Completion: OFF` becomes bright `[F2] Smart Completion: ON`.

Please confirm and advise.

amjith:
Reviewer: @tsroten

This fixes #334 

The smart_completion setting in the config file was ignored by the background refresher thread. This PR makes the CompletionRefresher class aware of that setting.

### DIFF
--- a/mycli/completion_refresher.py
+++ b/mycli/completion_refresher.py
@@ -12,9 +12,10 @@ class CompletionRefresher(object):
 
     refreshers = OrderedDict()
 
-    def __init__(self):
+    def __init__(self, smart_completion):
         self._completer_thread = None
         self._restart_refresh = threading.Event()
+        self.smart_completion = smart_completion
 
     def refresh(self, executor, callbacks):
         """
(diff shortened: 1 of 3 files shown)


tsroten:
@amjith Looks like some good improvements here!

When I run this with `smart_completion = False` in my `.myclirc`, the toolbar displays "Refreshing completions..." indefinitely.

It looks like in `mycli.main.MyCli.run_cli()`, we attempt to refresh the completions no matter what the config is set to.

amjith:
@tsroten Sorry about stomping on your work. 😞 

With smart_completion turned off we still need to refresh completions. The reason being, smart_completion only controls whether we try to be context sensitive or not. So a "dumb" completion would try to suggest everything in its arsenal for every location of the query. 

At least that's how I had imagined it when I introduced that option. But I can see why you might want to disable completions altogether (for really large databases). So I'm going to retract my PR and have you submit your PR. 

We can basically assume that you either get completions which are smart or you don't get any completions at all. Which seems like the intuitive thing to do. Thanks for tackling this. :)

PR #337, opened 2017-01-26
Honor smart_completion setting.

tsroten:
Fixes https://github.com/dbcli/mycli/issues/334.

**Reviewer**: @amjith


amjith:
Looks great!

Thanks for taking this approach. It's clearly a better alternative to mine.

PR #338, opened 2017-01-29
Fix #333 ctrl-c issue

### LINKED ISSUE #333: ctrl+c did not kill the query
After `ctrl+c`, the client told me the query is cancelled.
But actually, I can still see the query I just cancelled in `show processlist`.

I am using mycli 1.8.0 on Mac OS X, mysql server is v5.5

chainkite:
kill the query by `connection_id` (`thread_id`) instead of `document.text`

### DIFF
--- a/mycli/sqlexecute.py
+++ b/mycli/sqlexecute.py
@@ -41,6 +41,7 @@ class SQLExecute(object):
         self.local_infile = local_infile
         self.ssl = ssl
         self._server_type = None
+        self.connection_id = None
         self.connect()
 
     def connect(self, database=None, user=None, password=None, host=None,
@@ -90,6 +91,8 @@ class SQLExecute(object):
         self.socket = socket
         self.charset = charset
         self.ssl = ssl
+        # retrieve connection id
+        self.reset_connection_id()
 
     def run(self, statement):
         """Execute the sql in the database and return the results. The results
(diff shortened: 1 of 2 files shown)


amjith:
Reviewing the PR now.

mycli/sqlexecute.py
```diff
@@ -221,3 +224,16 @@ def server_type(self):
...
+        if not self.connection_id:
+            self.reset_connection_id()
+        return self.connection_id
+
+    def reset_connection_id(self):
+        # Remember current connection id
+        _logger.debug('Get current connection id')
+        res = self.run('select connection_id()')
+        for title, cur, headers, status in res:
+            self.connection_id = list(cur)[0][0]
```
amjith: Instead of converting it to a list, can you use `fetchone()` to fetch the first row of the results and then index into the first column? Something like: 

```
self.connection_id = cur.fetchone()[0]
```

bjarnagin:
@chainkite @amjith I believe you can also use PyMySQL's Connection.close() method to drop the connection to the server, which would eliminate needing to do a roundtrip to get the connection id.

chainkite:
@bjarnagin I tried `sqlexecute.conn.close()` doesn't really kill the query.

amjith:
Thanks @chainkite. Appreciate the contribution. 👍

amjith:
@bjarnagin Previously mycli was closing the connection when Ctrl-C was pressed and I throught (wrongly) that it would cancel the query as well. But it turns out that wasn't the case.

PR #339, opened 2017-02-06
Add always expand option.

AGhost-7:
This allows the user to specify to always expand output instead of
having to specify "\G" all the time.

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -114,6 +114,7 @@ class MyCli(object):
         config_files = ([self.default_config_file] + self.system_config_files +
                         [self.user_config_file])
         c = self.config = read_config_files(config_files)
+        self.expanded = c['main'].as_bool('expanded')
         self.multi_line = c['main'].as_bool('multi_line')
         self.key_bindings = c['main']['key_bindings']
         special.set_timing_enabled(c['main'].as_bool('timing'))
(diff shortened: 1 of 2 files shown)


amjith:
@AGhost-7 Thanks for the PR. 

I like the approach. I was wondering about a couple of things.

1. Did you see this command line option:

```
--auto-vertical-output        Automatically switch to vertical output mode
                                if the result is wider than the terminal
                                width.
```

What do you think about adding that as a config option? 

2. Is there a genuine reason for doing expanded mode all the time instead of having it `auto_expand` based on the size, as suggested above?

AGhost-7:
Not really any reason - wasn't aware that there was such an option. I'll close this and try to make a different PR some other time.

amjith:
You didn't have to close this one, you can always adjust this PR to do auto_expand.

Looking forward to your changes. 😄 

Let me know if you'd like some help to do it.

AGhost-7:
Well, isn't this already in the codebase? 

https://github.com/dbcli/mycli/blob/master/mycli/main.py#L131

https://github.com/dbcli/mycli/blob/master/mycli/myclirc#L70

amjith:
You're correct. I'm sorry I misled you. I forgot that it was implemented by @mdsrosa.

AGhost-7:
No worries, I should've taken a better look before implementing this.

tsroten:
Just a note that always-expanded output will be an option once #388 is merged in. This will be accomplished by setting `table_format` to `expanded` (in `myclirc`) or by changing the table type to `expanded` once mycli is running.

Issue #341, opened 2017-02-08
Startup time causes password entry to display in terminal


AlecHansen:
I'm not sure if this warrants a bug report, maybe this is a feature request? And it might not be a trivial problem, or related to Python (in that case likely unsolvable) but I'm using this as a replacement for the mysql terminal application. 

I immediately start entering the database password into the command line after hitting enter. 
With the Mysql application the startup time is such that (at least on my terminal) it's impossible to accidentally type a few characters of the password into the terminal window, but with mycli the startup takes a half second longer or so to being capturing keyboard entry, and so I consistently enter the first few characters of the DB password as plaintext into the terminal window. If I mash the keyboard I can consistently enter 12 characters before the password capture starts, in regular use it's more like 3. 

I can take a look at the code later, but if off the top of a devs head this is unsolvable just close this issue.

tsroten:
@AlecHansen I think there are ways to speed it up. For instance, I believe the official mysql client only asks you for a password if you use the `-p` flag (without a value). Mycli has the `-p` flag, but it's only used to enter a password in plaintext.

Without the `-p` flag, mycli attempts to login without a password -- if that fails, then the user is asked for a password. I imagine that's the reason for the delay. I could see an argument to replicate the mysql behavior of prompting the user immediately if the `-p` flag is used without a value.

On a related note, I'm not sure if you're aware that mycli supports [mysql's encrypted login path file](https://dev.mysql.com/doc/refman/5.7/en/option-files.html). That way you aren't prompted for your password, but it's not stored in plaintext.

Anyone else have any thoughts?

amjith:
>  I could see an argument to replicate the mysql behavior of prompting the user immediately if the -p flag is used without a value.

That's a good suggestion. I'm not sure why we didn't emulate that behavior from mysql. I'll be happy to merge a PR that implements this suggestion. :)

johnksterling:
I don't think 'click' options can have variable nargs, though there was a PR for that project to implement that which has not been accepted.  So I made a small PR that adds a parameter to force entering the password via the prompt --password-prompt.  Is that a good middle ground?

amjith:
@johnksterling I didn't see your comment here. It's unfortunate that click can't accept empty values for an option. 

I just started reading click's docs and I noticed that the prompt option is what we want. 

http://click.pocoo.org/5/options/#prompting

What do you think?

johnksterling:
[...]

PR #346, opened 2017-02-15
#341:  Password prompt prior to attempting to connect if the password…

johnksterling:
… is not defaulted from any of the places.

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -378,9 +378,6 @@ class MyCli(object):
         passwd = passwd or cnf['password']
         charset = charset or cnf['default-character-set'] or 'utf8'
 
-        if passwd is None:
-            passwd = getpass.getpass()
-
         # Favor whichever local_infile option is set.
         for local_infile_option in (local_infile, cnf['local-infile'],
                                     cnf['loose-local-infile'], False):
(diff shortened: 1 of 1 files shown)


amjith:
@johnksterling Thank you for the PR. What do you think about using [click's password input](http://click.pocoo.org/5/options/#password-prompts) for this instead of using `getpass()`? 

@tsroten I know you had planned on using the `-p` with an empty value to default to a password prompt, what are your thoughts on this approach?

I'd prefer to match MySQL's default behavior if possible.

tsroten:
Do we think that people are relying on providing their password via visible plaintext in the command-line? In other words, will people's scripts break if we use the password input option?

Or, perhaps that doesn't matter because we'd like to get rid of plaintext, visible password entry altogether, in which case the password prompt option works great (similar to mysql's).

Another option would be to convert the current password option to a [prompted option](http://click.pocoo.org/5/options/#prompting) that accepts input via the command-line or user input after it's run.

Personally, I think the password prompt option (where you cannot provide it via the command-line) is the right way to go. A few people might protest if they've hooked up mycli to scripts that add the password via the command-line, but in the long run it's probably the right move.

Issue #348, opened 2017-02-21
disable align (space padding) for table format tsv


meeuw:
Steps to reproduce:

```
$ mycli
Version: 1.8.1
Chat: https://gitter.im/dbcli/mycli
Mail: https://groups.google.com/forum/#!forum/mycli-users
Home: http://mycli.net
> \T tsv
Changed table Type to tsv
Time: 0.000s
> select 1, 2
  1       2
  1       2
1 row in set
Time: 0.001s
```

Now in the result `1` and `2` are suffixed by two spaces (for alignment), I'm working on a fix

PR #350, opened 2017-02-21
FIX: disable align (space padding) for table format tsv #348

meeuw:
See issue #348


amjith:
Thanks for the fix @meeuw. 👍

j-bennet:
@amjith I was also looking at tabulate repo, and there's been some fixes that look useful to us since we branched the package (when was it? 2015?), for example:

https://bitbucket.org/astanin/python-tabulate/pull-requests/29/optionally-disable-number-parsing/diff

I'll see if I can port those over.

Issue #91, opened 2015-07-31
Add ability to export tabular data from within mycli


mriehl:
Hi,

I've been using mycli for a few days and I really like it so far. A feature I'd love to see is being able to export tabular data to JSON or CSV, this would allow me to actually get rid of mysql workbench.
Is something similar planned or in the works?
I can probably find some time to implement the feature myself if it's desired.

amjith:
I did not have that feature planned in mind. But it will be a great feature to have. If you have the time to implement it, I'd love to take in the contribution. 

Let me know if you need some help getting started with the dev setup. 

Here's a starter:
http://mycli.net/develop

brewneaux:
Id also like this - I'm going to see if I can hack it together this weekend, probably as a pager implementation.

amjith:
@brewneaux If you'd like to take a stab at it, go for it. I'm swamped with other bug reports so I might not get to this for a while.

brewneaux:
@amjith @mriehl An update: I got started on this.  Just a few more bugs to work out.  Hopefully later this week.

amjith:
Thank you for the update. I look forward to the PR. :)

tsroten:
@brewneaux Thanks for taking a stab at this! Any progress?

tsroten:
Ok, I'd like to implement this one as I would use it *a lot*. The biggest hurdle I'm seeing is the interface for this feature.

We could do format-specific commands, e.g.:

- `\csv foo.csv select * from foo;`
- `\json foo.json select * from foo;`
- `\xlsx foo.xlsx select * from foo;`

Or, we could do a single command that takes a format, filename, and SQL command (`\o` or `\x`)

- `\o csv foo.csv select * from foo;`
- `\o json foo.json select * from foo;`
- `\o xlsx foo.xlsx select * from foo;`

@amjith @mdsrosa Any other ideas?

amjith:
@tsroten There's actually a current PR pending for pgcli that implements something similar. 

https://github.com/dbcli/pgcli/pull/483

My preference would be this: 

1. Setup the output format using `\T csv`, then setup the output file using `\o foo.csv` and everything after that will go into the file. This will be used for longer sessions. 

2. For a temporary run where I want to just dump the output of a single command, I could use: \csv foo.csv select * from foo;

What do you think?

tsroten:
@amjith I like that idea.

[...]

PR #353, opened 2017-02-25
unix like redirect output

meeuw:
## Description
Hi, could you please give me some feedback on this PR? (It might be a fix for issue #91)
I need to write some unit tests to improve the code coverage.

Output from a query can be redirected like:
mysql root@172.17.0.2:(none)> select 1 > /tmp/output

Output from a query can be appended like:
mysql root@172.17.0.2:(none)> select 2 >> /tmp/output

You can also use spaces for redirection like:
mysql root@172.17.0.2:(none)> select 1 > "/tmp/output 123"

And even escape quotes:
mysql root@172.17.0.2:(none)> select 1 > "/tmp/\\"weird\\""

After a query the redirection is automatically disabled. The complete output is written to a file, including the "1 row in set"

## Checklist

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -466,6 +466,7 @@ class MyCli(object):
                 document = self.cli.run()
 
                 special.set_expanded_output(False)
+                special.set_redirected_output(enabled=False)
 
                 # The reason we check here instead of inside the sqlexecute is
                 # because we want to raise the Exit exception which will be
@@ -580,6 +581,7 @@ class MyCli(object):
             else:
                 try:
                     special.write_tee('\n'.join(output))
+                    special.output_to_file('\n'.join(output))
                     if special.is_pager_enabled():
                         self.output_via_pager('\n'.join(output))
                     else:
--- a/mycli/packages/special/iocommands.py
+++ b/mycli/packages/special/iocommands.py
@@ -282,3 +282,65 @@ def write_tee(output):
         tee_file.write(output)
         tee_file.write(u"\n")
         tee_file.flush()
+
+
+@export
+def set_redirected_output(append=False, filename=None, enabled=False):
+    global use_redirected_output
+    use_redirected_output = {'append': append,
+                             'filename': filename, 'enabled': enabled}
+
+
+@export
+def get_redirected_output():
+    return use_redirected_output
+
+
+@export
+def redirectmatch(sql):
+    filename = None
+    if sql and sql[-1] in ("'", '"'):  # find filename in quoted redirect
+        quote = sql[-1]
+        m = re.match(r'.*([^\\]' + quote + r'.*' + quote + '$)', sql)
+        if m:
+            filename = m.group(1).replace('\\' + quote, quote)
+    else:  # find filename for non-quoted redirect
+        quote = ''
+        m = re.match(r'.*?([^\][^>]*$)', sql)
+        if m:
+            filename = m.group(1)
+
+    if filename is None:
+        redirect = None
+    else:
+        redirect = sql[:-len(filename)].strip()
+
+    if redirect and (redirect[-2:] == r'\>' or redirect[-3:] == r'\>>'):
+        append = redirect[-3:] == r'\>>'
+
+        sql = redirect[:-3] if append else redirect[:-2]
+        sql = sql.strip()
+        filename = filename.strip()
+
+        if quote:
+            filename = filename[len(quote):-len(quote)]
+        return (sql, {
+            "enabled": True,
+            "filename": filename,
+            "append": append,
+        })
+    else:
+        return (sql, {
+            "enabled": False,
+            "filename": None,
+            "append": False,
+        })
+
+
+@export
+def output_to_file(text):
+    output = get_redirected_output()
+    if output["enabled"]:
+        mode = 'a' if output['append'] else 'w'
+        with open(output["filename"], mode, encoding='utf-8') as f:
+            f.write(text)
--- a/changelog.md
+++ b/changelog.md
@@ -8,6 +8,11 @@ Internal Changes:
 * Move AUTHORS and SPONSORS to mycli directory. (Thanks: [Terje Røsten] []).
 * Add pager wrapper for behave tests (Thanks: [Dick Marinus]).
 
+Features:
+---------
+
+* Add unix like redirect output (Thanks: [Dick Marinus]).
+
 1.10.0:
 =======
 
--- a/mycli/packages/special/main.py
+++ b/mycli/packages/special/main.py
@@ -105,5 +105,7 @@ def show_keyword_help(cur, arg):
 @special_command('quit', '\\q', 'Quit.', arg_type=NO_QUERY)
 @special_command('\\e', '\\e', 'Edit command with editor. (uses $EDITOR)', arg_type=NO_QUERY, case_sensitive=True)
 @special_command('\\G', '\\G', 'Display results vertically.', arg_type=NO_QUERY, case_sensitive=True)
+@special_command('redirect', 'query \> filename', 'Write the query results to filename')
+@special_command('redirectappend', 'query \>> filename', 'Append the query results to filename')
 def stub():
     raise NotImplementedError
--- a/mycli/sqlexecute.py
+++ b/mycli/sqlexecute.py
@@ -123,6 +123,8 @@ class SQLExecute(object):
             if sql.endswith('\\G'):
                 special.set_expanded_output(True)
                 sql = sql[:-2].strip()
+            sql, redirect = special.redirectmatch(sql)
+            special.set_redirected_output(**redirect)
             try:   # Special command
                 _logger.debug('Trying a dbspecial command. sql: %r', sql)
                 cur = self.conn.cursor()
--- a/test/features/fixture_data/help_commands.txt
+++ b/test/features/fixture_data/help_commands.txt
@@ -1,27 +1,29 @@
-+-------------+-------------------+---------------------------------------------------------+
-| Command     | Shortcut          | Description                                             |
-+-------------+-------------------+---------------------------------------------------------+
-| \G          | \G                | Display results vertically.                             |
-| \dt         | \dt [table]       | List or describe tables.                                |
-| \e          | \e                | Edit command with editor. (uses $EDITOR)                |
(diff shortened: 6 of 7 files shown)


tsroten:
I really like this idea! I would definitely use this. I'll post some feedback on some of the code.

mycli/packages/special/iocommands.py
```diff
@@ -240,3 +240,54 @@ def execute_system_command(arg, **_):
...
+
+@export
+def redirectmatch(sql):
+    if sql[-1] in ("'", '"'): # find filename in quoted redirect
+        quote = sql[-1]
+        m = re.match(r'.*([^\\]'+quote+r'.*?'+quote+'$)', sql)
+        if m:
+            filename = m.group(1)
+        else:
+            filename = None
```
tsroten: Both here and on line 268, you set `filename` to `None`. You can set it to `None` before the conditionals so you don't have to set it twice.

meeuw: fixed :-)

mycli/packages/special/iocommands.py
```diff
@@ -240,3 +240,50 @@ def execute_system_command(arg, **_):
...
+@export
+def get_redirected_output():
+    return use_redirected_output
+
+@export
+def redirectmatch(sql):
+    filename = None
+    if sql[-1] in ("'", '"'): # find filename in quoted redirect
+        quote = sql[-1]
+        m = re.match(r'.*([^\\]'+quote+r'.*?'+quote+'$)', sql)
```
tsroten: Rather than putting the parenthesis around the quotes and the filename, put them just around the filename (inside the quotes), so that you don't have to remove the quotes later on. Then, `filename = m.group(1)` won't include the quotes.

Then you can get rid of line 262 (`quote = ''`) and lines 277-278 (`if quote: filename = filename[len(quote):-len(quote)]`).

Side note: if possible, put spaces around operators like `+`, e.g. `re.match(foo + bar)` instead of `re.match(foo+bar)`. It makes it easier to read.

tsroten: Hmm, actually I say that, but then I see that you use the length of the filename in line 267 to remove it from the query. That wouldn't work with my suggestion.

mycli/main.py
```diff
@@ -730,6 +733,15 @@ def run_query(self, query, table_format=None, new_line=True):
             for line in output:
                 click.echo(line, nl=new_line)
 
+    def output_to_file(self, text):
+        output = special.get_redirected_output()
```
tsroten: This is the second time `get_redirected_output()` gets called (see line 596). Can we pass an `append` argument to this function so we don't have to call it again?

tsroten:
I'll test it later on today, right now I've got to run. It would make the code more readable if you put newlines between conditional blocks.

So, instead of:
```python
if:
  foo
else:
  bar
if:
  oof
else:
  rab
```

Do this:
```python
if:
  foo
else:
  bar

if:
  oof
else:
  rab
```

mycli/main.py
tsroten: Gosh, I need to drink my coffee for the day! I didn't notice it was storing the response (`use_redirected_output`) instead of computing it again. Feel free to ignore that. Only suggestion here might be to condense the conditional `mode = 'a' if output['append'] else 'w'`.

meeuw:
@tsroten many thanks for the feedback

tsroten:
After testing this feature, here is some feedback:

1. This feature is very helpful. I've been wanting this for awhile.
2. The bash-like syntax of `>` has serious issues with SQL comparison operators. E.g. try to include a greater-than comparison in a where clause (`select * from foo where id > 100;`).
3. I wonder if a tee-like feature would be more handy than a strict redirect. In other words, I think it might be better to both redirect *and* display the output in the client (see mysql's `tee` command](https://dev.mysql.com/doc/refman/5.7/en/mysql-commands.html)). Otherwise, you can't ensure you are getting the correct data.
4. My guess is that it would be better to implement this using a `\` syntax like mycli's existing commands. For example, something like `\tee [file] [query]` or `\tee [query] [file]`. Or, we need to come up with a better syntax that won't clash with SQL.

I'd be interested in the rest of the team's thoughts, @dbcli/mycli-core.

j-bennet:
I agree with @tsroten on #4. I think it's better to not mix bash with sql, so +1 on making it into special command syntax.

`tee` looks like it would be a great addition to `mycli`. In mysql, `tee [filename]` sets file name to dump everything into (queries and output), and `untee` unsets it. We could add a flag to mycli's `\tee` to dump results only (`tee [filename] --results-only`).

darikg:
+1 with @j-bennet and @tsroten on using a `\special` command rather than redirect operators

meeuw:
Bah I forgot about the greater-than operator of SQL, thanks for spotting this.

I will rewrite this PR to use \>> and \> operators and create a new PR to use a \tee special command (which wouldn't require the hacks in sqlexecute). Then we can check what works better.

meeuw:
I've changed it to \>> and \>

meeuw:
Allright, I've changed the > and >> operators to \\> and \\>> so I guess it doesn't conflict with SQL operators anymore. In the end this is quite different from the tee implementation because tee adds the sql commands in the output and isn't automatically disabled. 

I'd like to use this to export query output to tsv or json, I suppose I only have to strip the n row(s) in set for maximum usability.

mycli/main.py
```diff
@@ -728,6 +731,12 @@ def run_query(self, query, table_format=None, new_line=True):
...
+        mode = 'a' if output['append'] else 'w'
+        with open(output["filename"], mode) as f:
```
tsroten: We should probably open with `encoding='utf8'`

tsroten:
@meeuw Okay, let's get this edited and merged in :) It's been sitting here awhile. Sorry about that!

Here are two thoughts:

1. Can we move the `output_to_file` to the special file so that the actual writing is out of `main.py` like you did in the tee command? Basically, let's make it consistent.

2. I think we can refactor the regular expression matching quite a bit. Here is an idea you might want to try:

```python
import re

sql = "SELECT first_name, last_name FROM customers \>> '../my file.txt'"
m = re.match(r'(?P<sql>.*)(?P<operator>\\>{1,2})\s+(?P<quote>[\'"]?)(?P<filename>.*)(?P=quote)', sql)

if m:
    append = m.group('operator').endswith('>>')
    print(m.group('sql'))
    print(m.group('operator'))
    print("Append: {}".format(append or 'False'))
    print(m.group('filename'))
else:
    print('No redirect match')
```

Outputs:
```
SELECT first_name, last_name FROM customers
Append: True
\>>
../my file.txt
```

meeuw:
@tsroten thanks for your feedback! I've moved output_to_file to iocommand.py and added utf-8 encoding.

The current regular expressions are indeed complicated but I couldn't find another way of verifying nested quotes.

For example, I can't find a single regular expression to pass (nested quotation):
```sql = "SELECT first_name, last_name, \" \>> \" FROM customers"```

Another problem is that the quotation cannot be checked (non matching quotation):
```sql = "SELECT first_name, last_name FROM customers \>> '../my file.txt\""```

As an alternative maybe we could use https://docs.python.org/3/library/shlex.html but I don't think that would make it less complicated and I'm not sure if SQL escaped strings are compatible with shell script.

Shell script is very flexible with opening/closing quotes:
```
>>> shlex.split('""""')
['']
```

And mycli/mysql isn't
```
mysql root@172.17.0.2:(none)> select """"
+---+
| " |
+---+
| " |
+---+
1 row in set
Time: 0.002s
```

**Comment by tsroten** [comment]
Thanks for the good examples. 👍 It looks like you've thought through some of the edge cases.

I'm interested in your thoughts on whether or not this is forcing SQL into places it shouldn't be (e.g. mixed with BASH-like syntax).

pgcli does it with a [`\o` special command](https://github.com/dbcli/pgcli/pull/483) that toggles output on/off. That would match the way our `\tee` command works.

I am not opposed to adding the redirect command, I'm just curious on your thoughts on how it fits within a database REPL.

**Comment by amjith** [comment]
> I'm interested in your thoughts on whether or not this is forcing SQL into places it shouldn't be (e.g. mixed with BASH-like syntax).

I've been having the exact same thought. It seems like a harmless addition that adds functionality but I've been having mixed feelings about adding a `>>` syntax to SQL. In some ways, I feel like it is intuitive and wish that SQL had this feature built-in. But the fact that it doesn't have this built-in makes me wonder if this feature will get enough adoption or if this will be an esoteric feature that only the devs of the tool know about.

**Comment by meeuw** [comment]
> I'm interested in your thoughts on whether or not this is forcing SQL into places it shouldn't be (e.g. mixed with BASH-like syntax).

Well, with a special command it is outside the SQL language. The advantage of the BASH-like syntax is that most people are familiair with redirects.

The most "MySQL" like way is to use ```SELECT ... INTO OUTFILE``` but it isn't possible to write to a local file.

I think you'll have to see \> as a special command and not like some shell script extension.

**Comment by meeuw** [comment]
Hmm I've noticed ```\.``` which also has it's roots in shell script.

**Comment by tsroten** [comment]
I think the issue is less that `\.`, `\>`, or `\>>` are inspired by the shell, and more that `\>` and `\>` are used within (after) actual SQL queries. In other words, the worry is more about shell-like syntax, rather than commands named like shell equivalents.

The only special command that uses queries right now is `\fs`, which begins with the special command, so it's easily parse-able. There is no ambiguity.

I agree with @amjith that is seems harmless. I'm worried about parsing edge cases.

It's interesting that both psql and sqlite3 implement this feature in separate commands:
- `.output` or `.once` for [sqlite3](https://sqlite.org/cli.html#writing_results_to_a_file)
- `\o` or `\out` for [psql](https://www.postgresql.org/docs/9.6/static/app-psql.html)

sqlite's `.once` is a neat idea because you don't have to disable it.

Plus, you can already do `>`/`>>` in a real shell with mycli:
```bash
mycli my_database --csv -e 'select id, first_name, last_name from users limit 5' > results.csv
```

meeuw:
The special command ```\e``` also uses queries (I've inspired this patch on the logic used for ```\e```)

But I could live with a new ```once``` special command, I'm not sure if it'll work. I will make a new PR/branch for that.

PR #354, opened 2017-02-26
fix doctest

meeuw:
this enables the doctest-modules tests and fixes a bug I've introduced in https://github.com/dbcli/mycli/pull/350 (in simple_separated_format)


amjith:
I don't think we were running doctests as part of our test suite. But this is a nice addition to make sure we don't break things in the future. 

Thank you!

:croissant:

Issue #288, opened 2016-07-18
Pager not triggered


georgexsh:
if table `foo` has 33 columns, type `desc foo` in a term which panel height is 40, pager will not be triggered, but the top most 8 lines not shown.
if panel resized to 39, paper will take over as expected.

retried reset `LESS` environ variable, not affected.

ps: runs inside a tmux session.

using mysli 1.7.1

johnksterling:
I reproduced this and was investigating resolving it.  I think it is caused because of the layers of tools being used to display the mycli console in a shell window.  It looked to me like it would be complicated to fix - and it isn't a very big pain point since it only happens when there is a small number of results.  I would vote for not fixing unless someone else sees an obvious way to work around it.

tsroten:
Hi @georgexsh! This is a combination of two things. First, mycli forces the less pager option `-F` which automatically causes the pager to quit if its contents can be displayed on one screen.

But, then the prompt toolkit we use to display the mycli prompt reserves extra blank lines at the bottom for the autocomplete functionality. You can see a related issue in a different app here: https://github.com/ipython/ipython/issues/9390

The easiest solution would probably be to allow users to override the less options that mycli uses. In other words, keep the current options (`-RXF`) as the default, but if the user specifies a config option, say `less_options`, then those will be used instead.

dbolser-ebi:
I have a related issue, I'd like the pager to be 'less -S' to "chop long lines". I'd also vote for a way to allow the pager to be customized.

amjith:
@dbolser-ebi You can configure the pager using `~/.my.cnf`. Here's what I have:

```
[client]
pager = 'less -XSRF'
```

This is `~/.my.cnf` NOT ~/.myclirc.

johnksterling:
I agree - that is what I do today.   It would be easy to also add a 'custom_pager_flags' configuration option that just over-rides the 'LESS' env variable with custom flags (they are hard-coded defaults today) - would there be any value in that?  I can spin up a PR for it if it sounds appealing for some reason.

amjith:
I'm reluctant to add more configuration options. I'm a big fan of the Fish shell's design documentation. http://fishshell.com/docs/current/design.html

[...]

PR #356, opened 2017-03-02
#288: only use the default less flags if they are not already set

johnksterling:
Small change to only over-ride the pager flags if they are not already set.


amjith:
Awesome! Thank you very much. 🍰

PR #358, opened 2017-03-02
Replace obsolete pycrypto with pycryptodome fork

mgorny:
The pycrypto library is abandoned since 2014 with a lot of bugs open.
It has been forked into pycryptodome which is mostly API-compatible with
pycrypto. FWICS, the bits used by mycli work fine with the new library.

Therefore, switch to using pycryptodome instead of obsolete pycrypto.
I've also enabled it for Windows since the latter library officially
supports this system and provides binary wheels for it, so
the dependency should no longer be a problem.

Since the actual Python code isn't changed, mycli can still formally
work with pycrypto or without the library at all. However, pip will now
install the new library.

### DIFF
--- a/mycli/config.py
+++ b/mycli/config.py
@@ -19,7 +19,7 @@ except ImportError:
 
 class CryptoError(Exception):
     """
-    Exception to signal about pycrypto not available.
+    Exception to signal about pycrypto(dome) not available.
     """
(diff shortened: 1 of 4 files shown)


amjith:
Awesome! Thank you for the contribution. 🎸

PR #359, opened 2017-03-02
tee special command

meeuw:
This is a implementation for tee proposed by @tsroten and @j-bennet for a \tee and \notee.

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -592,6 +592,7 @@ class MyCli(object):
                 self.output(str(e), err=True, fg='red')
             else:
                 try:
+                    special.write_tee(output)
                     if special.is_pager_enabled():
                         self.output_via_pager('\n'.join(output))
                     else:
@@ -649,6 +650,7 @@ class MyCli(object):
             while True:
                 one_iteration()
         except EOFError:
+            special.close_tee()
             if not self.less_chatty:
                 self.output('Goodbye!')
 
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -505,6 +505,7 @@ class MyCli(object):
             try:
                 logger.debug('sql: %r', document.text)
 
+                special.write_tee(self.get_prompt(self.prompt_format) + document.text)
                 if self.logfile:
                     self.logfile.write('\n# %s\n' % datetime.now())
                     self.logfile.write(document.text)
@@ -592,7 +593,7 @@ class MyCli(object):
                 self.output(str(e), err=True, fg='red')
             else:
                 try:
-                    special.write_tee(output)
+                    special.write_tee('\n'.join(output))
                     if special.is_pager_enabled():
                         self.output_via_pager('\n'.join(output))
                     else:
@@ -655,6 +656,7 @@ class MyCli(object):
                 self.output('Goodbye!')
 
     def output(self, text, **kwargs):
+        special.write_tee(text)
         if self.logfile:
             self.logfile.write(utf8tounicode(text))
             self.logfile.write('\n')
--- a/mycli/packages/special/iocommands.py
+++ b/mycli/packages/special/iocommands.py
@@ -258,7 +258,7 @@ def set_tee(arg, **_):
 
     try:
         tee_file = open(filename, mode)
-    except OSError as e:
+    except (IOError, OSError) as e:
         raise OSError("Cannot write to file '{}': {}".format(e.filename, e.strerror))
 
     return [(None, None, None, "")]
--- a/mycli/packages/special/iocommands.py
+++ b/mycli/packages/special/iocommands.py
@@ -251,7 +251,15 @@ def set_tee(arg, **_):
     else:
         mode = 'a'
         filename = arg
-    tee_file = open(filename, mode)
+
+    if not filename:
+        raise TypeError('You must provide a filename.')
+
+    try:
+        tee_file = open(filename, mode)
+    except OSError as e:
+        raise OSError("Cannot write to file '{}': {}".format(e.filename, e.strerror))
+
     return [(None, None, None, "")]
 
 @export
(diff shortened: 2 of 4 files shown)


amjith:
I just pulled this down and tried it out. The happy path works fine which is starting with `\tee  filename`. But there are a couple of corner case issues. 

1. If the file already exists it overwrites it. I would have expected it to append to the existing file. 
2. It doesn't actually write out to the file until I quit mycli. You can check this by starting a `\tee abc` and start writing commands, but in a different terminal run a `tail abc` and you'll see nothing. This can be problematic, because if mycli were to crash then the file won't be written to the disk at all. So we need to flush the buffer when we write to the file. 
3. I think the filename is a required parameter for `\tee`, so the doc string for it should say `\tee filename` and NOT `\tee [filename]`. The square brackets imply that it is optional.

Cheers!

meeuw:
@amjith thanks for testing, I've added a -a append option and added flush.

My use case for non-append mode is to export data like this:
`\T tsv`
`\P true`
`tee one.tsv ; select 1`
`tee two.tsv ; select 2`
(which can be repeated from readline history to update the results)

amjith:
The output that goes into the file has one missing newline. Here's what it looks like: 

```
|   2 |
|-----|
|   2 |
+-----+
1 row in set+-----+
|   3 |
|-----|
|   3 |
+-----+
1 row in set+-----+
|   4 |
|-----|
|   4 |
+-----+
1 row in set+-----+
|   5 |
|-----|
|   5 |
+-----+
1 row in set
```

amjith:
This is close to completion. If you can address the two latest comments we'll be ready to merge.

meeuw:
@amjith and everyone else, thanks for the feedback I'll fix it. There's no hurry for this fix, I have the time :-)

meeuw:
I've changed the "\n".join() to an for-loop writing the output items and writing a newline and updated the tests.

tsroten:
I think I like the `tee` command better than #353. This works great. Thanks for the tests, too. I'm excited about this coming to mycli.

Here are a few thoughts on possible changes.

I believe append should definitely be the default and not require a special argument. If anything, maybe `-o` for overwrite (or something similar) could be another approach that might provide a more sensible default. I think append as a default is less of a surprise than overwriting as the default.

One thing I'm worried about is that the tee file is not explicitly closed if the user quits mycli without executing `notee`. I can't imagine this being a real problem, but it's best practice to explicitly close the file.

So, I'm wondering if maybe it's better for `write_tee()` to use a context manager to open and write the file (so it gets closed immediately). For example, in `write_tee()`:
```python
global tee_file, tee_file_mode
if tee_file:
    with open(tee_file, tee_file_mode)
        for buf in output:
            tee_file.write(buf)
            tee_file.write(u"\n")
```

I think the term `logfile` will be confusing to users because they might think it has something to do with debug-type messages. What about `output file` or something similar?

meeuw:
Hi @tsroten thanks for the feedback.

Could you give me an example when you'd expect/require append mode? I'm following the (GNU) tee defaults here; hence the -a parameter. My queries usually can be re-run to get the results again.

I've added tee_close which is called when mycli quits and changed logfile to output file.

I don't like closing and opening the file for every resultset, flushing it everytime should be enough right?

tsroten:
Hi @meeuw -- I'm happy with the flushing on every write (instead of closing) since I don't think it will create any real problems. I just wanted to at least bring up so we would think through that piece of it.

As for the append mode as default -- ultimately it comes down to whether or not we want to imitate mysql's `tee`/`notee` commands, or the GNU `tee` shell command. As this pull request currently stands, here are the notable differences between this implementation and mysql's:
- append isn't the default
- the prompt and query are not logged in the file
- the tee filename isn't remembered so a subsequent `tee` call doesn't require passing a filename

To be clear, I'm not saying we have to hit all three of those points in this pull request. I just wanted to point out the differences.

In the face of two exiting paradigms (official mysql client and GNU command-line tool), I personally favor imitating mysql because many users use mycli as a replacement for the mysql client. Plus, mycli has a way of using command-line tools already (e.g. `\P tee outfile.txt` or `\P tee -a outfile.txt`).

Plus, it errs on the side of caution. In other words, we'd rather make the user explicitly ask to delete the data they already have, not ask to keep it. This might be especially helpful if the user is using the `tee` command as multi-session log file. Or, if they accidentally type in the filename they used last session when they really wanted a new one.

That being said, I'm totally up for others to weigh in differently (e.g. @amjith, @j-bennet).

amjith:
> we'd rather make the user explicitly ask to delete the data they already have, not ask to keep it.

I think @tsroten has hit the nail squarely on the head with that sentence. 😄 

In mycli we have deviated from mysql whenever necessary to make it safe for the user. In this case, simply following mysql's lead and defaulting to append mode seems like the superior choice.

j-bennet:
I would also prefer `mysql`'s take on it, because a user coming from `mysql` shell to `mycli` may be put off / frustrated if the behavior is different from what he's used to expect. If we have a good justification for the change (new way is clearly better), then it's a different story, but in this case there's no clear winner, so `mysql` should be the standard to fall back on.

meeuw:
Allright, I've changed it to -o to override (I don't really care). If we want mysql compatibility we should also support \T (tee) and \t (untee) which conflict with mycli's \t (\timing) and \T (tableformat). And I suppose the sql commands should also be logged (especially with appended results). I will try to fix this.

tsroten:
@meeuw Thanks for seeing this through! I'll test it out tonight. Then, I'll look at #353.

These are some great additions to mycli. Thanks again!

tsroten:
@meeuw Ok, this looks good. I made a couple of small changes to the wording/error handling. I'm going to go ahead and merge this. Thanks again!

meeuw:
@tsroten thanks for merging!

PR #360, opened 2017-03-03
Add Python 3.6 to tests and setup file.

tsroten:
This pull request adds Python 3.6 to tox, Travis CI, and the setup file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ python:
   - "3.3"
   - "3.4"
   - "3.5"
+  - "3.6"
 
 env:
   - PYMYSQL_VERSION=0.6.7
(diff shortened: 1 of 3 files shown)


amjith:
Nice!!

Do you think the time has come to kick out Python 2.6?

tsroten:
@amjith I was just thinking about that as well. I think one of the reasons for supporting it was RHEL 6. [Here is an article](https://tecadmin.net/install-python-2-7-on-centos-rhel/) from Jan 2017 about Python 2.7 on RHEL 6.

It looks like 2.6 is pre-installed, but it doesn't look that difficult to install Python 2.7 alongside it.

I think Python 2.6 has been unsupported for nearly 4 years now. Personally, I vote we drop support. October 2013 is when [official support ended](https://www.python.org/dev/peps/pep-0361/#release-lifespan). That's a long time ago!

PR #363, opened 2017-03-07
Update Detailed Install Instructions for Fedora

meeuw:


amjith:
Nice! I didn't know about `dnf`. It's pretty cool that mycli has a package in dnf. 😄

meeuw:
dnf is the new yum

PR #157, opened 2015-09-24
Backport context managers in pymysql from 0.6.6 to 0.6.2

amjith:

Issue #155, opened 2015-09-21
mycli crashes with pymysql 0.6.2


lhw:
The bug which caused the crash with `pymysql 0.6.2` (#83) still exists. `pymysql 0.6.6` only handles it gracefully. ConfigObj is unable to parse the MySQL config files. Which later on caused a crash with pymysql not handling an empty field/fields correctly. Version here is `mycli 1.4.0`
```
2015-09-21 12:58:51,768 (17883/MainThread) mycli DEBUG - Initializing mycli logging.
2015-09-21 12:58:51,769 (17883/MainThread) mycli DEBUG - Log file '~/.mycli.log'.
2015-09-21 12:58:51,770 (17883/MainThread) mycli.main ERROR - Error parsing u'/etc/mysql/my.cnf'.
2015-09-21 12:58:51,770 (17883/MainThread) mycli.main ERROR - Recovering partially parsed config values.
2015-09-21 12:58:51,772 (17883/MainThread) mycli.main ERROR - Error parsing u'/etc/mysql/my.cnf'.
2015-09-21 12:58:51,772 (17883/MainThread) mycli.main ERROR - Recovering partially parsed config values.
2015-09-21 12:58:51,773 (17883/MainThread) mycli.sqlexecute DEBUG - Connection DB Params: 
        database: None  user: u'root'   host: u'localhost'      port: 3306      socket: '/var/run/mysqld/mysqld.sock'   charset: u'utf8'
2015-09-21 12:58:54,534 (17883/MainThread) mycli.sqlexecute DEBUG - Connection DB Params: 
        database: None  user: u'root'   host: u'localhost'      port: 3306      socket: '/var/run/mysqld/mysqld.sock'   charset: u'utf8'
2015-09-21 12:58:54,537 (17883/MainThread) mycli.main DEBUG - Launch Params: 
        database: None  user: u'root'   host: None      port: None
2015-09-21 12:58:54,537 (17883/MainThread) mycli.main DEBUG - Original value for LESS env var: '-R'
2015-09-21 12:58:54,540 (17883/MainThread) mycli.main ERROR - Error parsing u'/etc/mysql/my.cnf'.
2015-09-21 12:58:54,540 (17883/MainThread) mycli.main ERROR - Recovering partially parsed config values.

```
My current config is basically the standard MariaDB config, I put them up in a gist here: https://gist.github.com/lhw/c4893fe72f9141129357

[...]

PR #365, opened 2017-03-11
Fix requirements

meeuw:
## Description
I've did some quite naïve tests (only run pytest & start mycli) and changed the following minimum requirements for mycli. 

Pygments lowered to >= 1.6
PyMySQL raised to >= 0.6.7
configobj lowered to 5.0.5
pycryptodome set minimum to 3

## Checklist


tsroten:
@meeuw Thanks for being proactive about the version numbers!

For PyMySQL, we intentionally support 0.6.2 (#155, #157) because at one time, that was the version that Debian used. Before we drop support for it, it would be nice to double-check that this isn't still required for Debian. If it's not, we can also remove lines 25-51 [here](https://github.com/dbcli/mycli/pull/157/files#diff-ac5c664011ca63abd99ea29aeeb4a641R25).

meeuw:
Hmm when using PyMySQL 0.6.2 I get errors:

traceback: 'Traceback (most recent call last):
  File "mycli/mycli/main.py", line 395, in connect
      socket, charset, local_infile, ssl)
        File "mycli/mycli/sqlexecute.py", line 45, in __init__
            self.connect()
              File "mycli/mycli/sqlexecute.py", line 80, in connect
                  conv=conv, ssl=ssl)
                    File "mycli/mycli/packages/connection.py", line 51, in connect
                        return Connection(*args, **kwargs)
                        TypeError: __init__() got an unexpected keyword argument \'local_infile\'
                        '
__init__() got an unexpected keyword argument 'local_infile'

tsroten:
Hmm, I guess when I added `local_infile` support I didn't test it against 0.6.2. [It looks like](http://metadata.ftp-master.debian.org/changelogs/main/p/python-pymysql/python-pymysql_0.7.6-1~bpo8+1_changelog) sometime last year they bumped the version in Debian. So, let's go ahead and drop it. In fact, if we bump it to something greater than 0.6.5 (it was set at 0.6.6 back in 2015), we can remove all the changes that #157 introduced, including the feature backport.

For more context on this, see #132. From that PR, you can see that bumping it to >= 0.6.7 actually fixes a bug. So, I'm going to suggest we use `>= 0.6.7`. Then we can:
- remove code that #157 introduced
- remove this pinned travis version that's no longer needed (it can just use the newest version from pip): https://github.com/dbcli/mycli/blob/master/.travis.yml#L10

It looks like configobj 5.0.0 - 5.0.5 had some unicode bugs. Since mycli needs to support unicode, even in the config file (e.g. favorite queries), I think I'd prefer that we don't change this version unless we have a good reason to.

Good call on pinning a major version for pycryptodome.

As far as Pygments goes, dropping a major version (to 1.6 from 2) where 1.6 was released over 4 years ago, I'm not sure if we gain anything there. What do you think?

amjith:
I'm in favor in pinning pycryptodome and raising the min version for pymysql. I don't fully understand the reason for lowering pygments or configobj. 

Maybe if you can give us some context, it'll help us make a decision.

meeuw:
There is no specific reason for lowering the required version of pygments, CentOS has an even lower version of pygments so it wouldn't help for that. But I think it's a good thing to have the lowest minimal requirements.

The most important change is PyMySQL because that one is broken right now, but I suppose that requires more changes/cleanup.

tsroten:
This is a good discussion on mycli's dependencies.

I'm thinking we definitely want to keep configobj where it's at so as to avoid the unicode bugs present in the earlier versions.

For pygments, I'm wary of supporting a previous major version without a strong reason to go in that direction. So, maybe we can hold off on that unless it begins to cause issues with our users?

It's going to be great to get rid of some of the extra code for PyMySQL. @meeuw, would you be able to remove that code in this PR?

meeuw:
roger that!

amjith:
I agree with the things that Thomas said. Here are my thoughts about dependencies. If we can get away with requiring a newer version of a library, I'd prefer that. Only because newer versions are (hopefully) less buggy. 

So unless we have a specific reason to support an older version (because Centos or Debian doesn't have a new version) then I'm all for it. Otherwise, newer the better.

meeuw:
I've just noticed pycryptodome is an optional dependency, maybe it should be removed from setup.py?

tsroten:
@meeuw Thanks for pointing out this discrepancy.

While there is code that treats it as optional, it's not technically optional anymore -- it is required when mycli is installed. Previously, it was not required with mycli on Windows (the user could install it manually later), but on all other systems it was a requirement. See #169 for more information on why that decision was made at the time.

Even though it was made a requirement on all systems, including Windows, the code to handle it as a missing dependency is still there. If we're going to remove anything, it should probably be the code that made it optional (#181).

Our other option is to [add it as an "extra"](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies), but that has its own complications for our user base.

Since many of our users are installing mycli from a package manager other than `pip`, it's probably better to keep `pycryptodome` in the `install_requirements`. If we don't (like if we put it as an "extra"), then people who want that feature will have to be told to install it with `pip`, which might be confusing to them.

`pycrtyptodome` is better packaged than `pycrypto`, so my vote is that we shouldn't make it optional unless we see a demand for that.

What do you think?

amjith:
I agree that listing it as an extra is cumbersome for the users.

I just checked their site and it looks like they supply wheel packages for windows. So the original reason to make it optional doesn't exist anymore. 

So I'm in favor. 👍

amjith:
Is this ready for merge? Are we waiting for more changes?

tsroten:
@amjith I think there are still a few issues:
- pymysql needs to be bumped to `>= 0.6.7` now that the backported fixes are removed.
- Pygments should probably be kept at `>=  2.0.0`
- configobj should probably be kept at `>= 5.0.6`

meeuw:
I'd like to add some unit tests to prove we need the higher versions of pygments and configobj.

tsroten:
## configobj
For configobj, you can't install mycli with configobj<=5.0.3 because prior to 5.0.4 configobj had strict requirements on what version of `six` to use, and this conflicts with python prompt toolkit's requirements.

For configobj 5.0.4, you can't save favorite queries that have unicode in them. E.g. `\fs foobar select * from foo where bar = '你好';`

It looks like we could get away with configobj>=5.0.5, so I'm on board with adjusting it to 5.0.5 and newer.

-------

## pygments

I guess we don't have a specific reason for pinning this to >=2.0.0. Version 1.6 is over 4 years old, supports Python 2.4 and 2.5, and doesn't make use of some Python 3-specific features. While we don't have an explicit reason for pinning this to >=2.0.0, it doesn't sound like we have an explicit reason to bump it down to >=1.6.0 (other than the idea of favoring wide dependency requirements). If you think this is important, go ahead and bump it to >=1.6.0, I'm fine with that.


-------

## pymysql

Our existing tests fail when pymysql==0.6.4.

For pymysql<=0.6.6, you can't view certain binary data. Here is an example: https://github.com/PyMySQL/PyMySQL/issues/363

That error is caught here: https://github.com/dbcli/mycli/blob/2d1dd499d54e9d2259c793dd10b630a9a8d80d74/mycli/main.py#L544

If we require pymysql>=0.6.7, then we can remove that exception catch code. Especially, now that it seems Debian's version is > 0.6.7.

Since our original reason for supporting <= 0.6.5 is no longer valid, we are left with the decision to start our requirement at 0.6.6 or 0.6.7. Since that's not a huge step up, and 0.6.7 removes a bug, it makes sense to go with >=0.6.7 and then also remove the exception handling required for 0.6.6 and below.

meeuw:
@tsroten I've updated the dependencies according to your suggestions and removed the pymysql<=0.6.6 workarounds/warnings. I only need to remove the optional pycrypto code.

tsroten:
Great! Once that is done, this looks like it's ready. Thanks for the work on this and for helping us think through the various dependency versions!

meeuw:
I've tried to save an query with unicode in it:
`\fs check select '✔'`

But I get, even with `configobj==5.0.6`:
`'ascii' codec can't encode character '\u2714' in position 3274: ordinal not in range(128)`

Could you please check which configobj version you're using?

tsroten:
@meeuw Haha, silly me. Looks like it was a bug in mycli as well. https://github.com/dbcli/mycli/pull/379

That fix works with 5.0.5 and 5.0.6. It makes the bug with 5.0.4 much more evident.

PR #366, opened 2017-03-11
pytest environment variables

meeuw:
use environment variables for pytest for alternative host/port/etc.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -3,9 +3,11 @@ from mycli.main import format_output, special
 from mycli.packages import connection
 from os import getenv
 
-# TODO: should this be somehow be divined from environment?
-USER, HOST, PORT, CHARSET = 'root', 'localhost', 3306, 'utf8'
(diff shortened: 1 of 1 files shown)


tsroten:
Good thinking on this. Environment variables are definitely helpful here. I don't see any downsides to this, but let's check with @amjith to see if there is any reason we shouldn't rename the existing environment variable from `PASSWORD` to `PYTEST_PASSWORD`. As far as I know, this will only affect people who are regularly running mycli tests, which is a small number of people.

An alternative idea, would be to look for a `mycli_test` section of the encrypted login path file and use the user/host/pass information from there. But, I think the environment variables work just fine.

amjith:
No objections from me. 

I'm not opposed to the idea of using login path file either, but I don't use it all that often. So either direction is fine by me.

tsroten:
Great, even if we want to expand to login path support later, this is an excellent starting point. Thanks @meeuw!!

PR #374, opened 2017-03-19
Use twine for uploading to PyPI.

tsroten:
## Description
This pull request switches mycli's release script to use [twine](https://github.com/pypa/twine) for uploading to PyPI. Setuptools is officially [not recommended](https://packaging.python.org/distributing/#upload-your-distributions) by the Python Packaging Authority.

And, it removes an extra call to `sdist`.

## Checklist

### DIFF
--- a/release.py
+++ b/release.py
@@ -72,12 +72,8 @@ def register_with_pypi():
     run_step('python', 'setup.py', 'register')
 
 
-def create_source_tarball():
-    run_step('python', 'setup.py', 'sdist')
-
-
-def create_python_wheel():
-    run_step('python', 'setup.py', 'bdist_wheel')
+def create_distribution_files():
+    run_step('python', 'setup.py', 'sdist', 'bdist_wheel')
 
 
 def upload_distribution_files():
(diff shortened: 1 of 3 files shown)


release.py
```diff
@@ -71,10 +71,11 @@ def create_source_tarball():
     run_step('python', 'setup.py', 'sdist')
 
 def create_python_wheel():
-    run_step('python', 'setup.py', 'sdist', 'bdist_wheel')
+    run_step('python', 'setup.py', 'bdist_wheel')
```
amjith: Does this mean we're only generating wheel packages? No more source packages?

tsroten: Good question. Nope -- there is still the `create_source_tarball()` function that calls `python setup.py sdist`.

In the call to twine, it uses `twine upload dist/*`, meaning, upload any distribution file, which will include the `*.tar.gz` and `*.whl` files.

I've gone ahead and merged the two create_* functions into one `create_distribution_files()` for clarity.

amjith:
👍

Issue #377, opened 2017-03-20
special command \e doesn't work with prompt_toolkit>=1.0.10


meeuw:
Special command \e doesn't seem to work with new version of prompt_toolkit

Steps to reproduce:
1. `mysql root@172.17.0.2:(none)> select 1\e` 
2. change 1 to 2 in editor, save/quit

Expected (works with prompt_toolkit==1.0.9):
`mysql root@172.17.0.2:(none)> select 2` 

Actually (with prompt_toolkit>=1.0.10)
`mysql root@172.17.0.2:(none)> `

meeuw:
we might want to change setup.py d41c29f1dd275df811af57ba376b5b742afac149

tsroten:
A workaround in mycli's Vi-mode is to hit <Esc>, then hit `v` to bring up the editor. Editing and saving the contents will correctly edit (and immediately execute) the query.

tsroten:
I think it has something to do with [this change](https://github.com/jonathanslenders/python-prompt-toolkit/commit/903018201e0b0f5ad4090c08879fa83f665413b3).

I'll take a look at it later and see if there is an easy way to fix it on mycli's side without changing `setup.py`, otherwise, we can bump it down like you suggested.

tsroten:
Homebrew was just updated with mycli 1.9.0 and prompt toolkit 1.0.13 :(

amjith:
I know that @j-bennet has a fix for it in pgcli. Not yet merged. Maybe she
can shed some light.

On Mar 20, 2017 7:11 AM, "Thomas Roten" <notifications@github.com> wrote:

Homebrew was just updated with mycli 1.9.0 and prompt toolkit 1.0.13 :(

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/dbcli/mycli/issues/377#issuecomment-287770243>, or mute
the thread

tsroten:
Thanks @amjith! Here is the issue for reference: https://github.com/dbcli/pgcli/issues/668
And, the pull request: https://github.com/dbcli/pgcli/pull/670

j-bennet:
Fix is now in master.

tsroten:
mycli version 1.10.0 is now out which fixes this.

PR #378, opened 2017-03-20
Fixes editor bug with prompt_toolkit 1.0.13.

j-bennet:
## Description

Fix for https://github.com/dbcli/mycli/issues/377.

Ported from https://github.com/dbcli/pgcli/pull/670.

## Checklist


meeuw:
Do we need backward compatibility for an older version of prompt_toolkit? Because this fix doesn't work with prompt_toolkit==1.0.9 (which is the current version for Fedora).

j-bennet:
Hmm. Or, the other option would be to bump minimal requirement of prompt_toolkit. Any specific reason we're keeping it at 1.0.9?

amjith:
Whatever solution we pick needs to work on both pgcli and mycli. Because we have users who use both and they are not able to install pgcli and mycli simultaneously because we've required conflicting prompt-toolkit versions in the past.

j-bennet:
Agree. I think we still have the mismatch in master, pgcli specifies 1.0.0 as the lowest, and mycli has 1.0.9.

meeuw:
> Any specific reason we're keeping it at 1.0.9?
Fedora 25 provides a packaged version of mycli which uses prompt_toolkit 1.0.9 but I've investigated and all other packages depending on prompt_toolkit require prompt_toolkit<1.1.0 or no specific version so that will be their own fault if it breaks. I'm all in for a new version of prompt_toolkit.

tsroten:
@j-bennet This looks good. Since it sounds like this fix does not work with <1.0.10 (right?), then we need to bump the `setup.py` prompt toolkit, which is currently at 1.0.0 through 1.1.0.

tsroten:
Oops, sorry about closing that!

j-bennet:
I've been swamped the last few days, but this is exactly what I'm going to do: bump prompt-toolkit to 1.0.10 in this PR and merge.

PR #379, opened 2017-03-22
Open config file using UTF8 encoding.

tsroten:
## Description
This opens the `myclirc` file using UTF-8 encoding so that unicode characters can be written properly.

## Checklist


meeuw:
That seems to be working, I've added an unit test.

tsroten:
@meeuw Thanks for the unit test!

Issue #375, opened 2017-03-19
Fixed precision decimals are not outputted correctly.


mikeerickson:
When executing the following  (compare results in AVG column)

```
select G, AB, H, REPLACE(ROUND((H/AB),3),'0.','.') as AVG from Batting limit 25;
```

It produces the following in mycli

![image](https://cloud.githubusercontent.com/assets/183153/24078178/c6d232ee-0c23-11e7-86bd-93928d958e86.png)

When using same command in mysql cli

![image](https://cloud.githubusercontent.com/assets/183153/24078179/f5350292-0c23-11e7-8582-d7547496abdd.png)

meeuw:
Hi @mikeerickson I see what you're trying to achieve here and my best guess is that this is a bug in our tabulate library. There is some code in it to detect python Decimal object and then handle it as a string (and prevent the adding/removal of leading/trailing zero's). I will have a look at it soon.

You can see it doesn't get triggered with a Dutch locale (which tabulate doesn't understand)
```
mysql root@172.17.0.2:(none)> select format(.1, 3, 'en_US')
+--------------------------+
|   format(.1, 3, 'en_US') |
|--------------------------|
|                      0.1 |
+--------------------------+
1 row in set
Time: 0.002s
mysql root@172.17.0.2:(none)> select format(.1, 3, 'nl_NL')
+--------------------------+
| format(.1, 3, 'nl_NL')   |
|--------------------------|
| 0,100                    |
+--------------------------+
1 row in set
Time: 0.002s
```

tsroten:
In our tabulate library, it converts columns where all numbers have decimals into floats (which removes any exact precision).

The only solutions I can think of are:
1. Modify tabulate to look at the precision, and then format it accordingly on output.
2. Remove the float conversion code from tabulate so that it treats floats as strings (meaning they will be outputted as mysql returns them).

meeuw:
@tsroten I was thinking about getting pymycli to output a Python Decimal object, that seems to be handled like a string in tabulate.
https://github.com/dbcli/mycli/blob/master/mycli/packages/tabulate.py#L338

But I need to dig into this, not sure if that works.

tsroten:
@meeuw Ah, neat idea.

It looks like PyMySQL automatically converts decimal fields into Decimal objects, but in this example it is not a decimal field, it is the result of a mysql function. We'd have to have some way of autodetecting the type (which would include checking *every* value in a given column). That might be resource-intensive.

[...]

PR #380, opened 2017-03-23
Use pymysql default conversions (instead of only our conversions)

meeuw:
## Description
This will convert a MySQL `DECIMAL` to the Python's `decimal.Decimal` type.
Tabulate handles Decimal as a string (and not a number) and will not remove the padding (fixes issue #375)

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -10,6 +10,7 @@ Bug Fixes:
 * Fix requirements and remove old compatibility code (Thanks: [Dick Marinus])
 * Fix bug where mycli would not start due to the thanks/credit intro text.
   (Thanks: [Thomas Roten]).
+* Use pymysql default conversions (issue #375). (Thanks: [Dick Marinus]).
 
 Internal Changes:
 -----------------
--- a/mycli/packages/tabulate.py
+++ b/mycli/packages/tabulate.py
@@ -320,11 +320,13 @@ def _type(string, has_invisible=True):
     True
     >>> _type("foo") is type("")
     True
-    >>> _type("1") is type(1)
+    >>> _type("1") is type("")
     True
-    >>> _type('\x1b[31m42\x1b[0m') is type(42)
+    >>> _type(1) is type(1)
     True
-    >>> _type('\x1b[31m42\x1b[0m') is type(42)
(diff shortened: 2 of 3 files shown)


tsroten:
@meeuw This is helpful. I think it makes sense to use PyMySQL's converters for the most part.

I'm not seeing how it fixes #375, though. In that issue, we need the decimal numbers to be treated as strings (so it can be formatted like baseball batting averages, e.g. `.300`).

What do you think?

meeuw:
@tsroten it fixes it for the first part; it's now formatted as `0.300` instead of `0.3`.

Maybe we also should remove `_isconvertible(int/float, string)` from tabulate, then you can use replace() to do number formatting.

meeuw:
I've added a fix to use python typed strings only as strings (and not float/int). This feels a bit like raping tabulate because this seems to be an important feature for tabulate. Maybe we should make this configurable.

Does @dbcli/mycli-core know why this is an embedded library and why we aren't using https://pypi.python.org/pypi/tabulate?

tsroten:
I do agree it's not the way tabulate was intended, but it suits our needs well. In general we want to display what MySQL returns, not a formatted version of what is fetched.

This is especially true in this PR's related issue, where explicit formatting is done in the query. We want this to be easy for the user (i.e. not have to turn on/off formatting settings in mycli).

So, I think your solution works well. I'm going to test it out a bit more before I merge it in.

We've made [various changes over time](https://github.com/dbcli/mycli/commits/master/mycli/packages/tabulate.py) to tabulate to make it work better for mycli. Tabulate tries to format data a lot (which makes sense in a lot of contexts outside of mycli). I wonder if we should consider (not in this PR), moving to a pretty print library that does not format data. Because, all we need it to do is to display data in an ASCII table. We don't want it to format data.

tsroten:
@meeuw These changes have made their way over to the new output formatter branch, right?

We don't need to close this just yet, unless you want to, but I wanted to make sure we weren't forgetting anything important from this PR in the new one.

meeuw:
@tsroten only the tabulate tests aren't in the new output formatter branch (because they don't apply anymore). Lets close this PR but keep the branch,

PR #386, opened 2017-03-25
Replace tabulate with terminaltables.

amjith:
@meeuw @tsroten @j-bennet 

This is my first attempt at replacing the `tabulate` library with an alternative. It is in no way near shippable quality. I'm just putting it out there as a prototype. I'll keep polishing it and you're all welcome to take a stab at it as well. Let's collaborate on this branch. 

## Checklist


tsroten:
@amjith Thanks for getting this started! The basics are already working in this branch, so this is a great step forward.

amjith:
I'll fix the failing tests. Didn't realize there were format_output() tests in test_main.py.

amjith:
I have fixed some of the tests. But I'm too tired to fix the rest right now. If someone else is interested in fixing things, just leave a comment here, so we're not duplicating effort. 

Things to note: 

* There is not `tsv` format. This means the batch mode automatically defaults to table output. 
* We're relying on terminaltable to tell us if the table is too wide to fit in the terminal. So we are no longer in the business of measuring the terminal width and determining the max-width. This means we can't write a test to verify that auto-vertical mode is working or not. We will just rely on the library to do this reliably.

meeuw:
I think I've fixed the failing pytest. Do we keep support for table_format or will that be removed?

I don't know what the purpose is of the batch mode of mycli, the official mysql client is much better at it (faster/less memory usage).

amjith:
Great questions:

>Do we keep support for table_format or will that be removed?

That will be down-sized to the formats supported by `terminaltables`. We can work on adding more formats later.

>I don't know what the purpose is of the batch mode of mycli, the official mysql client is much better at it (faster/less memory usage).

Batch mode is the option that allows you to run a quick query right from the command line using the `-e` option so you don't have to launch mycli to run just one command. You're right that mysql client will be much faster but some people (like me) have aliased `mysql` with `mycli`. So I never ever have to interact with the official mysql client. That's why we support that option. I know @tsroten added that feature. So he might be able to weight in on how he uses it. 

When I said that batch mode won't support tsv, I wasn't complaining I was merely pointing out. It could be an improvement. We'll add tsv later. 

When we do change to `terminaltables` library we should release that as mycli v2.0. Because we'll be breaking the myclirc format and the command line options.

amjith:
@meeuw Thanks for being proactive in fixing the tests and making progress on this PR. 🍄

tsroten:
@meeuw @amjith The `--execute`/`-e` option was added by @mdsrosa after I added stdin batch mode.

I don't actually use the batch mode much myself. I added it based on https://github.com/dbcli/mycli/issues/179 because I thought mycli should be unix-y and be able to accept standard input.

Personally, I think standard input support should be a given for a command-line app, so I'm hesitant to suggest getting rid of it.

That being said, it is related to an overall vision discussion, i.e. is mycli a full-fledged mysql replacement, or is it simply a better *interactive* client (without support for batch mode commands/features).

What do you think?

tsroten:
As far as not having tsv support for this PR -- I don't think that's a big deal. In fact, I'd say, don't add tsv support in this PR -- keep this PR limited to strictly adding terminaltables and it's capabilities.

This might be a good excuse for another PR to standardize a way to export in a variety of formats in both interactive and batch mode (including tsv), i.e. [Tablib](http://docs.python-tablib.org/en/latest/). That was part of the discussion on https://github.com/dbcli/mycli/issues/91.

amjith:
> keep this PR limited to strictly adding terminaltables and it's capabilities.

Yes! I agree.

meeuw:
Maybe it's a good idea to make a generic interface to different table format engines, and not only limit this to one project. This way we can keep tabulate and add tablib/ terminaltables.

I've had another problem with terminaltables, it doesn't play nice with less (pager) when using the SingleTable formatter.

amjith:
> Maybe it's a good idea to make a generic interface to different table format engines, and not only limit this to one project. This way we can keep tabulate and add tablib/ terminaltables.

That's a great idea. Let's abstract the format_tables into a separate file and use that as a starting point for a unified interface. It'll also help us benchmark the different libraries and pick the one that has the right balance of features and performance.

>I've had another problem with terminaltables, it doesn't play nice with less (pager) when using the SingleTable formatter.

That's unfortunate. I haven't tried the other formats yet.

amjith:
@tsroten Do you want to take the lead on designing the interface?

tsroten:
@amjith @meeuw Here are some thoughts to get us started on an abstract interface. Let's collaborate on this to make sure it's intuitive and abstract.

## Goal

Create a single, stand-alone interface that mycli uses to format results from mysql prior to output.

## Requirements

- The interface should know nothing of mycli or mysql. It's dealing with generic data that is passed as an iterable.
- The interface needs to have a way to list its supported formats (like tabulate's table_formats).
- The interface should allow mycli to specify (in one call) a list of headers, the data itself, and the format to use.

## The Interface

The interface needs at least two endpoints for mycli to hit:
1. Get a list of supported formats.
2. Format the data

The interface needs to have a way for us to register a format, with the following information:
- A name (e.g. `simple` or `fancy_grid`)
- A callable that accepts a standard set of arguments, e.g. `simple(headers, data)`

For example:
```python
register_output_format('simple', lambda headers, data: tabulate(data, headers, 'simple'))
```

Any package could be used to provide the logic for an output format, it simply needs to be registered with a callable that accepts `headers` and `data`.

The formatted data should be a list (or tuple --basically any iterable) of rows for output. The formatted data should not include line endings between rows.

mycli would then be able to call something like `format_results(headers, data, format='fancy_grid')` (assuming it knows the user wants format `fancy_grid`).

## Additional Thoughts

- mycli's expanded output should be moved from `main.py` into this interface, e.g. `format_results(headers, data, format='expanded')`.

What do you think? Please provide suggestions/critiques. Let's make this a quality formatting interface that will be extensible for the future.

meeuw:
@tsroten nice work!

> The interface should know nothing of mycli or mysql. It's dealing with generic data that is passed as an iterable.
Hmm but we need to tell the formatter how it should handle the data and the Python provided types won't be sufficient. For example there is no nice mapping for "binary data" for Python 2 (there is no bytes type) and for example Spatial Data Types aren't provided by Python at all.

It would be nice to give parameters to the formatter, for example to set parameters (quote style, separator value etc) for csv output. The "expanded" format (this is `\G` right?) could also be supplied this way.

Maybe we also should give some meta information from the output from the formatter like how the output was formatted? Like does it contain terminal escape characters which aren't compatible with less (pager).
This also could be done by requiring the formatter to always give output that will be accepted by less.

amjith:
> The interface should know nothing of mycli or mysql. It's dealing with generic data that is passed as an iterable.

I agree with this design requirement. I know that @meeuw disagrees but I think if we can't express the data using Python data types then we have failed.

@tsroten It looks like the `register_output_format` is deciding on the configuration and then the `format_output` is using that configuration to do the right thing. I think they should both be part of a class. So the `register_output_format` will be the constructor that takes in various args to create an object and the `format_output` method will be called on that object. 

We need one more argument for the `register_output_format` method (or constructor, if we decide to go with classes), a `preprocessor` to manipulate the incoming data before formatting. This can be used to replace the `None` with a special character or format floats with a special string formatting etc.

amjith:
>if we can't express the data using Python data types then we have failed. 

What I mean is that, I understand there is no spatial data type in Python, but we just need a way to display it, we are not in the business of manipulating it or doing arithmetic with it. Same goes for binary data. We convert to hex and present the resulting output, we don't interpret the data for them.

meeuw:
@amjith I think you are right, it simplifies things a lot. If I understand you correctly you want to pass only formatted strings to the formatting library? I was on a different track because tabulate does some formatting, but I think we won't be using that right?

I think we only should use this for outputting data for a terminal and nothing else (like csv/tsv/xml/json/html/latex/etc) and use something different (out of this scope)

tsroten:
@meeuw I think you're right that this interface, at least in its first iteration, should be for terminal display only. Let's keep the quality high and complexity low.

I think @amjith is right that this interface should be for formatting rows/columns in a certain way -- any preprocessing should be handled separately. Passing in a preprocessor sounds great.

----

> So the register_output_format will be the constructor that takes in various args to create an object and the format_output method will be called on that object.

@amjith -- are you saying mycli would call `format_output()` on the object? If so, how do you envision mycli getting the specific object in question (e.g. a simple table formatter).

amjith:
@tsroten Here's a pseudo-code: 

Inside the MyCli class's constructor in main.py:

```
    self.output_formatter = OutputFormatter(table_format='simple', table_formatter=tabulator, preprocessor=preprocess)
```

where, `tabulator` is as follows:

```
def tabulator(header, rows, table_format):
    return tabulate(header, rows, table_format)
```

where `preprocessor` is as follows:

```
def preprocessor(header, rows):
    for row in rows:
        for cell in row:
            if not cell
                cell = '<null>'
```

Once we `self.output_formatter` we call `self.output_formatter.format_output(headers, rows)` to convert an array of rows into a table.

That should produce the appropriate the output based on the chosen table format.

If we need to change the table format during runtime, we simply do:

```
self.output_format.table_format = 'fancy'
```

When we call the self.output_formatter.format_output(headers, rows), it will automatically call the appropriate table_formatter with the correct table_format.

meeuw:
@amjith why do you want to pass the preprocessor to OutputFormatter? Do you want a configurable preprocessor (besides the table_format and tabulator?)

It might be a good idea to have preconfigured presets like "fancy" which maps to table_format=fancy, preprocessor=passthrough_preprocessor, tabulator tabulate and "single" which maps to table_format=Single,preprocessor=align_preprocessor, terminaltables etc.

tsroten:
@amjith Thanks for the thoughts/examples. In your example, mycli itself knows about tabulate, tabulate's supported formats, etc.

Here is another possibility (beware, I've been writing PHP all day today, so there are probably some oddities here... :) In this example, mycli doesn't know the details of tabulate, terminaltables, or any future library for output formats.

`main.py`:
```python
from format_output import OutputFormatter

class Mycli:

    # At some point, we create the formatter.
    self.formatter = OutputFormatter()

    # Based on user input `table_format` (e.g. myclirc)
    if table_format in self.formatter.supported_formats():
        self.output_format = table_format

    def format_output(headers, data, **kwargs):
        return self.formatter.format_data(headers, data, self.output_format, **kwargs)
    
    # Whenver we need to format output:
    self.format_output(headers, data)
    # Keyword arguments can be passed to override various options in the module below, e.g. preprocessor, etc.

    # To change output format:
    self.output_format = new_output_format
```


And, in a module `output_formatter.py`:
```python
import tabulate
import terminaltables
# plus any other packages we need for possible output formats

# Maybe a method on the class below?
def tabulate_wrapper(headers, data, table_format=None):
    # Potentially other keyword arguments that would allow for other tabulate-specific logic.
    return tabulate(data, headers, table_format)

class OutputFormatter(object):

    def __init__(self):
        self.output_formats = {}
        self.register_output_format('simple', tabulate_wrapper, table_format='simple')
        self.register_output_format('fancy', self.tabulate_wrapper, table_format='fancy')
        # Any number of keyword arguments can be added to the above call. They will
        # be passed to, in this case, tabulate_wrapper, which can handle tabulate-specific
        # logic. They can be overridden by mycli when it calls `format_output`.

    def register_output_format(self, name, callable, **kwargs):
        self.output_formats[name] = (callable, kwargs)

    def supported_formats(self):
        return self.output_formats.keys()

    def format_output(self, headers, data, format, **kwargs):
        callable, callable_kwargs = self.output_formats[format]
        callable_kwargs.update(kwargs)
        return callable(headers, data, **callable_kwargs)
```

If we need presets like @meeuw mentioned, they can be set through the `register_output_format` function above. E.g.:
```python
register_output_format('simple', tabulate_wrapper, table_format='simple', preprocessor=align_preprocessor)
```
Obviously, `tabulate_wrapper` and any other callables used in output formatting would have to have support for options passed this way.

amjith:
Yes! I like the pseudo implementation. I think once we start writing code things will start to solidify. 

If you want to create a branch and start pushing a prototype we can all collaborate.

tsroten:
Sounds good, I'll get one started so you and @meeuw can start making edits as well.

amjith:
Closing this in favor of #388

PR #388, opened 2017-03-29
Add generic output formatter interface.

tsroten:
## Description

- Adds a generic `OutputFormatter` class with all of tabulate's output formats.
- Moves expanded output to the output formatter.
- Adds terminaltables formats.

## Checklist


tsroten:
@meeuw @amjith This includes tabulate's formats, but not terminaltables. It also doesn't include any of @meeuw's handling of decimals as strings.

tsroten:
I believe @meeuw's number-as-string is working now in this PR since tabulate has `disable_numparse`, which seems like what we want here.

amjith:
This is an excellent refactor. I love seeing the progression of changes as I step through the commits. 

I have a few questions that I'll sprinkle throughout the diffs. Nicely done! 👍

mycli/output_formatter.py
```diff
@@ -0,0 +1,113 @@
...
+                                    preprocessor=override_missing_value,
+                                    missing_value='null')
+        self.register_output_format('tsv', csv_wrapper, delimiter='\t',
+                                    preprocessor=override_missing_value,
+                                    missing_value='null')
+
+        self.register_output_format('expanded', expanded_table,
+                                    preprocessor=override_missing_value,
+                                    missing_value='<null>')
+
```
meeuw: I think register_output_format should be called outside of OutputFormatter (from `MyCli.__init__` ?)

Or else we could also just directly initialize the self._output_formats dict like:
```
self._output_formats = {
    'csv' : csv_wrapper, {preprocessor=override_missing_value, missing_value='null')
...
}
```

And remove register_output_format

tsroten: I think that's a good point.

I favor the idea of `OutputFormatter` setting up the dict when it's initialized. My main reason for liking this is keeping all details of formatting away from MyCli (e.g. wrappers, format names, format packages, etc.).

mycli/output_formatter.py
```diff
@@ -0,0 +1,96 @@
...
+        *format_name* must be a formatter available in `supported_formats()`.
+
+        All keyword arguments are passed to the specified formatter.
+        """
+        function, fkwargs = self._output_formats[format_name]
+        fkwargs.update(kwargs)
+        preprocessor = fkwargs.pop('preprocessor', None)
+        if preprocessor:
+            data = preprocessor(data, **fkwargs)
```
tsroten: Great point! We should definitely include headers in the preprocessor since they are part of the output.

mycli/main.py
```diff
@@ -715,15 +705,41 @@ def get_prompt(self, string):
...
         for result in results:
             title, cur, headers, status = result
-            output = format_output(title, cur, headers, None, table_format)
+            output = self.format_output(title, cur, headers, None)
             for line in output:
                 click.echo(line, nl=new_line)
 
+    def format_output(self, title, cur, headers, status, expanded=False,
+                      max_width=None):
+        table_format = 'expanded' if expanded else self.table_format
```
tsroten: That's a good point. What do you think of having an optional keyword override for table format as well. For instance:

```python
formatter = OutputFormatter(format_name='simple')
formatter.format_output(data, headers)  # formats as simple
formatter.format_output(data, headers, format_name='expanded')  # formats as expanded
```

mycli/output_formatter.py
meeuw: Hmm yes, that's true. Let's keep it this way. I was thinking of letting the user configure more output formatters at runtime but we can always refactor it by then (now it's too far fetched).

tsroten:
@meeuw Thanks for working on the tests!

meeuw:
@tsroten great work on this branch, I couldn't have done it better 😉

mycli/output_formatter.py
tsroten: I do think there is a lot of room for new features in the future. I think you're right that we should keep it simple for now.

tsroten:
@amjith @meeuw One thing we lose when we switch to the official tabulate package is the ability to support leading/trailing whitespace in results, e.g. https://github.com/dbcli/mycli/pull/184

I'm a fan of trying to submit pull requests upstream. In this case the issue [was reported awhile ago](https://bitbucket.org/astanin/python-tabulate/issues/78/whitespace-from-the-cell-values-is) and the maintainer said he'd consider adding an option for this.

I don't think this is a show-stopper on getting rid of our packaged version of tabulate, but I do think we should submit a pull request to tabulate and at least get the ball rolling over there.

Any other thoughts/ideas?

meeuw:
Hmm it seems to me only the leading whitespace is removed. I can think of the following fixes:
1. Add some non whitespace (like adding quotes for strip()-able strings)
2. Make a fix for tabulate and try to get it fixed (until then, maintain a fork)
3. Use another tabulator (terminaltables) and keep tabulate this way

meeuw:
Oh and I'm working on a preprocessor to align floats on the decimal seperator, but that would be kind of impossible when tabulate strips the whitespace 😉

tsroten:
@meeuw Related to your preprocessor, I was working on passing the column types to `OutputFormatter`. For example, you could easily check if a column has decimal points by checking if it's type is in `[FIELD_TYPE.DECIMAL, FIELD_TYPE.FLOAT, FIELD_TYPE.DOUBLE, FIELD_TYPE.NEW_DECIMAL]`

That way, you don't have to try to cast every value to a number to see if a column should be treated a certain way.

I was struggling to find a clean way to make OutputFormatter aware of column types, but not aware of MySQL (e.g. the explicit field types above). I was thinking of passing the Python type name to OutputFormatter, instead of the MySQL type. So, those types above would be passed as `float`.

What do you think? Would that help with a preprocessor?

tsroten:
Or, maybe we can rely on the type of the first row's values?

meeuw:
I think we should rely on the fact that everything is typed to Python types (by pymysql conversions). So if the first column is a int, every value in this column is an int. I'm trying to cherry-pick b0e03972ddaa9c74c75fd8549fb13099bdb911f0 into this PR's branch but some pytests are failing. Trying to cast strings to int's is something we don't need and should avoid.

amjith:
@tsroten @meeuw Are we set on using tabulate going forward? 

Tabulate doesn't have support for multi-line strings (did this change in the newer versions?). The tables gets misformatted when a cell has multi-line string. That was my original impetus in trying to use the new Terminaltables library which has support for multiline strings. 

I also liked the fact that Terminaltables didn't do any kind of data type detection. It only worked with strings. 

That restriction sparked the idea of using preprocessors to format the ints and floats as strings and pass that into Terminaltables. 

We can then choose to format the floats with appropriate padding to allow the decimal alignment.

meeuw:
I think we can keep tabulate for the table formats we already had and use new table formats using terminaltables or whatever fits best. Terminaltables can be added already I guess

amjith:
I'm fine going either way but I think we should have a clear direction on which library to choose. Because picking one or the other will influence the work we have to do to prevent regression.

tsroten:
@meeuw I really like your creative use of the preprocessors.


However, I'm not a fan of the quoted decimals, e.g.:
```
+--------+-----------+
| text   | numeric   |
|--------+-----------|
| abc    | ' 1'      |
| defg   | '11.1'    |
| hi     | ' 1.1'    |
+--------+-----------+
```

I'd prefer the more computationally-intense version of using placeholders that we replace later, like is hinted at in the bug report: https://bitbucket.org/astanin/python-tabulate/issues/78/whitespace-from-the-cell-values-is

But, at that point, I think it's better to vendor the library, use Terminaltables, or override the [`_align_column()` function in tabulate](https://bitbucket.org/astanin/python-tabulate/src/dfbf9245bb29494c49a647bd45b95c8088eba99a/tabulate.py?at=master&fileviewer=file-view-default#tabulate.py-532) (not ideal since it's a "private" function).

tsroten:
I'm going to take this branch and add terminal tables (instead of tabulate) and put it in a different PR to see what we think.

meeuw:
@tsroten true, `quote_whitespaces` is a stinking workaround and it should be removed (but I had to use it to have this working with the current version of tabulate). The `align_decimals` preprocessor can also be used for terminaltables. Anyhow, I don't see why you want to branch this. Terminaltables can be used in this branch using different table formats right? For example we could add a "psql-terminaltables" output format (and slowly port more of the formats from tabulate)?

tsroten:
@meeuw Yeah, the align_decimals preprocessor is working great with terminal tables, so excellent job!

Good point on the branch -- I'll bring the changes here.

tsroten:
Ok, terminaltables appears to be working well with the decimal aligning (without the quoting).

However, the `single` table format (from terminaltables) has some weird issues where it screws up the terminal's formatting. `ascii`, `double`, and `github` appear to work just fine.

tsroten:
Looks like others had the problem with the single table, too, hence the recent PR: https://github.com/Robpol86/terminaltables/pull/42

meeuw:
I've had the same issue with `single` table format; if you don't use less (pager) it's working fine. Maybe we need to add a postprocessor to make the output safe for less?

mycli/output_formatter.py
```diff
@@ -67,23 +67,19 @@ def align_decimals(data, headers, **_):
     """
     pointpos = len(data[0]) * [0]
     for row in data:
-        i = 0
-        for v in row:
+        for i, v in enumerate(row):
```
meeuw: Hey this is just great! I never knew `enumerate` existed, thanks a lot!

amjith:
I've refactored the code in this branch quite a bit. Can I push to this branch or would you like me to push it to a separate branch and review the changes? 

The reason for the refactor is to make the OutputFormat a completely standalone package so we can use it in mycli as well as pgcli. I've also made OutputFormat a little bit more plugin oriented.

/cc @tsroten @meeuw

tsroten:
@amjith Sounds good to me :) Push to this branch

mycli/output_formatter.py
tsroten: @meeuw `enumerate()` is one of those gems that's awesome once you find it. I think I found out about it from an article about idiomatic Python, [like this](http://python.net/~goodger/projects/pycon/2007/idiomatic/handout.html).

amjith:
Before I push the changes, I'm going to leave a bunch of comments on the existing code with my thoughts. Then I'll push my commits. So you'll at least have the reasoning behind some of my decisions.

mycli/output_formatter.py
```diff
@@ -0,0 +1,232 @@
...
+                'preprocessor': (override_missing_value, convert_to_string),
+                'missing_value': '<null>'
+            })
+        }
+        self._format_name = None
+
+        tabulate_formats = ('plain', 'simple', 'grid', 'fancy_grid', 'pipe',
+                            'orgtbl', 'jira', 'psql', 'rst', 'mediawiki',
+                            'moinmoin', 'html', 'latex', 'latex_booktabs',
+                            'textile')
```
amjith: Adding module specific code into OutputFormatter makes this class extremely specialized. We should try to avoid tight coupling as much as possible. 

Since OutputFormatter is supposed to be an interface, we should make this completely agnostic to the implentation of the formatter itself.

mycli/output_formatter.py
```diff
@@ -0,0 +1,232 @@
...
+    if table_format == 'ascii':
+        table = terminaltables.AsciiTable
+    elif table_format == 'single':
+        table = terminaltables.SingleTable
+    elif table_format == 'double':
+        table = terminaltables.DoubleTable
+    elif table_format == 'github':
+        table = terminaltables.GithubFlavoredMarkdownTable
+    else:
+        raise ValueError('unrecognized table format: {}'.format(table_format))
```
amjith: Whenever we see a series of `if/elif/else` section it is better to replace it with a dictionary as follows: 

```
table_formats = { 
    'ascii': terminaltables.AsciiTable, 
    'single': terminaltables.SingleTable,  
    .....
    }

try:
    table = table_formats[table_format]
except KeyError:
    raise ValueError('unrecognized table format: {}'.format(table_format))
```

mycli/output_formatter.py
```diff
@@ -0,0 +1,232 @@
...
+        tabulate_formats = ('plain', 'simple', 'grid', 'fancy_grid', 'pipe',
+                            'orgtbl', 'jira', 'psql', 'rst', 'mediawiki',
+                            'moinmoin', 'html', 'latex', 'latex_booktabs',
+                            'textile')
+        for tabulate_format in tabulate_formats:
+            self._output_formats[tabulate_format] = (tabulate_wrapper, {
+                'preprocessor': (bytes_to_string, align_decimals),
+                'table_format': tabulate_format,
+                'missing_value': '<null>'
+            })
```
amjith: The structure of the `self._output_formats` dictionary is arbitrary. There is an undocumented convention that the first value in the tuple is the formatter function and the second value is a dictionary of kwargs to that formatter function. The dictionary of kwargs also has the `preprocessor` tucked away in it which has a special meaning. 

If we were to replace that tuple with a named tuple it will become self-documenting. We should also move the preprocessor outside of the kwargs dictionary.

mycli/sqlcompleter.py
```diff
@@ -57,7 +59,8 @@ def __init__(self, smart_completion=True):
         self.name_pattern = compile("^[_a-z][_a-z0-9\$]*$")
 
         self.special_commands = []
-        self.table_formats = table_formats()
+        formatter = OutputFormatter()
+        self.table_formats = formatter.supported_formats()
```
amjith: The fact that we are instantiating an `OutputFormatter` object just to get the list of supported formats is a bit of a code smell. 

It forces our `SqlCompleter` class to know about the `OutputFormatter` class. This leads to tight coupling which makes refactoring harder to do in the future. Instead, if we supply the list of supported formats as an argument to the `SqlCompleter`'s constructor then the completer object gets exactly what it needs without any coupling.

tsroten:
@amjith I like the formatter as a package. Thanks for taking the time to help think this through :)

I'll push a few minor edits/style changes in.

mycli/output_formatter/output_formatter.py
```diff
@@ -0,0 +1,97 @@
...
+                                       {'missing_value': '<null>'})
+OutputFormatter.register_new_formatter('tsv', delimiter_adapter,
+                                       delimiter_preprocessors,
+                                       {'missing_value': '<null>',
+                                        'delimiter': '\t'})
+OutputFormatter.register_new_formatter('expanded', expanded_table,
+                                       (override_missing_value,
+                                        convert_to_string),
+                                       {'missing_value': '<null>',
+                                        'delimiter': '\t'})
```
tsroten: @amjith Is expanded supposed to have the `delimiter`? I'm guessing the good ol' copy-paste-error :)

Also, what do you you think about moving `expanded.py` to the formatter package? It's only ever called from the output formatter now. Plus, then we could define the preprocessors in the `expanded.py` module itself.

amjith:
I appreciate you doing the pep8 edits. I wonder if we should standardize on the formatting and use an automated tool like https://github.com/google/yapf to do this in the future. 

If you'd like to have that conversation separately, I'm happy to open a separate ticket for that.

mycli/output_formatter/delimited_output_adapter.py
```diff
@@ -0,0 +1,23 @@
+import contextlib
+import csv
+try:
+    from cStringIO import StringIO
+except ImportError:
+    from io import StringIO
+
+from .preprocessors import override_missing_value, bytes_to_string
+
+supported_formats = ('csv',)
```
tsroten: @amjith Should `tsv` be here, too?

mycli/output_formatter/output_formatter.py
amjith: Yup. That is indeed a copy/paste error. 😞 

I should have moved expanded.py to the output_formatter package. That's a good call.

mycli/output_formatter/delimited_output_adapter.py
amjith: Yes. We should add `tsv` to the list.

tsroten: Hmm, if we add `tsv` to the list, it's not really different from `csv` because there is nothing in the adapter that treats `tsv` differently. In other words, it's up to output_formatter to know to associate `tsv` with `\t`.

I wonder if the `delimiter_adapter` should take a format argument that maps format to delimiter.

tsroten:
Personally, I'd love to make mycli more PEP8-compliant. I've never used an auto-formatter on a project before, just a linter in travis CI. Sounds exciting :)

I do think we should do that in a different context from this PR, though. Thanks for bringing that up!

amjith:
There is a bug when there are no rows to display. 

```
mysql root@localhost:test> create TABLE test_table (id INTEGER);
Query OK, 0 rows affected
Time: 0.052s
mysql root@localhost:test> SELECT * FROM test_table;
list index out of range
```

This only happens when the table is completely empty and there is nothing to display.

tsroten:
@amjith Great catch! I just pushed a fix for that.

mycli/output_formatter/output_formatter.py
```diff
@@ -0,0 +1,98 @@
...
+    'format_name preprocessors formatter formatter_args')
+
+
+class OutputFormatter(object):
+    """A class with a standard interface for various formatting libraries."""
+
+    _output_formats = {}
+
+    def __init__(self, format_name=None):
+        """Register the supported output formats."""
```
amjith: This doc string is outdated. Looks like I missed it in my refactor.

amjith:
Looks like we lost the ability to auto-complete on the `\T ` command.

tsroten:
@amjith Good catches. Looks like the background refresher wasn't setting the table formats when it created the `SQLCompleter` object. Should be fixed now.

amjith:
I'm on linux (Ubuntu). When I choose 'Single' or 'fancy_grid' as the table format the output is completely garbled. Is anyone else seeing this? 

It works fine with other table formats.

tsroten:
@amjith Yes, that happens to me as well on MacOS. I'll try to track down what's going on later today.

meeuw:
@tsroten is has to do with the less pager and escape codes. If you use cat as a pager it works.

amjith:
Are we ready to merge this or should we wait for cli_helpers to land?

tsroten:
@amjith I'm up for it being merged now. Then, when CLI Helpers is released we can update mycli to use it instead.

PR #403, opened 2017-04-14
remove temporary hack

meeuw:
## Description
remove temporary hack for sqlparse

## Checklist

### DIFF
--- a/mycli/packages/completion_engine.py
+++ b/mycli/packages/completion_engine.py
@@ -34,7 +34,8 @@ def suggest_type(full_text, text_before_cursor):
     # partially typed string which renders the smart completion useless because
     # it will always return the list of keywords as completion.
     if word_before_cursor:
-        if word_before_cursor[-1] == '(' or word_before_cursor[0] == '\\':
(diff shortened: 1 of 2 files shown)


tsroten:
Thanks for staying on top of these types of changes :)

mycli/packages/completion_engine.py
```diff
@@ -28,32 +28,27 @@ def suggest_type(full_text, text_before_cursor):
...
-            else:
-                parsed = sqlparse.parse(
-                        text_before_cursor[:-len(word_before_cursor)])
+    # If we've partially typed a word then word_before_cursor won't be an empty
+    # string. In that case we want to remove the partially typed string before
+    # sending it to the sqlparser. Otherwise the last token will always be the
+    # partially typed string which renders the smart completion useless because
+    # it will always return the list of keywords as completion.
+    if word_before_cursor:
+        if word_before_cursor[-1] == '(' or word_before_cursor[0] == '\\':
```
tsroten: I know this was already like this, but we should change this to use `str.startswith()` and `str.endswith()`.

Issue #402, opened 2017-04-14
Switch to a simpler prompt format when length threshold is hit


tsroten:
The default prompt can become laboriously long when connecting to certain hosts (e.g. AWS). pgcli has a feature that shortens the prompt to the database name when:
1. The prompt is the default prompt (i.e. the user has not customized the prompt)
2. The prompt is over 30 characters

Here is the pgcli issue: https://github.com/dbcli/pgcli/issues/587
Here is the pgcli pull request: https://github.com/dbcli/pgcli/pull/639

It'd be nice to add this to mycli.

tsroten:
In testing this, 30 characters is quite short. On a normal installation, where connecting to localhost for development, it is too easy to hit this limit. For example, this is four characters too long:
```
mysql root@localhost:my_cool_app> 
```

I'm going to recommend we try for 45 for now. That will give the user plenty of room before their host is cut out of their prompt. Of course, if they customize their prompt so it doesn't match the default, then they don't need to worry about this at all.

johnksterling:
I submitted a PR for this and defaulted to 45 as recommended.

tsroten:
Thanks @johnksterling! Great work!

tsroten:
mycli version 1.10.0 is now out which adds this. You can install/upgrade by running `pip install --upgrade mycli`

PR #404, opened 2017-04-15
#402 use abbreviated prompt if the default causes it to be huge

johnksterling:
## Description
Add logic to the prompt definition to use a very short prompt with only the db name if the default prompt is too long once generated.

## Checklist

- [X ] I've added this contribution to the `changelog.md`.


amjith:
Looks good to me. :keyboard:

PR #405, opened 2017-04-19
Switch from pycryptodome to cryptography.

tsroten:
## Description
This pull request switches mycli from using [pycryptodome](https://github.com/Legrandin/pycryptodome) to using [cryptography](https://github.com/pyca/cryptography).

pycryptodome cannot be installed alongside of pycrypto, which is a problem from some Linux distributions like Fedora. By using cryptography, which has a totally different API, we should not run into that problem. Plus, cryptography appears to be better maintained and better tested.

## Checklist

- [x] I've added this contribution to the `changelog.md`.
- [x] I've added my name to the `AUTHORS` file (or it's already there).

PR #406, opened 2017-04-19
Switch from pycryptodome to pycryptodomex.

tsroten:
## Description
This is another approach to solve the issue that #405 addresses. This uses pycryptodomex so that pycrypto installations do not conflict with pycryptodome.

## Checklist


tsroten:
Closing in favor of #405.

PR #410, opened 2017-04-20
mv authors

meeuw:
## Description
This patch is used by Fedora to get AUTHORS and SPONSORS in /usr/lib/pythonN/site-packages/mycli instead of /usr/lib/pythonN/site-packages.

## Checklist


tsroten:
Thanks for figuring this out. However, I think the best practice is to not include AUTHORS and SPONSORS in `package_data` at all.

They should be in the manifest file.

tsroten:
Oh wait, we need them for the credits when we load mycli. Hmm. Mycli really shouldn't load things outside of it's package. But, I don't like the AUTHORS file not being in the project root directory.

Let's think through a better way to do this.

tsroten:
Ok, what about `data_files` where we can control where they are installed? https://docs.python.org/3.4/distutils/setupscript.html#installing-additional-files

tsroten:
It appears that `data_files` is not to be used anymore.

I'm not seeing a

meeuw:
I admit this is a bit lame, but is it acceptable to have AUTHORS.rst and SPONSORS.rst to redirect to the text files in our mycli directory?

amjith:
Instead of adding a link inside the AUTHORS to mycli/AUTHORS why not just create a symlink? 
We move AUTHORS and SPONSORS inside mycli folder and then symlink it to the outside folder?

meeuw:
I don't know if it is important but last time I've used Windows/NTFS it didn't support symlinks.

And I don't know how github handles symlinks, I'm pushing a test commit...

meeuw:
Yikes, github doesn't follow symlinks:
https://github.com/meeuw/dmtest

tsroten:
@meeuw I removed the `project_root` variable. This pull request looks good to me. Anything holding it back now? Or are we good to merge?

meeuw:
I've thought hard about alternatives but I cannot think of anything better than this, so -- good to merge!

PR #415, opened 2017-04-27
behave quit mycli nicely

meeuw:
## Description

Before this patch mycli is killed by expect and it the coverage data cannot be written.

## Checklist

### DIFF
--- a/test/features/environment.py
+++ b/test/features/environment.py
@@ -8,6 +8,8 @@ import db_utils as dbutils
 import fixture_utils as fixutils
 import pexpect
 
+from steps.wrappers import run_cli, wait_prompt
+
 
 def before_all(context):
     """Set env parameters."""
@@ -71,6 +73,11 @@ def after_all(context):
 
 def before_step(context, _):
     context.atprompt = False
+
+
+def before_scenario(context, _):
+    run_cli(context)
+    wait_prompt(context)
 
 
 def after_scenario(context, _):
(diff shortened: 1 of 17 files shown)


test/features/environment.py
```diff
@@ -68,12 +69,27 @@ def after_all(context):
...
+
+
 def after_scenario(context, _):
     """Cleans up after each test complete."""
 
     if hasattr(context, 'cli') and not context.exit_sent:
-        # Terminate nicely.
-        context.cli.terminate()
+        # Quit nicely.
+        if not context.atprompt:
```
tsroten: I'm wondering -- why are we expecting a certain output after scenarios have already been run? Wouldn't it be more explicit if the scenario ended with `and we wait for prompt` since that step does the same thing as this code?

Then we can drop the `atprompt` piece of the context and just rely on the scenario itself.

j-bennet: I think the cleanup code makes sense in `after_scenario`, because we don't want to make the scenarios too boilerplate. In fact, I was thinking that the first 2 steps that are present in every scenario:

```
When we run dbcli
and we wait for prompt
```

could also be pulled out into the `before_scenario` handler. The scenarios are already becoming verbose, especially the CRUD ones that include multiple steps.

tsroten: Ah ok, I can see what you mean.

@meeuw - do you want to add the `before_scenario` code into this pull request as well since it's so closely related?

meeuw: done 😄

tsroten: @meeuw Thanks! Looks like the tests are failing for Python 3.

meeuw: :angry: yes, I had to add some excludes for pytest :smile:

PR #417, opened 2017-04-29
behave fix clean up

meeuw:
## Description
In an earlier commit I've changed the current working directory and
the removal of a temporary file didn't take that into account.

## Checklist

### DIFF
--- a/test/features/steps/iocommands.py
+++ b/test/features/steps/iocommands.py
@@ -9,10 +9,12 @@ from behave import when, then
 @when('we start external editor providing a file name')
 def step_edit_file(context):
     """Edit file with external editor."""
-    context.editor_file_name = 'test_file_{0}.sql'.format(context.conf['vi'])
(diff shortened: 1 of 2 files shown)


test/features/steps/iocommands.py
```diff
@@ -1,6 +1,7 @@
 # -*- coding: utf-8
 from __future__ import unicode_literals
 import os
+import os.path
```
tsroten: `os.path` is automatically imported when `os` is imported (in line 3).

tsroten:
Rather than all the `os.path.join` calls, why not just keep the code as it was before with `../` prefixed to the name. Then in the one place you need the filename, call `os.path.basename()`?

Then, the filename doesn't need to be a list anymore either, which is a little bit unintuitive.

meeuw:
Hmm I didn't think of that, thanks!

test/features/steps/iocommands.py
meeuw: right, thanks!

PR #418, opened 2017-04-29
test using behave the tee command

meeuw:
## Description
test using behave the tee command

## Checklist


tsroten:
Like the other PR, let's keep the filename as an actual filename and use `os.path.basename()` when we need the name part only.

meeuw:
Thanks! Fixed.

PR #419, opened 2017-04-30
Add test for handling bug with sqlparse 0.2.2.

meeuw:
## Description
sqlparse==0.2.2 gives:
AttributeError: 'NoneType' object has no attribute 'normalized'

Added a regression test to test_completion_engine.py

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -12,6 +12,7 @@ Internal Changes:
 * Test using behave the tee command (Thanks: [Dick Marinus]).
 * Behave fix clean up. (Thanks: [Dick Marinus]).
 * Added a regression test for sqlparse >= 0.2.3 (Thanks: [Dick Marinus]).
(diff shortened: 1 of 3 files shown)


tsroten:
Nice catch!

sqlparse 0.2.3 has been out less than two months. I'm wondering about how this will affect various Linux distributions. What do you think?

meeuw:
That a valid point, I'll have a look at the compatibility code which I've removed (see if it can catch this error).

I think Linux distributions will follow (because not much project depend on a specific version of sqlparse)

meeuw:
I've reverted my commit to remove the temporary hack and added some code to make pytest happy

tsroten:
@meeuw Looking good! Looks like the changelog edit is still in there

tsroten:
@meeuw Thanks for working out the tests on this one :)

PR #420, opened 2017-05-01
Do not add time from multiple queries together.

tsroten:
## Description

This fixes a timing bug when you run queries from a file. The time outputted at the end was incorrect because it added time together like a factorial for each query.

See https://github.com/dbcli/pgcli/issues/575 for more information.

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,11 @@
 TBD
 ===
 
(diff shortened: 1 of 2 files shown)


amjith:
:lo

PR #421, opened 2017-05-02
Auto-calculate reserved space for completion menu.

tsroten:
## Description

This is influenced by pgcli: https://github.com/dbcli/pgcli/pull/642

~~This adds the ability for the user to control how many lines are reserved for the completion menu.~~

This automatically calculates how much space to reserve for the completion menu. The maximum number of lines is 8. The minimum is 0. The ratio used is 0.2, e.g. a terminal window with 20 rows would reserve 4 for the completion menu.

## Checklist

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -113,7 +113,6 @@ class MyCli(object):
         self.less_chatty = c['main'].as_bool('less_chatty')
         self.cli_style = c['colors']
         self.wider_completion_menu = c['main'].as_bool('wider_completion_menu')
-        self.min_num_menu_lines = c['main'].as_int('min_num_menu_lines')
         c_dest_warning = c['main'].as_bool('destructive_warning')
         self.destructive_warning = c_dest_warning if warn is None else warn
         self.login_path_as_host = c['main'].as_bool('login_path_as_host')
@@ -613,7 +612,7 @@ class MyCli(object):
                 processor=HighlightMatchingBracketProcessor(chars='[](){}'),
                 filter=HasFocus(DEFAULT_BUFFER) & ~IsDone()
             )],
-            reserve_space_for_menu=self.min_num_menu_lines
+            reserve_space_for_menu=self.get_reserved_space()
         )
         with self._completer_lock:
             buf = CLIBuffer(always_multiline=self.multi_line, completer=self.completer,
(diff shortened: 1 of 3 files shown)


amjith:
Can we do better than this? 

I'm sure we can detect the terminal size using prompt-toolkit or click. Can we adjust the reserved space based on the available space? 

ps: I'm just not a fan of adding new config entries.

tsroten:
Hmm, yeah I'll give it a shot :)

tsroten:
@amjith Ok, I've set it at a ratio of .2, with a maximum of 8 reserved lines. Eight lines is the current behavior (which is probably on the high-end, hence it being the new max). We might need to tweak the .2 ratio at some point, but in my casual testing, it appeared to be a reasonable value.

I'm curious about your distaste for adding config options. Could you elaborate a bit?

mycli/main.py
```diff
@@ -747,6 +749,13 @@ def format_output(self, title, cur, headers, status, expanded=False,
 
         return output
 
+    def get_reserved_space(self):
+        """Get the number of lines to reserve for the completion menu."""
+        reserved_space_ratio = .2
+        max_reserved_space = 8
+        _, height = click.get_terminal_size()
+        return min(int(height * reserved_space_ratio), max_reserved_space)
```
amjith: Is 0.2 the right ratio? It requires you to have 10 lines before you get a completion menu that is 2 lines tall. 

Also what happens if the height comes back as 4 lines or lesser? Does it not show any completion menu at all?

tsroten: At 4 or less, it comes back as 0. At 9 or less it comes back as 1, but that is then given to the toolbar. So, right now the completion window goes away at less than 10 lines. I guess that's a little harsh :)

I think I set it that way out of personal preference. For me, if I have a window that is 9 lines or less, I'd like to have at most, 1 line for completions (probably even zero).

But, I can see the argument for trying to keep completions present as much as possible. Let me try .4 and see how it looks.

At 4 lines or less (where 1 goes to the toolbar), we probably shouldn't have completions at all. Would you agree?

In master, mycli currently ceases to function below 9 lines.

tsroten:
@amjith Ok, it's now at .45 ratio and uses `round()` instead of `int()`.

Height | Reserved Lines
------------ | -------------
2 | 0
3 | 0
4 | 1
5 | 2
6 | 3
7 | 3
8 | 4
9 | 4
10 | 5
11 | 5
12 | 5
13 | 6
14 | 6
15 | 7
16 | 7
17 | 8

amjith:
Thank you for running that experiment and choosing a sane default. I was trying to get mysql working on my laptop to test this and finally failed. :disappointed: 

But I'm glad you found the sweet spot. :+1:

Issue #35, opened 2015-05-29
Make edit command default to previous query


fpietka:
I am wondering, on psql when triggering `\e` on a new line, it will edit previous query. Here we have to append it to the query itself. Is that a design choice?
I must say that if not, I'd like to have that feature :grin:

tsroten:
It looks like mysql does this as well. It can tell the difference between a command and a query, so the editor will only ever display the last query.

For example, here is a sample narrative:
- `use my_db`
- `edit` -- editor opens blank
- `select count(*) from my_table;`
- `edit` -- editor opens with previous query
- `use my_db_number_2`
- `edit` -- editor opens with same query as before (`select count(*)...`)

amjith:
I'm not sure if we need to emulate exactly what mysql does. I feel like we can just edit whatever was the previous command. 

Simply because there are many corner cases like what if the previous command was a multi-query separated by semi-colons and one of them happens to be `use my_db`? 

At least for the first attempt let's do just previous command and so special cases.

tsroten:
Agreed 👍

tsroten:
@fpietka This is included as of version 1.11.0. If you installed via `pip`, you can upgrade by running `pip install --upgrade mycli`.

Thanks for the idea!

fpietka:
Great, thanks!

PR #425, opened 2017-05-03
Default to last command in external editor.

tsroten:
## Description

This pull request mimics some of the behavior in the official mysql client.
- When you type `\e` without a query in front of it, mycli use the last-run query as the default text in the editor.
- If you type `\e` and there were no queries run before it, then the editor will be blank.

This addresses https://github.com/dbcli/mycli/issues/35

## Checklist

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -419,9 +419,10 @@ class MyCli(object):
         saved_callables = cli.application.pre_run_callables
         while special.editor_command(document.text):
             filename = special.get_filename(document.text)
-            sql, message = special.open_external_editor(
-                filename, sql=document.text,
-                default_text=self.get_last_query() or '')
+            query = (special.get_editor_query(document.text) or
+                     self.get_last_query() or '')
+            sql, message = special.open_external_editor(filename,
+                                                        sql=query)
             if message:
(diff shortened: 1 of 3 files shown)


mycli/main.py
```diff
@@ -419,8 +419,9 @@ def handle_editor_command(self, cli, document):
         saved_callables = cli.application.pre_run_callables
         while special.editor_command(document.text):
             filename = special.get_filename(document.text)
-            sql, message = special.open_external_editor(filename,
-                                                          sql=document.text)
+            sql, message = special.open_external_editor(
+                filename, sql=document.text,
+                default_text=self.get_last_query() or '')
```
meeuw: can `or ''` be removed here? I think `None` should be handled correctly in `special.open_external_editor`.

actually I'm not sure about `open_external_editor`, why does it need to decide to open `sql` or `default_text`?

tsroten: Yeah, I'd like to streamline this a bit. I'll think through it.

`open_external_editor` is where the \e is removed, which `handle_editor_command` doesn't know about.

tsroten:
@meeuw Ok, I think I've addressed your points. What do you think?

meeuw:
:+1: :+1:

PR #427, opened 2017-05-06
add behave to tox.ini

meeuw:
## Description
add behave to tox.ini

## Checklist

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,13 @@
 [tox]
 envlist = py27, py33, py34, py35, py36
+
 [testenv]
 deps = pytest
     mock
-commands = py.test
+    pexpect
+    behave
+    coverage
+commands = python setup.py test
+passenv = PYTEST_HOST
+    PYTEST_USER
+    PYTEST_PASSWORD
--- a/changelog.md
+++ b/changelog.md
@@ -29,6 +29,7 @@ Internal Changes:
 * Added a regression test for sqlparse >= 0.2.3 (Thanks: [Dick Marinus]).
 * Reverted removal of temporary hack for sqlparse (Thanks: [Dick Marinus]).
 * Add setup.py commands to simplify development tasks (Thanks: [Thomas Roten]).
+* Add behave tests to tox (Thanks: [Dick Marinus]).
 
 1.10.0:
 =======


tox.ini
```diff
@@ -1,6 +1,13 @@
...
-commands = py.test --doctest-modules --doctest-ignore-import-errors
+deps = pytest: pytest
+    pytest: mock
+    behave,pytest: pexpect
+    behave: behave
+    behave: coverage
+changedir = behave: test
+commands = pytest: py.test
+    behave: behave
+passenv = PYTEST_HOST
```
tsroten: Should we also pass the password, user, port, and charset?

tox.ini
```diff
@@ -1,6 +1,13 @@
 [tox]
-envlist = py27, py33, py34, py35, py36
+envlist = {py27,py33,py34,py35,py36}-{pytest,behave}
```
tsroten: This creates ten different tox environments. I think the `--{env, env}` is typically used for testing against different versions of packages. In our case, pytest and behave can be run in the same tox environment. This would speed up testing.

Why don't we add pytest and behave (`cd test && behave`) as the commands to run in every testenv? Then, for `deps`, just do `-rrequirements-dev.txt`

tsroten:
I think it's a great idea to have behave in the tox tests :) I like making testing as easy as possible.

tox.ini
meeuw: sure, good point

tox.ini
meeuw: I'll have to check if && is possible, I thought it was impossible to chdir in `tox.ini` (only with `changedir=`)

meeuw: `cd test && behave` doesn't seem to work:
```
py35-behave runtests: commands[0] | cd test && behave
WARNING:test command found but not installed in testenv
  cmd: /usr/bin/cd
  env: /home/meeuw/git/mycli/.tox/py35-behave
```
But this isn't necessary with your setup.cfg fixes, let's wait until that's merged.

tsroten:
@meeuw Once you clear up the conflicts and reduce the number of tox environments, I'll merge this in :)

meeuw:
done

meeuw:
Oh I was thinking of adding tox to setup.py, but I don't think that's very usefull (`python setup.py --tox` vs just running `tox` isn't very helpfull). Maybe we can add something to `CONTRIBUTING.md` about tox.

tsroten:
@meeuw I think it's in there:
> Mycli supports Python 2.7 and 3.3+. You can test against multiple versions of Python by running:
> ```
> $ ./setup.py test -all
> ```

Is that what you mean?

meeuw:
😮 yes

PR #428, opened 2017-05-06
Update release script

tsroten:
## Description

This makes the following changes to the release script:
- Adds a docstring.
- Uses `io.open` since we can specify a file encoding.
- Makes the version reader not execute Python code from a file (i.e. no `ast.literal_eval`).
- Adds click as a development dependency and uses its `click.confirm()` to simplify the code in the release script.
- Removes the register with PyPI step since this is not needed any longer with PyPI.
- Removes the old release procedure text file.

## Checklist

- ~I've added this contribution to the `changelog.md`.~ (Not needed)


tsroten:
@amjith Is it safe to remove the `create-deb.sh` script and the `debian` folder?

amjith:
You can remove all of that. 

They were all added when I thought I could created deb/rpm packages. Since then I've come to realize it is more work than I anticipated and there are people who make official packages for mycli and pgcli so those files are unnecessary. 

I'm not looking at the TODO file either. So it can be removed.

amjith:
Changes look good to me. 👍

PR #429, opened 2017-05-06
Add development tasks to setup.py.

tsroten:
## Description

- This adds ~a Makefile~ `setup.py` commands, that makes it trivial to test/lint.
- Pins the specific version of pep8radius needed for mycli.
- Updates the development guide to use the ~make~ `setup.py` commands.

Examples:

```bash
./setup.py test  # Run unit tests and integrated CLI tests
./setup.py test --all  # (or -a) Run tox
./setup.py test --coverage  # (or -c) Run unit tests/integrated tests and measure code coverage
./setup.py test --dry-run  # (or -n) Run test runners without executing tests.
./setup.py test --all --dry-run  # (or -a -n) Run tox, create environments, but don't execute tests
```
```bash
./setup.py lint  # Check for PEP 8 violations against master
./setup.py lint --fix  # (or -f) Fix PEP 8 violations.
./setup.py lint --branch=test  # (or -b test) Compare against branch `test`
```

Both of these commands support `-q/--quiet` and `-v/--verbose`.

`-q/--quiet`, `-v/--verbose`, and `-n/--dry-run` are global `setup.py` arguments that we can choose to respect or not to respect.

## Checklist


j-bennet:
@tsroten Do Makefiles work on Windows?

tsroten:
@j-bennet Good question. Sometimes they work via nmake. But, not always. I haven't tested this one.

It looks like the only other part of our development instructions that might be *nix-specific is the source command for virtualenv (should be `\path\to\env\Scripts\activate` on Windows) -- I'll update the file with those instructions.

While I like Makefile's, I'm open to adding the same functionality in the form of setuptools commands so you could do commands like this:
```bash
$ ./setup.py lint
$ ./setup.py lint-fix
```

What do you think?

j-bennet:
@tsroten I'm not on Windows, I think @darikg is. I like the Makefile approach, but I suggest we also keep a set of "non-makefile" instructions in DEVELOP.rst.

j-bennet:
@tsroten I also never used setuptools commands or pyinvoke. I'll look into that.

tsroten:
@meeuw Good ideas! I'll try to address some of those this week.

j-bennet:
Looking into setuptools commands, [this documentation section](http://setuptools.readthedocs.io/en/latest/setuptools.html#adding-commands) seems to suggest that custom commands should be mapped to functions in some python module. I wonder if they can be mapped to arbitrary shell commands also. Did you ever try that @tsroten? Or were you thinking of adding a "service" python module to encapsulate the commands?

tsroten:
@j-bennet I think it has to be a python function (or class inherited from one of `distutils` commands. I envisioned a `dev_tasks.py` (or something like it) that `setup.py` would load as commands.

Basically, another file that would function like a Makefile, but it would be the thing that setup.py loads to get the commands. The nice thing about it is that our `setup.py` is not cluttered with development logic. It simply declares the commands.

j-bennet:
Ah great, that's about what I thought. I like the idea.

tsroten:
@meeuw @j-bennet Ok, I added two commands to `setup.py` and hooked it up to Travis. Let me know what you think. Here are some examples:

```bash
./setup.py test  # Run unit tests and integrated CLI tests
./setup.py test --all  # (or -a) Run tox
./setup.py test --coverage  # (or -c) Run unit tests/integrated tests and measure code coverage
./setup.py test --dry-run  # (or -n) Run test runners without executing tests.
./setup.py test --all --dry-run  # (or -a -n) Run tox, create environments, but don't execute tests
```
```bash
./setup.py lint  # Check for PEP 8 violations against master
./setup.py lint --fix  # (or -f) Fix PEP 8 violations.
./setup.py lint --branch=test  # (or -b test) Compare against branch `test`
```

Both of these commands support `-q/--quiet` and `-v/--verbose`.

`-q/--quiet`, `-v/--verbose`, and `-n/--dry-run` are global `setup.py` arguments that we can choose to respect or not to respect.


@meeuw - I'll work on the text of the development guide next.

meeuw:
I really like this! Would it be a good idea to add a single command which will run all tests that will be run by travis (tox+pep8radius) ?

tsroten:
@meeuw It looks like you can chain them together :) That's a nice surprise.

```bash
./setup.py test --coverage lint
```

I'll add that to travis.

tsroten:
@meeuw Ok, I added `master` as an argument to the `git checkout` in the development guide along with a note that we branch off master.

I also added pushing to GitHub, creating a pull request, test environment variables, testing coverage, and deactivating the virtual environment. 

Unless you or @j-bennet have other thoughts/changes in mind, I think this might be ready to merge.

tsroten:
Hmm, apparently, the contributing file has to be `.txt` or `.md` to show up on the create new pull request page.

https://github.com/blog/1184-contributing-guidelines

I'll change the syntax to Markdown later today.

tsroten:
Ok, converted it to Markdown. @meeuw - if you feel like this is ready, can you merge it in?

meeuw:
I can wait using this, great work!

PR #431, opened 2017-05-07
Clean up testing config files.

tsroten:
## Description

This moves the configuration from `.coveragerc`, `pytest.ini`, and `conftest.py` into the existing `setup.cfg` so we don't have our configuration scattered across so many files.

`conftest.py` is useful when you need to ignore files during testing based on dynamic criteria (like Python version or operating system) -- which we don't need to do (as of right now).

## Checklist

- ~I've added this contribution to the `changelog.md`.~ not needed


tsroten:
Looks like the code coverage isn't quit working, yet.

meeuw:
This is a nice clean up! I think something goes wrong because the `pytest-cov` doesn't do the parallel execution or it doesn't combine the output correctly.

tsroten:
@meeuw Thanks! Remove pytest-cov still has the same code coverage problem. It looks like the gap is in code that behave would have executed. I wonder if the `COVERAGE_PROCESS_START` doesn't support `setup.cfg`. I'll look into it a bit more.

meeuw:
I've pinpointed the problem to `test/features/environment.py` where you've changed .coveragerc to setup.py. I think coverage doesn't recognize `setup.cfg`'s `[coverage:run]` for `COVERAGE_PROCESS_START`.

After running behave you should see a lot of `.coverage*` files (if it's using `parallel = True`), now there is only one `.coverage` (`parallel = False`).

tsroten:
Switching it to `run` works. However, I can't get it to work when running behave from the project directory, only the `test` directory. Very odd.

meeuw:
That is caused by `cwd='..'` in https://github.com/dbcli/mycli/blob/1136a1f88c4fe333348b9031341fc5ed2acbbcc5/test/features/steps/wrappers.py#L41 :blush:

tsroten:
Yes! thanks for pointing that out. Now that I looked, I had *tons* of `.coverage*` files outside of my `mycli` folder because it was storing them outside of the project directory. Switching that to an absolute path now works.

meeuw:
:+1: do you want pytest-cov back? Maybe it works if we first run the behave tests and then pytest-cov?
I think it's a good idea to sent a bug report for python-coverage (for .coveragerc / setup.cfg), please let me know if you've already done that.

tsroten:
@meeuw Sure, added it back. I think this is ready to be merged, what do you think?

meeuw:
:+1: very nice

PR #433, opened 2017-05-09
output once special command

meeuw:
## Description
output once special command (related to https://github.com/dbcli/mycli/pull/353)

the usage is a bit tricky, it can only be used like this:
`> \once testfile ; select 1`
or
`select 1; \o testfile`

this doesn't work:
```
> \once testfile
> select 1
```
(gives empty file)

## Checklist

~- [ ] I've added my name to the `AUTHORS` file (or it's already there).~

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -7,6 +7,7 @@ Features:
 * Handle reserved space for completion menu better in small windows. (Thanks: [Thomas Roten]).
 * Display current vi mode in toolbar. (Thanks: [Thomas Roten]).
 * Opening an external editor will edit the last-run query. (Thanks: [Thomas Roten]).
+* Output once special command. (Thanks: [Dick Marinus]).
 
 Bug Fixes:
 ----------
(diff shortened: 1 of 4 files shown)


tsroten:
@meeuw Clean and well-written :)

I'm wondering if there is some other variable we could use in `iocommands.py` to indicate whether or not the once file has been written to (e.g. `once_file_written_to`). Then, we only close it if it has been written to -- that way we can have the next command's output written, even if it's not executed at the same time:
```sql
> \T csv
> \o outfile.csv
> select * from users limit 100;
```

meeuw:
That's a good idea, to make it conditional on output! This is much cleaner, could you please check again?

tsroten:
It works great :) I think I'll definitely be using it.

Looking at its code and the code you wrote for the `tee` command -- a lot of it is identical. Can we create some functions to make the code less duplicated?

tsroten:
@meeuw Is this good now? I'm happy with it :)

meeuw:
sure, please merge 😄

Issue #248, opened 2016-03-23
Remove newline outputted with blank query


jinleileiking:
```
➜  ~ git:(master) ✗ mycli              
Version: 1.5.2
Chat: https://gitter.im/dbcli/mycli
Mail: https://groups.google.com/forum/#!forum/mycli-users
Home: http://mycli.net
Thanks to the contributor - Whitane Tech
mysql jinlei04@dbbk-baeond.xdb.all.serv:(none)> 








Time: 0.000s
mysql jinlei04@dbbk-baeond.xdb.all.serv:(none)> nopa
(1064, u"You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'nopa' at line 1")
mysql jinlei04@dbbk-baeond.xdb.all.serv:(none)> nopager;
Pager disabled.
Time: 0.000s
mysql jinlei04@dbbk-baeond.xdb.all.serv:(none)> 

Time: 0.000s
mysql jinlei04@dbbk-baeond.xdb.all.serv:(none)> 

```

jinleileiking:
version is 16-03-22.  github main version.

jinleileiking:
Press enter will have a blank line before `Time: 0.000s`

tsroten:
@jinleileiking Thanks for reporting this! Is the expected behavior this?
```
mysql jinlei04@dbbk-baeond.xdb.all.serv:(none)> 
Time: 0.000s
mysql jinlei04@dbbk-baeond.xdb.all.serv:(none)>
```

jinleileiking:
@tsroten YES!

tsroten:
@jinleileiking This is fixed as of [version 1.11.0](https://pypi.org/project/mycli/). If you installed via `pip`, you can upgrade by running `pip install --upgrade mycli`.

Thanks for the idea!

PR #436, opened 2017-05-11
Do not display time and empty line for blank query.

tsroten:
## Description
This pull request makes mycli work more like a shell when you just hit enter without typing something. This addresses https://github.com/dbcli/mycli/issues/248.

Before:
```
mycli> 

Time: 0.00
mycli >
```

After:
```
mycli>
mycli>
```

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -12,6 +12,7 @@ Bug Fixes:
 ----------
 
 * Fixed incorrect timekeeping when running queries from a file. (Thanks: [Thomas Roten]).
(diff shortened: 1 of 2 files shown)


meeuw:
Seems to work fine, maybe it's better to `if not document.text.strip()` to ignore all white space.

tsroten:
@meeuw Good idea, I'll change that.

tsroten:
@meeuw Ok, this should be good now.

PR #443, opened 2017-05-26
Special helper for showing `SHOW CREATE TABLE`

tanzoniteblack:
## Description

I regularly want to list both the columns in the table as well as the foreign
keys and in depth information about indexes (I'm used to the behavior of `\d`
from psql/pgcli), `SHOW CREATE TABLE`, which lists the create table statement is
the closest that mysql seems to have for this view.

```
mysql user@some_db> \d django_admin_log
***************************[ 1. row ]***************************
Table        | django_admin_log
Create Table | CREATE TABLE `django_admin_log` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `action_time` datetime(6) NOT NULL,
  `object_id` longtext,
  `object_repr` varchar(200) NOT NULL,
  `action_flag` smallint(5) unsigned NOT NULL,
  `change_message` longtext NOT NULL,
  `content_type_id` int(11) DEFAULT NULL,
  `user_id` int(11) NOT NULL,
  PRIMARY KEY (`id`),
  KEY `django_admin__content_type_id_c4bce8eb_fk_django_content_type_id` (`content_type_id`),
  KEY `django_admin_log_user_id_c564eba6_fk_some_user_id` (`user_id`),
  CONSTRAINT `django_admin__content_type_id_c4bce8eb_fk_django_content_type_id` FOREIGN KEY (`content_type_id`) REFERENCES `django_content_type` (`id`),
  CONSTRAINT `django_admin_log_user_id_c564eba6_fk_some_user_id` FOREIGN KEY (`user_id`) REFERENCES `some_user` (`id`)
) ENGINE=InnoDB AUTO_INCREMENT=1205 DEFAULT CHARSET=latin1

Time: 0.039s
```

## Checklist

### DIFF
--- a/test/features/fixture_data/help_commands.txt
+++ b/test/features/fixture_data/help_commands.txt
@@ -3,6 +3,7 @@
 +-------------+-------------------+------------------------------------------------------------+
 | \G          | \G                | Display current query results vertically.                  |
 | \dt         | \dt [table]       | List or describe tables.                                   |
+| \dt+        | \dt+ [table]      | List tables or show create table.                          |
 | \e          | \e                | Edit command with editor (uses $EDITOR).                   |
 | \f          | \f [name]         | List or execute favorite queries.                          |
 | \fd         | \fd [name]        | Delete a favorite query.                                   |
--- a/changelog.md
+++ b/changelog.md
@@ -8,6 +8,7 @@ Features:
 * Display current vi mode in toolbar. (Thanks: [Thomas Roten]).
 * Opening an external editor will edit the last-run query. (Thanks: [Thomas Roten]).
 * Output once special command. (Thanks: [Dick Marinus]).
+* Add special command to show create table statement. (Thanks: [Ryan Smith])
 
 Bug Fixes:
 ----------
--- a/mycli/AUTHORS
+++ b/mycli/AUTHORS
@@ -50,6 +50,7 @@ Contributors:
   * chainkite
   * Michał Górny
   * Terje Røsten
+  * Ryan Smith
 
 Creator:
 --------
--- a/mycli/packages/completion_engine.py
+++ b/mycli/packages/completion_engine.py
@@ -109,7 +109,7 @@ def suggest_special(text):
     if cmd in ['\\f', '\\fs', '\\fd']:
         return [{'type': 'favoritequery'}]
 
-    if cmd in ['\\dt', '\\d']:
+    if cmd in ['\\dt', '\\dt+']:
         return [
             {'type': 'table', 'schema': []},
             {'type': 'view', 'schema': []},
--- a/mycli/packages/completion_engine.py
+++ b/mycli/packages/completion_engine.py
@@ -109,7 +109,7 @@ def suggest_special(text):
     if cmd in ['\\f', '\\fs', '\\fd']:
         return [{'type': 'favoritequery'}]
 
-    if cmd in ['\\dt']:
+    if cmd in ['\\dt', '\\d']:
         return [
             {'type': 'table', 'schema': []},
             {'type': 'view', 'schema': []},
--- a/mycli/packages/special/dbcommands.py
+++ b/mycli/packages/special/dbcommands.py
@@ -23,7 +23,7 @@ def list_tables(cur, arg=None, arg_type=PARSED_QUERY):
         return [(None, None, None, '')]
 
 
-@special_command('\\d', '\\d [table]', 'List tables or show create table.', arg_type=PARSED_QUERY, case_sensitive=True)
+@special_command('\\dt+', '\\dt+ [table]', 'List tables or show create table.', arg_type=PARSED_QUERY, case_sensitive=True)
 def list_or_show_create_tables(cur, arg=None, arg_type=PARSED_QUERY):
     if arg:
         query = 'SHOW CREATE TABLE {0}'.format(arg)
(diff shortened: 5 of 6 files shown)


amjith:
@tanzoniteblack I like the spirit of this PR to provide more info about the table using the `SHOW CREATE TABLE` output. 

Would it make more sense to show this in addition to the columns? Similar to how postgres does it? 

For example:

```
mysql root@localhost:world> \dt city
+-------------+----------+--------+-------+-----------+----------------+
| Field       | Type     | Null   | Key   | Default   | Extra          |
|-------------+----------+--------+-------+-----------+----------------|
| ID          | int(11)  | NO     | PRI   | <null>    | auto_increment |
| Name        | char(35) | NO     |       |           |                |
| CountryCode | char(3)  | NO     | MUL   |           |                |
| District    | char(20) | NO     |       |           |                |
| Population  | int(11)  | NO     |       | 0         |                |
+-------------+----------+--------+-------+-----------+----------------+
CREATE TABLE `city` (
  `ID` int(11) NOT NULL AUTO_INCREMENT,
  `Name` char(35) NOT NULL DEFAULT '',
  `CountryCode` char(3) NOT NULL DEFAULT '',
  `District` char(20) NOT NULL DEFAULT '',
  `Population` int(11) NOT NULL DEFAULT '0',
  PRIMARY KEY (`ID`),
  KEY `CountryCode` (`CountryCode`),
  CONSTRAINT `city_ibfk_1` FOREIGN KEY (`CountryCode`) REFERENCES `country` (`Code`)
) ENGINE=InnoDB AUTO_INCREMENT=4080 DEFAULT CHARSET=latin1
```

amjith:
The way to achieve that is to set the last value in the tuple to the results of the `show create table `. The last value corresponds to the footer (or status) of the table. So it will be printed just below the table.

tanzoniteblack:
I like the idea of having the nicely printed columns with the extra info in the footer, but my concern is that for many tables showing the whole create table after pretty printing the columns is going to be excessively large.

meeuw:
Good idea, I use `show create table` a lot (because it's more complete). I don't think we should combine the output from `show create table` and `describe` because it contains a lot of duplicate information.

Maybe the output from `show create table` can be shortened by only showing the create statement. (and removing `**** row ****` and `Table` etc.). It would also be nice if the output would be colorized (don't know if that's easy).

amjith:
> I like the idea of having the nicely printed columns with the extra info in the footer, but my concern is that for many tables showing the whole create table after pretty printing the columns is going to be excessively large.

I can certainly understand the concern. Since we're taking inspiration from psql, why not borrow the idea of using a plus sign after the `\dt` like `\dt+` to print the results of `show create table` output? 

That'll maintain a sense of consistent behavior between pgcli and mycli. Thoughts?

tanzoniteblack:
Using `dt+` for what I have instead of `\d` is a great idea. And at some point, maybe I'll see if I can't replicate psql's footer info on the regular `\d` command.

I'm about to leave town for a few days, so I won't be able to make that minor change for a few days. I have permissions for maintainers to make changes to my branch, so please feel free to do that yourself if you don't want to wait.

test/features/fixture_data/help_commands.txt
```diff
@@ -2,12 +2,14 @@
...
 +-------------+-------------------+---------------------------------------------------------+
 | \G          | \G                | Display results vertically.                             |
+| \d          | \d [table]        | List tables or show create table.                       |
 | \dt         | \dt [table]       | List or describe tables.                                |
 | \e          | \e                | Edit command with editor. (uses $EDITOR)                |
 | \f          | \f [name]         | List or execute favorite queries.                       |
 | \fd         | \fd [name]        | Delete a favorite query.                                |
 | \fs         | \fs name query    | Save a favorite query.                                  |
 | \l          | \l                | List databases.                                         |
+| \once       | \o [-o] filename  | Output for the next SQL command only to FILENAME        |
```
simnim: Is this supposed to be part of the PR?

tanzoniteblack: This command was included in a recent commit, but wasn't added to this file, I figured I might as well add it? (And by that, I really mean, I just copied and pasted this output from running `\?`, saw this change, and figured it might as well go in?)

tanzoniteblack:
I've updated this to use `\dt+` instead of `\d`

tsroten:
@tanzoniteblack Thanks for the update! This is looking really good.

The only thing I think it needs now is to combine the `\dt+` and `\dt` commands into the same function. Kind of like how pgcli does it.

The help from pgcli uses brackets to show that the `+` is optional:
```
| \dt[+] [pattern]                     | List tables.                                  |
```

pgcli uses this to detect if it's "verbose":
```
verbose = '+' in command
```

So, both commands can be combined into one function:
```
@special_command('\\dt', '\\dt+ [table]', 'List tables or show create table.', arg_type=PARSED_QUERY, case_sensitive=True)
```

amjith:
This is a valiant effort but I think a few pieces got missed during translation. I'm going to merge this in and refactor the code a bit to conform to what @tsroten has requested.

amjith:
@tanzoniteblack I'd like to thank for sticking with it and making changes to the PR when requested. :tropical_drink:

tsroten:
Thanks @tanzoniteblack! This is definitely a useful command. We appreciate your contribution to mycli 👨‍🚀

Issue #445, opened 2017-05-27
Attempting to quit sometimes raises NotImplementedError


tsroten:
If a user attempts to quit immediately after running the external editor, a `Not Yet Implemented` error is displayed.

```
mysql root@localhost:(none)> \e
mysql root@localhost:(none)> quit
Not Yet Implemented.
mysql root@localhost:(none)> quit
Goodbye!
```

tsroten:
This is now fixed as of [version 1.11.0](https://pypi.org/project/mycli/1.11.0/).

PR #446, opened 2017-05-29
Make quit command use special command logic.

tsroten:
## Description
This fixes an issue where the quit command was not being recognized once the editor was called.

This addresses #445.

## Checklist


amjith:
`:q` can no longer be used to quit mycli. 

😭

tsroten:
@amjith Lol. I did notice that :) Do you think we have enough Vim users to warrant keeping it around?

amjith:
Nah. That was supposed to be an Easter egg. No need to keep it.

j-bennet:
What!!!

Issue #376, opened 2017-03-19
How do I set the options that the LESS pager uses?


mikeerickson:
Just installed `1.9.0` and there seems to be difference

In 1.8.1, the console would scroll with the various entered commands as usual

![image](https://cloud.githubusercontent.com/assets/183153/24078200/68002ffe-0c24-11e7-83ee-c7f31b649e14.png)

But with 1.9.0, it appears it clears screen rather "aggressively"  After each command, it clears the screen and then output results

![image](https://cloud.githubusercontent.com/assets/183153/24078205/98c922bc-0c24-11e7-8418-203480c42472.png)

Is there a setting for this?
If not, how do I go about reverting 1.8.1

meeuw:
Hmm that's stange, I think it is related to this commit (c3ed9a67aeb1a30ddf0ed67203def3aa4c9753ab) where os.environ['LESS'] only get set when it's empty. Could you please check your LESS environment variable by running `echo $LESS` in your shell?

johnksterling:
Agree - let me know what you see and I can help troubleshoot.  Hopefully you it is working as designed - i.e. passing through your default less settings.

golovan:
Same issue. $LESS gives -R for me. export LESS="-RXF" fixes it. Placed it into my .zshrc file for now. :)

meeuw:
@golovan, thanks for reporting! Do you know why you had LESS="-R" ? What operating system are you using (and optionally which toolchain (cygwin/msys/homebrew etc))?

huwenchao:
@meeuw, I got the same issue with @golovan, my toolchain is OS  X EI Capitan, zsh, python environment is anaconda, installed mycli with pip.

tonydeng:
setting $LESS=-RXF

glasala:
Happened to me, solved setting "-RXF".
I had
```
me:~> echo $LESS
-M -I -R
```

amjith:
Hey Folks. We're considering making some changes to the pager stuff in mycli. Can you answer a few questions to help us decide? 

Can you tell us what version of mycli you're using? You can find out the version by:

```
mycli -v
```

Can you also tell us the output of the following commands in your shell (not inside mycli)

```
echo $PAGER
echo $LESS
```

glasala:
Hello,
mycli Version: 1.10.0
pager: less
-RXFI

tsroten:
There is now a page in the documentation devoted to the pager: http://mycli.net/pager

[...]

PR #450, opened 2017-06-03
less warning

meeuw:
## Description
Warning if `-F` isn't set in `LESS`, add an option to config to snooze the warning.

## Checklist


amjith:
@meeuw Why do we need this warning? 

We have reasonable defaults in our config file. The only way a user can override the default is if they have set a custom `PAGER` variable or a custom `LESS` variable. 

If they already have that set then the user knows exactly what he's doing and we shouldn't be warning about a scenario where their choice is different from ours.

meeuw:
Hmm, since we're using the `LESS` variable there have been multiple issues from users complaining about the less pager (always have to close less after each command using `q`). I get the feeling they don't set `LESS` without `-F` on purpose (might be a system default from anaconda?).

We could also enforce the use of `-F` instead of a warning; the user experience is killed when you have to manually quit less after each command.

tsroten:
@amjith @meeuw I think you both have good points. We need to be careful about adding extra warnings/config options. We also should recognize that there appears to be a common complaint about mycli's behavior.

I've been perusing the pager-related issues on mycli and pgcli. Here are some thoughts:

**1.** mycli defaults to always using a pager. You have to enable it in mysql. So, for many users, mycli's behavior can be considered odd.

**2.** People are not used to using a pager for REPL output. This, I think, is the crux of the issue. For many uses (e.g. piping output to `less`) you might not want the `-F` option because you're explicitly telling `less` to read your output -- you do not want it to automatically quit, no matter what. But, in a REPL it makes sense that you would not want the pager to fire at all if the results fit in the terminal window.

With number 2, I think I'm saying that solely relying on `less` by default is maybe not a great idea because people often haven't used pagers with an application like mycli. They might have `$LESS` options set up that end up not being appropriate for day-to-day mycli usage.

-----------

So, what if we did this:
1. If the user does not have the `pager` config option explicitly set, we try to be intuitive:
    - If the output fits on the screen, output it directly and don't pass it through click's pager output command. In other words, don't rely on `less`'s `-F` option.
    - If the output does not fit on the screen, we output it through click's pager command.
2. If the user has the pager disabled (currently via `skip-pager`), then we never use a pager.
3. If the user has the pager option explicitly set, then we always output via the pager.

What this gives us is an intuitive behavior with zero extra config options. It solves the relatively-common complaint about having to press `q` all the time. Those that like to configure things can still do so via `$LESS` and the `pager` config option.

amjith:
> If the user does not have the pager config option explicitly set, we try to be intuitive:

> If the output fits on the screen, output it directly and don't pass it through click's pager output command. In other words, don't rely on less's -F option.

If the user does not have the pager config explicitly set then we're already using `less` with the options `-XRF`. Which includes the `-F` option.

So the user complaints are not really due to the absence of `-F`. I'm beginning to suspect that the users who complain were running old version of mycli which used to have the `-S` option which would enable the pager if the output was too wide to fit on the screen (not just too long). But that is just a theory. 

I think we need a little bit of data gathering. Let's revisit all the issues regarding the pager disabling and ask the question about what version of mycli are they using and how big was the output of their query.

tsroten:
@amjith Revisiting sounds good to me 😄 . Thanks for starting up that conversation on those threads.

> If the user does not have the pager config explicitly set then we're already using less with the options -XRF. Which includes the -F option.

If the user didn't explicitly set the `pager` option, that doesn't mean we are using `-XRF`. For example, if I don't set the `pager` option, but have `$LESS` defined, then mycli won't use `-XRF`. That, I think, is the problem most of these people are having. They've set `$LESS` in the past for whatever reason, and it wasn't in the context of a REPL. So, they didn't even worry about `-F`, and now they run into a seemingly-odd behavior in mycli.

Especially in comments like [this one](https://github.com/dbcli/mycli/issues/376#issuecomment-288063595) (that have five thumbs up), where the problem was due to their `$LESS` environment variable. In that case it appears they desired the `-F` option. Also note the two other people with the same issue in that thread that didn't do thumbs up.

amjith:
I didn't think about the scenario of missing $PAGER but a valid $LESS. For that case can we just append the `-F` if it is not already present in their $LESS?

I know click used to do that appending in the past, but they've now changed their behavior to only set it if $LESS is missing (similar to our behavior). 
https://github.com/pallets/click/blob/master/click/_termui_impl.py#L315

tsroten:
@amjith The whole reason we started using the user's `$LESS` is because someone brought up a great reason why they *did not* want `-F` -- [the reserved lines for autocompletion](https://github.com/dbcli/mycli/issues/288).

amjith:
:disappointed: 

You have a point. I think your suggestion to not use the pager at all if the contents will fit on a single page is the best way forward.

meeuw:
Yes, I'm also a proponent for manually doing the `-F` functionality in mycli, I'll try to make a new PR/branch for that.

PR #451, opened 2017-06-03
Refactor \dt+ command.

amjith:
## Description

* Add the ability to handle verbose flag in special command. 
* Convert `\dt+` to use the verbose flag.

/cc @tanzoniteblack this PR refactors your code. After I did the refactor I realized why you chose to implement `\dt+` as a separate function. It was because never had the ability to handle the `+` sign in special commands. But pgcli has that ability. So I chose to port over that code from pgcli. Once again thank you for this clever idea to use pgcli style verbosity in mycli. :smile: 

Reviewer: @tsroten

### DIFF
--- a/mycli/packages/completion_engine.py
+++ b/mycli/packages/completion_engine.py
@@ -94,7 +94,7 @@ def suggest_type(full_text, text_before_cursor):
 
 def suggest_special(text):
     text = text.lstrip()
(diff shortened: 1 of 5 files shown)


tsroten:
@amjith Looks good 🥇  Thanks!

amjith:
@tsroten Thank you for going through and fixing up the deficiencies in the PR.

PR #452, opened 2017-06-04
Feature/conditional pager

meeuw:
## Description
Only use pager if output doesn't fit.

I think `MyCli.output` should be refactored to take a line generator instead of a string, we're now joining the lines and later splitting the lines. But that's out of this PRs scope.

The hardcoded `12` margin should also be variable, does someone know how to get this value?

## Checklist


mycli/main.py
```diff
@@ -643,18 +640,26 @@ def one_iteration(document=None):
...
-            self.logfile.write(text)
+            self.logfile.write(utf8tounicode(output))
             self.logfile.write('\n')
-        click.echo_via_pager(text)
+
+        size = self.cli.output.get_size()
+
+        pager = False
+        for i, line in enumerate(output.splitlines()):
+            if len(line) > size.columns or i > (size.rows - 12):
```
tsroten: Can you explain the `12` a bit? We use a maximum of `8` for the completions (which includes the bottom toolbar). Then, you need to consider the prompt.

```
mycli.get_reserved_space() + mycli.prompt.get_prompt().count('\n') + 1
```

tsroten:
@meeuw Thanks for jumping on top of this one!

I agree we'll need to look at the output logic later on, not now 😄 

So, one thing is that your `if special.is_pager_enabled() and pager:` part is overriding the user's preference. If they set `pager` in `.my.cnf`, then the pager should *always* be used, even if the output is small enough for the screen. We only want to do the size logic if the user did not explicitly set the pager config option.

```
if cnf['pager'] or (special.is_pager_enabled() and pager):
```

The only thing is that we don't have easy  access to `cnf['pager']`, we might need to modify `mycli.configure_pager` to store `cnf['pager']` so we can reference it as needed.


Also, what about changing the `pager = True` and `pager = False` to a function call so it's only called if needed -- kind of like this:

```
if cnf['pager'] or (special.is_pager_enabled() and not output_fits_on_screen(output)):
```

tsroten:
@meeuw Thanks for updating the pull request. I just updated the tests to use an explicit pager so they will pass.

tsroten:
@meeuw This looks good to me. Am I clear to merge this or do you have anything else you're doing on this branch?

Thanks again for adding this. I think it will definitely help out the mycli newcomers 😄

meeuw:
Please merge, next thing is refactoring `MyCli.output` to take a line generator instead of a string.

meeuw:
Oh, the screen size detection doesn't work very well for me, but it was broken even with less so I think this is an improvement anyhow.

tsroten:
@meeuw For the screen detection, I think we forgot about the timing output, which is an extra line. Also, we're comparing `i` (index), which starts at `0`, not `1`. So, between those two things, we're missing two lines.

meeuw:
oh for me it's two off... how much is it for you?

tsroten:
Yeah, two as well :)

 It should be fixed in #454, if you get time to take a look.

PR #454, opened 2017-06-05
Fix screen margin detection.

tsroten:
## Description

This pull request fixes a few things with the automatic pager behavior.
- The screen margin includes the timing row, if enabled.
- Refactors the code so that output isn't passed to `splitlines`, if an explicit pager is enabled. Makes long queries slightly faster.
- Makes echo'ed statements not display in the pager, if explicitly set.
- Doesn't write to the `tee` file twice for each command.

## Checklist

- [ ] ~I've added this contribution to the `changelog.md`.~ not needed

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -129,7 +129,8 @@ class MyCli(object):
             try:
                 self.logfile = open(os.path.expanduser(c['main']['audit_log']), 'a')
             except (IOError, OSError) as e:
-                self.output('Error: Unable to open the audit log file. Your queries will not be logged.', err=True, fg='red')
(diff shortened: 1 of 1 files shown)


meeuw:
Hmm I'll have to dig into it but the pager is now always on for me. I'm intentionally using `LESS=-RX`.

tsroten:
@meeuw Check your `~.my.cnf`'s `[client]` section. That's the config that should determine if the pager is always on or not.

meeuw:
I think I was using the wrong venv :blush: you did a nice clean up.

PR #457, opened 2017-06-07
Test sqlexecute.run return values, not formatted output.

tsroten:
## Description

This finishes a `TODO` in the test code. For the `test_sqlexecute.py` tests, we now test the direct return values from `sqlexecute.run()` instead of the formatted output.

The formatted output is tested in CLI Helpers (and to some extent in behave).

The `test_sqlexecute.py` tests should, for the most part, target the actual execution of a query and its return values.

## Checklist

- ~~I've added this contribution to the `changelog.md`.~~ Not needed.

### DIFF
--- a/test/test_sqlexecute.py
+++ b/test/test_sqlexecute.py
@@ -1,64 +1,45 @@
 # coding=UTF-8
 
+import os
+
 import pytest
 import pymysql
-import os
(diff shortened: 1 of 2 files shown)


amjith:
Looks great. 🚴

Issue #387, opened 2017-03-25
mycli fails to return multiple result sets


sampablokuper:
MySQL and MariaDB are capable of returning multiple result sets, for example [from a stored procedure](https://forums.mysql.com/read.php?98,365143,365804#msg-365804).

In such cases, the `mysql` client displays all the returned result sets, as expected, but `mycli` only displays the first one.

meeuw:
Thanks for reporting, I can reproduce this by:

Create stored procedure:
```
DELIMITER //
CREATE PROCEDURE dmtest
()
BEGIN
  SELECT 1;
  SELECT 2;
END //
DELIMITER ;
```

Call stored procedure
```
call dmtest;
```

And it returns:
```
+---+
| 1 |
+---+
| 1 |
+---+
1 row in set
Time: 0.002s
```

It should return both 1 and 2

pymysql supports multiple resultsets by calling `Cursor.nextset()`, I'll check if this can easily be done.

dbolser-ebi:
Why aren't you calling this a bug?

tsroten:
@dbolser-ebi Mostly because it appears that this was not an oversight, but intentionally left out to make the initial mycli release simpler.

I could easily be wrong :) @amjith

dbolser-ebi:
Hehe, I guess it's semantics.

j-bennet:
This is now fixed in master. To install from master:

`pip install -U https://github.com/dbcli/mycli/archive/master.zip`

PR #459, opened 2017-06-08
Display multiple result sets.

tsroten:
## Description

This displays all result sets returned by a query (e.g. stored procedure).

See #387 for more information.

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -9,6 +9,7 @@ Features:
 * Opening an external editor will edit the last-run query. (Thanks: [Thomas Roten]).
 * Output once special command. (Thanks: [Dick Marinus]).
 * Add special command to show create table statement. (Thanks: [Ryan Smith])
+* Display all result sets returned by stored procedures (Thanks: [Thomas Roten]).
 
 Bug Fixes:
 ----------
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -509,6 +509,7 @@ class MyCli(object):
                 start = time()
                 res = sqlexecute.run(document.text)
                 successful = True
+                result_count = 0
                 for title, cur, headers, status in res:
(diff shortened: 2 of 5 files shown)


j-bennet:
Looks good and seems to work great! One small thing: it would be nice to have a newline between results sets. For example, I have this procedure that selects from two tables:

```
mysql> delimiter //
mysql> create procedure t1()
    -> begin
    ->   select * from aaa;
    ->   select * from bbb;
    -> end//
```

calling in `mysql`:

```
mysql> call t1();
+------+
| xxx  |
+------+
| boo  |
+------+
1 row in set (0.00 sec)

+------+------+
| id   | name |
+------+------+
|    1 | foo  |
+------+------+
1 row in set (0.00 sec)

Query OK, 0 rows affected (0.00 sec)
```

calling in mycli:

```
mysql root@localhost:test> call t1()
+-------+
| xxx   |
|-------|
| boo   |
+-------+
1 row in set
Time: 0.001s
+------+--------+
|   id | name   |
|------+--------|
|    1 | foo    |
+------+--------+
1 row in set
Time: 0.001s
mysql root@localhost:test>
```

I think an extra newline in between would add readability.

meeuw:
Really nice! The only problem I see is that only one result is outputted for the `\once` command.

tsroten:
@j-bennet @meeuw Thanks for both of those points :) You're definitely right, I'll fix those things.

tsroten:
@j-bennet @meeuw Can you check this again? There should be an extra newline between result sets. And, multiple results should end up in the once file.

meeuw:
:ok_hand: great

j-bennet:
Looks perfect now. 👍

Issue #455, opened 2017-06-05
Don't output status text to pager, output file, or tee file.


tsroten:
When you execute a query, various status lines are outputted to undesirable places like the pager, output file (i.e. `\once` command), and tee file.

Example status text that shouldn't be outputted:
- `2 rows in set`
- `Changed table format to []`
- `You are now connected to database [] as user []`

tsroten:
This is now fixed as of [version 1.11.0](https://pypi.org/project/mycli/1.11.0/).

PR #460, opened 2017-06-08
Echo status line

tsroten:
## Description

This calls `mycli.echo` for outputting the status line. This means the status line won't end up in the pager, once file, or tee file.

See https://github.com/dbcli/mycli/issues/455 for more information.

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -11,6 +11,7 @@ Features:
 * Add special command to show create table statement. (Thanks: [Ryan Smith])
 * Display all result sets returned by stored procedures (Thanks: [Thomas Roten]).
 * Add current time to prompt options (Thanks: [Thomas Roten]).
+* Output status text in a more intuitive way (Thanks: [Thomas Roten]).
 
 Bug Fixes:
 ----------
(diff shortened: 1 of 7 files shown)


tsroten:
@meeuw -- I'm having trouble getting this to work with pexpect/behave. Can you lend a hand?

meeuw:
sure, pexpect/behave was using exactly this functionality 😄 

`expect_pager` should be changed to `expect_exact` for all status lines and ```
 wrappers.expect_exact(
        context, context.conf['pager_boundary'] + '\r\n', timeout=5)
```

should be removed in `crud_database.py`.

Still would it be nice if we could recognize the status line boundaries some other way in pexpect/behave, maybe be using some fancy formatting?

tsroten:
@meeuw Thanks! I just pushed those changes you mentioned, but it's having trouble recognizing the status lines with `expect_exact()`.

meeuw:
Ah, it seems there's no `\n` for `click.echo` and the `\r` is prefixed with some escape codes. If you remove `\r\n` from the expect_exact lines it only seems to fail on `iocommands` -- I'm working on that.

tsroten:
@meeuw Thanks!! If you think this is ready, feel free to merge it.

PR #461, opened 2017-06-09
output via generator

meeuw:
## Description
output via generator to prevent:
- buffering of the complete output
- joining and splitting the output

## Checklist


mycli/main.py
```diff
@@ -790,10 +801,12 @@ def format_output(self, title, cur, headers, status, expanded=False,
                 formatted = self.formatter.format_output(
                     rows, headers, format_name='vertical')
 
-            output.append(formatted)
+            if isinstance(formatted, str) or isinstance(formatted, unicode):
```
tsroten: You might want to try adding `from mycli.encodingutils import text_type` to `main.py` so you can use that here.

mycli/main.py
```diff
@@ -41,6 +41,8 @@
 from .lexer import MyCliLexer
 from .__init__ import __version__
 
+from mycli.encodingutils import text_type
```
tsroten: Looks like I told you wrong - you can add this to line 39 where we are already importing from `encodingutils` ;) Sorry about that!

mycli/main.py
```diff
@@ -803,7 +810,9 @@ def format_output(self, title, cur, headers, expanded=False,
                 formatted = self.formatter.format_output(
                     rows, headers, format_name='vertical')
 
-            output.append(formatted)
+            if isinstance(formatted, text_type):
+                formatted = formatted.splitlines()
```
tsroten: When is `formatted` not a `text_type`? I think CLI Helpers always returns a `text_type`.

tsroten:
@meeuw Thanks for looking for more ways to improve mycli 😄 

So, I'm wondering, what's the advantage of operating on the line level? I'm having trouble understanding what that gets us versus what we have now.

Tabulate and Terminaltables don't return lists or iterables, so it's not like we can process the lines separately. This seems like it calls quite a bit of code for every line.

belartodmarinus:
Terminaltables is capable of returning an iterable, I'm working on a fix for cli_helpers. Terminaltables still needs to process the complete result set before outputting (to calculate the column width). But for csv/json/xml/tsv/wide output it's possible to iterate the database cursor and format/output the data without buffering the complete result set in memory.

meeuw:
Terminaltables is capable of returning an iterable, I'm working on a fix for cli_helpers. Terminaltables still needs to process the complete result set before outputting (to calculate the column width). But for csv/json/xml/tsv/wide output it's possible to iterate the database cursor and format/output the data without buffering the complete result set in memory.

tsroten:
@meeuw Cool, sounds good to me 😄. Should I review this pull request? Or, wait until the CLI Helpers pull request?

meeuw:
Related to https://github.com/dbcli/cli_helpers/pull/13

mycli/main.py
```diff
@@ -688,15 +691,36 @@ def output(self, output, status=None):
...
-            special.write_once(output)
-
-            if (self.explicit_pager or
-                    (special.is_pager_enabled() and not self.output_fits_on_screen(output, status))):
-                click.echo_via_pager(output)
-            else:
-                click.secho(output)
+            size = self.cli.output.get_size()
+
+            margin = self.get_reserved_space() + self.get_prompt(self.prompt_format).count('\n') + 1
```
tsroten: I get why you're not calling `output_fits_on_screen` any longer (it compares the entire output), but let's rework that function so we can call it to get the margin so the margin code isn't directly in the `output()` function.

E.g. `get_output_margin` or something.

mycli/main.py
```diff
@@ -799,7 +823,10 @@ def format_output(self, title, cur, headers, expanded=False,
                 formatted = self.formatter.format_output(
                     rows, headers, format_name='vertical')
 
-            output.append(formatted)
+            if not isinstance(formatted, types.GeneratorType):
+                formatted = formatted.splitlines()
```
tsroten: @meeuw, so if I understand this correctly, this line won't be needed once the CLI helpers changes are made, right?

tsroten:
@meeuw I'd like to get this in for our next mycli release (so mycli will be ready whenever the next CLI Helpers is released with the generator change).

I'll test it out this weekend. Other than that, the only change I think would be good is [this one](https://github.com/dbcli/mycli/pull/461#discussion_r122721114).

mycli/main.py
meeuw: exactly

mycli/main.py
meeuw: I think I accidentally removed `output_fits_on_screen` when I was resolving the merge conflicts, thanks for noticing, I'll try to fix it as soon as possible 😄

meeuw: Hmm that's a good idea, fixed this.

tsroten:
So, one thing that's come up in my testing is that the output is always outputted to the pager, even if it fits on the screen. That's how mycli used to output results, but we recently changed it to only output via the pager if it doesn't fit (or there is the explicit pager).

meeuw:
Ah you're right. I think I fixed it but I'd like to add additional tests.

meeuw:
@tsroten I've added the additional tests but I think they could be a little bit more Pythonic and/or use more pytest features.

meeuw:
I'm thinking of using [this](https://stackoverflow.com/questions/1055360/how-to-tell-a-variable-is-iterable-but-not-a-string/44328500#44328500) to check if a generator returned from cli_helpers.

meeuw:
I've created a PR: https://github.com/pallets/click/pull/815 to allow outputting to the pager using a generator

mycli/main.py
```diff
@@ -799,7 +823,10 @@ def format_output(self, title, cur, headers, expanded=False,
                 formatted = self.formatter.format_output(
                     rows, headers, format_name='vertical')
 
-            output.append(formatted)
+            if isinstance(formatted, (str, text_type)):
```
tsroten: Did you mean to hardcode `str` in here? Isn't `text_type` enough?

tsroten:
@meeuw Good job on the tests 😄 . I'm interested to see how Click responds to the PR. I think it's a good idea. Hopefully they'll accept it.

I'll take a look at this today and then let's get this merged :smile:.

mycli/main.py
meeuw: This is a workaround for Python 2.7, without str pytest breaks at `test/test_main.py:133`

tsroten: Oh! I think it's fixed by CLI Helpers branch `fix_unicode_problems`. It's related to the standard-library `csv` module in 2.7. It returns bytes instead of unicode, while the [`backports.csv`](https://github.com/ryanhiebert/backports.csv) library emulates Python 3 and uses unicode.

Let's just leave this as is for now.

mycli/main.py
```diff
@@ -690,15 +690,39 @@ def output(self, output, status=None):
...
+                special.write_once(line)
+
+                if fits or output_via_pager:
+                    # buffering
+                    buf.append(line)
+
+                if len(line) > size.columns or i > (size.rows - margin):
+                    # doesn't fit, flush buffer
+                    fits = False
+                    if special.is_pager_enabled:
```
tsroten: I think this needs to be a function call. If I set `skip-pager` to `on`, the pager is always *on*, not *off*.

mycli/main.py
```diff
@@ -690,15 +690,39 @@ def output(self, output, status=None):
...
+
+            fits = True
+            buf = []
+            output_via_pager = self.explicit_pager and special.is_pager_enabled()
+            for i, line in enumerate(output, 1):
+                self.log_output(line)
+                special.write_tee(line)
+                special.write_once(line)
+
+                if fits or output_via_pager:
```
tsroten: If the output doesn't fit and it's not going to be output via a pager, then we lose all the rows after `fits` is set to `False`.

This is the case when the pager is disabled.

mycli/main.py
meeuw: ah yes, good catch

mycli/main.py
meeuw: hmm how can I reproduce this? It seems to work for me if the pager is enabled and/or disabled.

meeuw:
@tsroten thanks, could you please update the tests or tell me how, "if the output doesn't fit and it's not going to be output via a pager, then we lose all the rows after fits is set to False?"

tsroten:
@meeuw Ok, I added a test that is failing. You can replicate it by:
1. opening mycli
2. turn off the pager by running `nopager`.
3. run a query that will give you results longer or wider than the console.

You should see the results cut off because the output stops appending to `buf`.

meeuw:
You're the best 😄 thanks a lot! Fixed.

mycli/main.py
```diff
@@ -701,15 +701,43 @@ def output(self, output, status=None):
...
+
+            margin = self.get_output_margin(status)
+
+            fits = True
+            buf = []
+            output_via_pager = self.explicit_pager and special.is_pager_enabled()
+            for i, line in enumerate(output, 1):
+                self.log_output(line)
+                # line needs to be casted to unicode for click with Python 2.7
+                special.write_tee(text_type(line))
```
tsroten: We need to fix whatever is not returning unicode. Do you know what it is? I know the CSV part of CLI Helpers wasn't at one point, but that should be fixed in the version that mycli requires.

If we can't fix it, then, we'll need to convert to unicode for the call to `write_once()` as well (line 715).

mycli/main.py
```diff
@@ -825,14 +853,23 @@ def sanitize(col):
                 cur, headers, format_name='vertical' if expanded else None,
                 column_types=column_types,
                 **output_kwargs)
-            first_line = formatted[:formatted.find('\n')]
+
+            if isinstance(formatted, (str, text_type)):
```
tsroten: I think we can drop `str` now that the unicode problems in CLI Helpers were fixed.

meeuw:
👍 thanks, the cast isn't required anymore.

tsroten:
@meeuw Good work! I think this is ready to merge. If you could just move the changelog line up to the pending release, we'll be all set. 

I wonder if there is a way to word this change so that the end-user understands it.

meeuw:
I've rephrased the changelog, English isn't my first language, could you please review? Feel free to change the text.

tsroten:
@meeuw Ok, I reworded the changelog to be a bit shorter. It indicates that this is preliminary work to use less memory (preliminary, since CLI Helpers is still returning a string and not a generator).

If you agree with the change, then go ahead and merge this in! It'll be good to get this one in finally. Sorry it took so long!

👍 🚀

meeuw:
I'm not in a hurry, thanks for your help!

PR #464, opened 2017-06-15
Use standard library logging.NullHandler.

tsroten:
## Description

Use the standard library `logging.NullHandler` since we don't need to support Python 2.6 any longer.

## Checklist

- [ ] ~~I've added this contribution to the `changelog.md`.~~ N/A


j-bennet:
👍

PR #466, opened 2017-06-22
Use appropriate CLI Helpers formatting.

tsroten:
## Description

- Disable tabulate's numparse.
- Uses tabulate's preserve whitespace (so whitespace is not stripped).
- Aligns decimal columns on decimal place (this preprocessor needs to be explicitly included since the pgcli change to CLI Helpers).

## Checklist

- ~~I've added this contribution to the `changelog.md`.~~ Not needed, part of the CLI Helpers switch


meeuw:
nice, seems to improve alignment!

PR #470, opened 2017-06-26
Fix auto vertical mode

tsroten:
## Description

The auto vertical mode was using unformatted output (with its best guess of table delimiters) to calculate whether or not the output was too big. This makes it use the first line of the formatted output.

It's not fool-proof because of TSV and CSV formats where the first line (header), might be shorter than other lines, but the previous method of measuring width had the same problem. So, this improves in some but not all ways.

Also, adds behave tests for auto vertical mode.

## Checklist

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -793,9 +793,10 @@ class MyCli(object):
             rows = list(cur)
             formatted = self.formatter.format_output(
                 rows, headers, format_name='vertical' if expanded else None)
+            first_line = formatted[:formatted.find('\n')]
 
-            if (not expanded and max_width and rows and
-                    content_exceeds_width(rows[0], max_width) and headers):
+            if (not expanded and max_width and headers and rows and
+                    len(first_line) > max_width):
                 formatted = self.formatter.format_output(
                     rows, headers, format_name='vertical')
 
@@ -961,13 +962,6 @@ def cli(database, user, host, port, socket, password, dbname,
             click.secho(str(e), err=True, fg='red')
             exit(1)
 
-
-def content_exceeds_width(row, width):
-    # Account for 3 characters between each column
-    separator_space = (len(row)*3)
-    # Add 2 columns for a bit of buffer
-    line_len = sum([len(str(x)) for x in row]) + separator_space + 2
-    return line_len > width
 
 def need_completion_refresh(queries):
     """Determines if the completion needs a refresh by checking if the sql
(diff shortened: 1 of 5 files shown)


tsroten:
@meeuw I think this will clash with the generator pull request. What do you think?

mycli/main.py
```diff
@@ -793,9 +793,10 @@ def format_output(self, title, cur, headers, expanded=False,
             rows = list(cur)
             formatted = self.formatter.format_output(
                 rows, headers, format_name='vertical' if expanded else None)
+            first_line = formatted[:formatted.find('\n')]
```
j-bennet: I guess splitting on `\n` and finding a longest line would be a bad idea because of performance?

meeuw:
yes, this doesn't work with the generators, I've also noticed that the rows generator is unrolled to a list (I'll fix that in the output generator branch).

maybe we should move the auto vertical mode to the cli_helpers tabular_output?

tsroten:
@meeuw I don't think CLI Helpers should know about terminal sizes. Plus, we'd have to add postprocessors (instead of preprocessors) in CLI Helpers to make it work (since we have to compare the formatted output). Which also causes problems for it being a generator.

mycli/main.py
tsroten: @j-bennet Yeah, I think so. Plus, @meeuw is trying to get all this to work with generators, in which case we don't want to loop through the whole output.

meeuw:
hmm yes, I have to reconsider what to do, maybe refactor MyCli.output to include the auto vertical mode.

meeuw:
Oh, I think this can actually be done with generators:
First, get the first_line: ```first_line = next(formatted)```
Next, chain the first_line back: ```formatted = itertools.chain([first_list], formatted)```

Is this ready for merging?

tsroten:
@meeuw Let's check with @j-bennet, first. Irina, are you ok with just looking at the first line (re: your comment ^^)?

j-bennet:
@tsroten @meeuw I think so. This is what we're doing right now anyway, so at least we're not breaking things.

One other thing we can consider though, since `cli_helpers` have to loop through all the lines, they know how long the longest was. Could we perhaps return this value from `format_output`, along with `rows` and `headers`?

tsroten:
@j-bennet Good idea. CLI Helpers could return something like `rows, headers, metadata`, where `metadata` is a dict with various pertinent information (e.g. `rows`, `columns`, etc.)

That would have to be something for >= v1 of CLI Helpers.

tsroten:
Merging for now. We can try to address CLI Helpers returning length in a later pull request.

PR #474, opened 2017-07-02
Submit against the latest tag instead of master

meeuw:
## Description
Submit against the latest tag instead of master, this way the patches from bug fixes can be cleanly applied by packagers.

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -128,3 +128,5 @@ $ ./setup.py lint --fix
 ```
 
 Be sure to commit and push any PEP 8 fixes.
+
+For bugfixes: Submit against the latest tag instead of master!


tsroten:
@meeuw I'm not sure I understand, how can you submit a bugfix against a tag? I don't see the tags coming up in the Github pull request base branch menu.

meeuw:
Yes, maybe this needs to be reworded; what I mean is that the bugfix needs to be branched from the latest tag and then appended with a merge to latest master. Like:
```
git fetch
git checkout v1.9.0 bugfix/test
... changes ...
git add .
git commit -m 'bugfix test'
git merge
git push
```

But maybe this is just a bit too much git-voodoo, the advantage is that we have commits in our history that can be used by packages to apply bugfixes without merging new features which are also in our master branch.

tsroten:
@meeuw Ah, ok, I see what you mean now. My fear is that new contributors will submit a PR and then have to redo it when we tell them it's not against a tag. Do you think packagers will actually use this?

I'm also interested in what @amjith and @j-bennet think. I don't know very much about this.

meeuw:
These are examples of mycli commits which are used by Fedora as bugfixes (copied from our git repository):

https://src.fedoraproject.org/cgit/rpms/mycli.git/tree/02629c91.patch
https://src.fedoraproject.org/cgit/rpms/mycli.git/tree/71f503af.patch
https://src.fedoraproject.org/cgit/rpms/mycli.git/tree/mycli-1.7.0-mv-authors.patch?id=9cd87750cffbe5e0b10c2149119eda98255386b6
https://src.fedoraproject.org/cgit/rpms/mycli.git/tree/mycli-1.8.0-sqlparse.patch?id=9cd87750cffbe5e0b10c2149119eda98255386b6

As I think of it; this can also be solved by releasing new versions after each bugfix.

amjith:
Since we are more knowledgeable about git why don't we cherry pick the bug
fix commits as we see fit on to a tag?

This should reduce the burden for a contributor.

meeuw:
Hmm do you mean that we do the above for a contributor before merging a bugfix?

For example when a contributor submits a PR contributor/test:
```
git fetch
git checkout v1.9.0 bugfix/test
git merge contributor/test
... (git mergetool / git commit)
git merge master
git push
```

amjith:
> Hmm do you mean that we do the above for a contributor before merging a bugfix?

Not quite. I was thinking that if we have to do a bug fix release of mycli but the master is not in a releasable state then we can create a new branch from the previous tag, cherry-pick the commits that fix the specific bug and release from that branch. 

The part that I'm unclear about is, do package maintainers release bug fix versions that don't match our our release cycle? I would expect that the official release versions of mycli will match the version numbers provided by Fedora package. Is that not true?

meeuw:
That's not true indeed; for example when mycli was incompatible with the latest version of sqlparse the patches got cherry picked from master and applied to the last release. The same for the "is_group" incompatibily and my accidental removal of the sqlparse workaround.

As I think of it, I think we should release more frequent (that would solve it as well).

tsroten:
I'm in favor of releasing more often as a way to help package maintainers and users.

tsroten:
Speaking of this. It's time for another mycli release! 🎆

amjith:
I like the idea of releasing more often. If the release process is painful let's take the time to make it as much automated as possible. 

Here's a goal that we can aim: Release a new version every day (if there are changes). 😄

meeuw:
Fine, decided haha. I'm closing this PR.

Issue #128, opened 2015-08-10
Add option for lowercase keyword completion


dgoosens:
Hi,

Thanks for this great tool...

I personally like my SQL statements to be in lowercase...
Would it be possible to add this feature ?
I guess a boolean in the config file could enable/disable it...

amjith:
It's doable but adds more complexity that makes it hard to maintain. For example, I have to be careful not to change the case for table/column names, but only change the case for keywords, functions etc. 

Again it's doable but it'll get a bit cumbersome to maintain. I'd prefer to keep the code base small, if I can help it.

darikg:
I wonder if this could be addressed simultaneously with dbcli/pgcli#287 by moving all the hard-coded completions to external config files, which could then be re-ordered or re-capitalized at will. That's another step down the road to config hell though.

amjith:
That's an interesting idea. I'm still hesitant about adding more config options. If there are more requests for this feature, I'll definitely consider implementing it.

tsroten:
This was requested via https://github.com/dbcli/mycli/issues/473.

tsroten:
This is now addressed by https://github.com/dbcli/mycli/pull/476. It will be in the next release of mycli.

tsroten:
@dgoosens This is included as of version 1.11.0. If you installed via `pip`, you can upgrade by running `pip install --upgrade mycli`.

To use this, just add this to your `~/.myclirc`:
```conf
keyword_casing = lower
```

We also support `auto` which will suggest upper or lower case based on how you type.

Thanks for the idea!

dgoosens:
Brilliant !  
Thanks

PR #476, opened 2017-07-09
Completions in lowercase.

j-bennet:
## Description

Pgcli can do `upper`, `lower` and `auto` casing for completions; ported the same functionality to `mycli`.

Issue: https://github.com/dbcli/mycli/issues/128

## Checklist

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -141,6 +141,7 @@ class MyCli(object):
         self.prompt_format = prompt or prompt_cnf or c['main']['prompt'] or \
                              self.default_prompt
         self.prompt_continuation_format = c['main']['prompt_continuation']
+        keyword_casing = c['main'].get('keyword_casing', 'upper')
 
         self.query_history = []
 
@@ -148,7 +149,8 @@ class MyCli(object):
         self.smart_completion = c['main'].as_bool('smart_completion')
         self.completer = SQLCompleter(
             self.smart_completion,
-            supported_formats=self.formatter.supported_formats)
+            supported_formats=self.formatter.supported_formats,
+            keyword_casing=keyword_casing)
         self._completer_lock = threading.Lock()
 
         # Register custom special commands.
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -728,7 +728,8 @@ class MyCli(object):
         self.completion_refresher.refresh(
             self.sqlexecute, self._on_completions_refreshed,
             {'smart_completion': self.smart_completion,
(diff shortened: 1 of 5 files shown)


mikeerickson:
+1 please merge :)

tsroten:
@j-bennet This has been on my wishlist for awhile. Thanks for doing this!!

mycli/main.py
```diff
@@ -141,14 +141,16 @@ def __init__(self, sqlexecute=None, prompt=None,
         self.prompt_format = prompt or prompt_cnf or c['main']['prompt'] or \
                              self.default_prompt
         self.prompt_continuation_format = c['main']['prompt_continuation']
+        keyword_casing = c['main'].get('keyword_casing', 'upper')
```
amjith: Should this default to 'auto'?

mycli/myclirc
```diff
@@ -77,6 +77,9 @@ login_path_as_host = False
 # and using normal tabular format otherwise. (This applies to statements terminated by ; or \G.)
 auto_vertical_output = False
 
+# keyword casing preference. Possible values "lower", "upper", "auto"
+keyword_casing = upper
```
amjith: Same here. I think 'auto' might be preferable.

mycli/sqlcompleter.py
```diff
@@ -59,6 +59,9 @@ def __init__(self, smart_completion=True, supported_formats=()):
 
         self.special_commands = []
         self.table_formats = supported_formats
+        if keyword_casing not in ('upper', 'lower', 'auto'):
+            keyword_casing = 'upper'
```
amjith: 'auto'

mycli/myclirc
j-bennet: I was thinking more along the lines of preserving current mycli behavior in defaults, but it's true, auto sounds like a better default. I'll change that.

j-bennet:
Ideally, I'd rework the unit tests also, to parametrize the completer with lower case and upper case, but I don't have time for something that massive right now. It will have to be a TODO for later.

tsroten:
@j-bennet First of all, this is awesome!

If I set my keyword casing to "lower", it's getting changed to "auto". I think this happens in `mycli.refresh_completions()` where it has the `dict` of arguments -- we need to add `keyword_casing` there.

j-bennet:
@tsroten Good catch! That should be fixed now (losing the casing setting on refresh).

PR #481, opened 2017-07-24
Add fish-style autosuggestion.

amjith:
## Description
This adds fish-style autosuggestion. Use right arrow to choose a suggestion.

## Checklist

~- [ ] I've added my name to the `AUTHORS` file (or it's already there).~

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,12 @@
+TBD
+===
+
+Features:
+---------
+
+* Add fish-style auto-suggestion from history. (Thanks: [Amjith Ramanujam])
+
+
 1.11.0:
 =======
 
--- a/mycli/clitoolbar.py
+++ b/mycli/clitoolbar.py
@@ -2,7 +2,7 @@ from pygments.token import Token
 from prompt_toolkit.enums import DEFAULT_BUFFER, EditingMode
 from prompt_toolkit.key_binding.vi_state import InputMode
 
-def create_toolbar_tokens_func(get_is_refreshing):
+def create_toolbar_tokens_func(get_is_refreshing, show_fish_help):
     """
     Return a function that generates the toolbar tokens.
     """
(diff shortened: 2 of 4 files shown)


j-bennet:
Something is wrong with integration tests - `can't drop database mycli_behave_tests`. I think it should be `mycli_behave_tests_{python-version}`.

meeuw:
I agree it is a good idea to ignore pep8radius for this line. I really like the fish-style autosuggestion, but I think we should add some documentation about it. I expected "tab" or "enter" would complete the suggestion, after some experiments "cursor key" right accepted the completion...

tsroten:
@meeuw 👍 for the documentation. Also, `<Ctrl>-F` is another way to trigger the completion. Personally, I find it easier to type than right arrow, which is hard for me to reach.

amjith:
> I really like the fish-style autosuggestion, but I think we should add some documentation about it. 

I agree. The first time I was introduced to fish I was frustrated because I didn't know about the right arrow and I kept cursing for the first 5 mins before someone rescued me. 

I wonder if we can add a temporary status in the status bar to press right arrow. When they press the right arrow for the first time and we successfully complete the auto-suggestion we can stop showing the status.

amjith:
> `<Ctrl>-F` is another way to trigger the completion. 

You can also use `<Ctrl>-E` to complete the suggestion.

meeuw:
Hmm, maybe it's a good idea to show multiple status lines like nano with all keyboard shortcuts (`^L` `^R` `^U` `^C` `Tab` etc) ? Which will be removed (except for the shortcut `^L Clear Screen`) if the output buffer doesn't fit (anymore)?

amjith:
I have updated the PR to show help in the toolbar for the first two commands.

I've also removed a couple of unnecessary status indicators in from the toolbar. My reason being people don't ever use the smart completion toggle. It made sense when mycli was first released when the completion engine was nascent but it has grown mature now. I've also removed the status that shows it's in emacs mode. This is an unnecessary toggle since people never really toggle between the two modes. When they need to switch they can change the config for it in the rc file.

amjith:
This PR is now ready for review.

tsroten:
@amjith Looks great! Works as expected. I like the new status bar. Good job!

PR #484, opened 2017-08-02
keep cursor

meeuw:
## Description
keep cursor in cli_helpers

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,9 +6,10 @@ Bug Fixes:
 
 * Prevent missing MySQL help database from causing errors in completions (Thanks: [Thomas Roten]).
 
-Features:
----------
-* Keep cursor for cli_helpers. (Thanks: [Dick Marinus])
+Internal Changes:
+-----------------
+
+* Use less memory when formatting results for display (Thanks: [Dick Marinus]).
 
 1.12.0:
 =======
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,11 @@
+TBD
+===
+
+Features:
+---------
+* Keep cursor for cli_helpers. (Thanks: [Dick Marinus])
+
+
 1.12.0:
 =======
 
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -40,7 +40,7 @@ from .completion_refresher import CompletionRefresher
 from .config import (write_default_config, get_mylogin_cnf_path,
                      open_mylogin_cnf, read_config_files, str_to_bool)
 from .key_bindings import mycli_bindings
-from .encodingutils import utf8tounicode
+from .encodingutils import utf8tounicode, text_type
 from .lexer import MyCliLexer
 from .__init__ import __version__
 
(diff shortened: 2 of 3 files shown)


mycli/main.py
```diff
@@ -811,16 +811,29 @@ def format_output(self, title, cur, headers, expanded=False,
...
+                def sanitize(col):
+                    if col is converters.through:
+                        return text_type
+                    else:
+                        return col
+                column_types = [sanitize(converters.decoders[col[1]])
+                                for col in cur.description]
+
+            if max_width is not None:
+                cur = list(cur)
```
tsroten: I'm probably just missing something here -- why do we need to convert it to a list if auto vertical output is turned on?

changelog.md
```diff
@@ -1,3 +1,11 @@
+TBD
+===
+
+Features:
+---------
+* Keep cursor for cli_helpers. (Thanks: [Dick Marinus])
```
tsroten: Can we change this to a more user-friendly change message? Maybe something like:

Internal Changes:
* Use less memory when formatting results for display.

mycli/main.py
```diff
@@ -811,16 +811,29 @@ def format_output(self, title, cur, headers, expanded=False,
             output.append(title)
 
         if cur:
-            rows = list(cur)
+            column_types = None
+            if hasattr(cur, 'description'):
+                def sanitize(col):
+                    if col is converters.through:
```
tsroten: It looks like PyMySQL has other converters that we'd need to check for.

I think you can just do `return col if type(col) is type else text_type`

tsroten:
@meeuw This will be a good change 😄 . I put a few comments on here for you to check out.

mycli/main.py
meeuw: Auto vertical output has the potential to re-evaluate the cursor after the first try of horizontal output. This isn't possible without re-running the query which I find scary because a query might contain mutations.

changelog.md
meeuw: That's much better, thanks.

mycli/main.py
meeuw: Looks good, great idea!

mycli/main.py
tsroten: Ah ok, I see now! That makes sense.

mycli/main.py
```diff
@@ -811,16 +811,29 @@ def format_output(self, title, cur, headers, expanded=False,
...
+                column_types = [sanitize(converters.decoders[col[1]])
+                                for col in cur.description]
+
+            if max_width is not None:
+                cur = list(cur)
+
             formatted = self.formatter.format_output(
-                rows, headers, format_name='vertical' if expanded else None,
+                cur, headers, format_name='vertical' if expanded else None,
+                column_types=column_types,
```
tsroten: Should we set column_types in `output_kwargs` (i.e. `output_kwargs['column_types'] = ...`) so that this call to `format_output()` *and* the call below will both get the column types?

meeuw: You're right, I forgot this one. Fixed.

tsroten:
Thanks @meeuw! Well done 👍

Issue #485, opened 2017-08-03
Missing MySQL help command table causes background completion error


WinstonN:
Hi Hi

Using v 1.12.0 on Ubuntu 14.04 
on connecting to mysql I get error

```
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/usr/local/lib/python2.7/dist-packages/mycli/completion_refresher.py", line 61, in _bg_refresh
    refresher(completer, executor)
  File "/usr/local/lib/python2.7/dist-packages/mycli/completion_refresher.py", line 121, in refresh_show_commands
    completer.extend_show_items(executor.show_candidates())
  File "/usr/local/lib/python2.7/dist-packages/mycli/sqlcompleter.py", line 98, in extend_show_items
    for show_item in show_items:
  File "/usr/local/lib/python2.7/dist-packages/mycli/sqlexecute.py", line 197, in show_candidates
    cur.execute(self.show_candidates_query)
  File "/usr/local/lib/python2.7/dist-packages/pymysql/cursors.py", line 166, in execute
    result = self._query(query)
  File "/usr/local/lib/python2.7/dist-packages/pymysql/cursors.py", line 322, in _query
    conn.query(q)
  File "/usr/local/lib/python2.7/dist-packages/pymysql/connections.py", line 856, in query
    self._affected_rows = self._read_query_result(unbuffered=unbuffered)
  File "/usr/local/lib/python2.7/dist-packages/pymysql/connections.py", line 1057, in _read_query_result
    result.read()
  File "/usr/local/lib/python2.7/dist-packages/pymysql/connections.py", line 1340, in read
    first_packet = self.connection._read_packet()
  File "/usr/local/lib/python2.7/dist-packages/pymysql/connections.py", line 1014, in _read_packet
    packet.check_error()
  File "/usr/local/lib/python2.7/dist-packages/pymysql/connections.py", line 393, in check_error
    err.raise_mysql_exception(self._data)
  File "/usr/local/lib/python2.7/dist-packages/pymysql/err.py", line 107, in raise_mysql_exception
    raise errorclass(errno, errval)
ProgrammingError: (1146, u"Table 'mysql.help_topic' doesn't exist")
```

tsroten:
[...]

PR #487, opened 2017-08-03
Catch error for missing help_topic database.

tsroten:
## Description

The `mysql.help_topic` database doesn't always exist on MySQL servers. We query it for completion data.

This PR catches the PyMySQL error in the case that the database doesn't exit.

See https://github.com/dbcli/mycli/issues/485 for more information.

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,11 @@
+TBD:
+=======
+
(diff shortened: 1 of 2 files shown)


amjith:
🗡️

Issue #488, opened 2017-08-03
Terminal size with height < 18 rows makes mycli not start in Python 2


tsroten:
In Python 2, with a terminal sized so that there are less than 18 height rows, makes mycli not start:
```
  File "/Users/troten/.local/share/virtualenv/mycli2/lib/python2.7/site-packages/prompt_toolkit/layout/containers.py", line 1209, in _copy_body
    for y in range(ypos, ypos + write_position.height):
TypeError: integer argument expected, got float
```

This is because in Python 3, `round(n)` returns an integer (like Prompt Toolkit expects), where as in Python 2 `round(n)` returns a floating point decimal.

tsroten:
This is fixed in mycli 1.12.1, which is currently available via pip:
```bash
pip install -U mycli
```

PR #489, opened 2017-08-03
Make sure that terminal size is returned as an integer.

tsroten:
## Description

Under Python 2 `round(n)` returns a float. In Python 3 it returns an integer. We use `round` to determine the amount of reserved space we should use for the completion menu. Prompt Toolkit requires this to be an integer (because it passes it to `range()`).

See https://github.com/dbcli/mycli/issues/488

## Checklist


amjith:
Nice find. 🍫

Issue #253, opened 2016-04-11
Exception after dropping database being used


bryant1410:
The following causes a crash:

``` sql
CREATE DATABASE D;
USE D;
DROP DATABASE D;
```

tsroten:
Thanks for reporting this @bryant1410. I'm sorry you ran into a bug! Just so we have it, I've pasted the full stack trace below. It looks like the database is successfully dropped. The error comes from the background thread that is refreshing the completion list.

```
root@localhost:foo1> drop database foo1;
Query OK, 0 rows affected
root@localhost:foo1> Exception in thread completion_refresh:
Traceback (most recent call last):
  File "/usr/local/Cellar/python3/3.5.0/Frameworks/Python.framework/Versions/3.5/lib/python3.5/threading.py", line 923, in _bootstrap_inner
    self.run()
  File "/usr/local/Cellar/python3/3.5.0/Frameworks/Python.framework/Versions/3.5/lib/python3.5/threading.py", line 871, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/troten/code/mycli/mycli/completion_refresher.py", line 51, in _bg_refresh
    e.socket, e.charset, e.local_infile, e.ssl)
  File "/Users/troten/code/mycli/mycli/sqlexecute.py", line 41, in __init__g completions...
    self.connect()
  File "/Users/troten/code/mycli/mycli/sqlexecute.py", line 70, in connect
    ssl=ssl)
  File "/Users/troten/.virtualenvs/mycli/lib/python3.5/site-packages/pymysql/__init__.py", line 88, in Connect
    return Connection(*args, **kwargs)
  File "/Users/troten/.virtualenvs/mycli/lib/python3.5/site-packages/pymysql/connections.py", line 679, in __init__
    self.connect()
  File "/Users/troten/.virtualenvs/mycli/lib/python3.5/site-packages/pymysql/connections.py", line 891, in connect
    self._request_authentication()
  File "/Users/troten/.virtualenvs/mycli/lib/python3.5/site-packages/pymysql/connections.py", line 1097, in _request_authentication
    auth_packet = self._read_packet()
  File "/Users/troten/.virtualenvs/mycli/lib/python3.5/site-packages/pymysql/connections.py", line 966, in _read_packet
    packet.check_error()
  File "/Users/troten/.virtualenvs/mycli/lib/python3.5/site-packages/pymysql/connections.py", line 394, in check_error
    err.raise_mysql_exception(self._data)
[...]

PR #490, opened 2017-08-03
Gracefully handle dropping default database.

tsroten:
## Description

In mycli, the current behavior when dropping the current default database is to:
- display a stack track for the background completion refresher's error
- keep the now-dropped database as the default database for the current connection

This PR:
- resets the current database in the connection when it is dropped

This means that the prompt is updated to display `(none)` as the current database once you drop it (i.e. same behavior as mysql).

See https://github.com/dbcli/mycli/issues/253 for more information.

## Checklist

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -1014,17 +1014,16 @@ def need_completion_refresh(queries):
 
 def is_dropping_database(queries, dbname):
     """Determine if the query is dropping a specific database."""
-    for query in sqlparse.split(queries):
-        try:
-            stmt = sqlparse.parse(query)[0]
-            first_token = stmt.token_first(skip_cm=True)
-            _, second_token = stmt.token_next(0, skip_cm=True)
-            if (first_token.value.lower() == 'drop' and
-                    second_token.value.lower() in ('database', 'schema') and
-                    stmt.get_name().lower() == dbname.lower()):
-                return True
-        except Exception:
-            return False
+    if dbname is None:
+        return False
+
+    for query in sqlparse.parse(queries):
+        first_token = query.token_first(skip_cm=True)
+        _, second_token = query.token_next(0, skip_cm=True)
+        if (first_token.value.lower() == 'drop' and
+                second_token.value.lower() in ('database', 'schema') and
+                query.get_name().lower() == dbname.lower()):
+            return True
 
 
 def need_completion_reset(queries):
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -1017,8 +1017,8 @@ def is_dropping_database(queries, dbname):
     for query in sqlparse.split(queries):
         try:
             stmt = sqlparse.parse(query)[0]
-            first_token = stmt.token_first(skip_ws=True, skip_cm=True)
-            _, second_token = stmt.token_next(0, skip_ws=True, skip_cm=True)
+            first_token = stmt.token_first(skip_cm=True)
+            _, second_token = stmt.token_next(0, skip_cm=True)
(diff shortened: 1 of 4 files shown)


mycli/main.py
```diff
@@ -1007,6 +1011,22 @@ def need_completion_refresh(queries):
...
+    for query in sqlparse.split(queries):
+        try:
+            stmt = sqlparse.parse(query)[0]
+            first_token = stmt.token_first(skip_cm=True)
+            _, second_token = stmt.token_next(0, skip_cm=True)
+            if (first_token.value.lower() == 'drop' and
+                    second_token.value.lower() in ('database', 'schema') and
+                    stmt.get_name().lower() == dbname.lower()):
+                return True
+        except Exception:
```
meeuw: Hmm this is awful for debugging; can we use a more specific Exception or at least use:
```
            self.logger.debug('In is_dropp_database: %r.', e)                                                                          
            self.logger.error("traceback: %r", traceback.format_exc())                                                                       
```

BTW, I was thinking about adding a debug flag to disable the "catch all" exceptions in mycli (to help with debugging).

test/features/steps/crud_database.py
```diff
@@ -85,6 +93,16 @@ def step_see_db_dropped(context):
     wrappers.expect_exact(context, 'Query OK, 0 rows affected', timeout=2)
 
 
+@then('we see database dropped and no default database')
+def step_see_db_dropped_no_default(context):
+    """Wait to see drop database output."""
+    user = context.conf['user']
+    host = context.conf['host']
+    context.currentdb = '(none)'
```
meeuw: Maybe `context.currentdb = None` might be better (I'm not sure). dbname=None seem to be used in other places in the behave tests.

meeuw:
This is indeed useful and the same behavior of mysql, I've added some minor comments could you please have a look?

mycli/main.py
tsroten: Hmm, good point. I'm surprised I put the `Exception` in there, 😆 .

I think the scenario that needs to be handled is `sqlparse.parse(query)[0]` getting an `IndexError`. I'll do some testing and fix this.

test/features/steps/crud_database.py
tsroten: Ah, my problem was I was relying on `after_scenario` to test the prompt (thus needing to set the current db to `(none)` since that's what the prompt displays.

Now that I noticed you can do `context.atprompt = True`, this should be more appropriate.

mycli/main.py
tsroten: In my testing, the only error I could get was when `dbname` was `None`. I've handled that error and removed the `try/except`.

tsroten: I don't know if we need to disable the catch all Exceptions for debugging. What if we just try to figure out what the appropriate exception is and catch that instead of catching all exceptions? I've noticed a few of those around mycli.

meeuw:
I think this is ready to merge, I don't know whether it's an easy fix but when I run (notice the back ticks):
```
drop `_test_db`
```
mycli fails with `pymysql.err.InternalError: (1049, "Unknown database '_test_db'")`. I think you can drop a database from a stored procedure as well (which makes it impossible to detect by parsing the sql ourselves).

If it isn't an easy fix I think we should merge this PR first.

tsroten:
@meeuw Great point on the backticks. I've addressed that (and double-quotes as well).

I agree with you that the stored procedure situation can be left out for now.

meeuw:
this is a nice fix, thanks!

Issue #500, opened 2017-08-16
Feature request: source command filename completion


chrisamow:
Allowing completion on the source command would really add to the integrated feeling of mycli as a development tool.  It would allow very low friction execution of convenient batches of sql functionality like snippets or environmental database setup. This a followup to a comment left in #108

tsroten:
@chrisamow Thanks for the suggestion! It's a great idea 👍 !

Not sure when we can get to this, but anyone is welcome to give it a shot.

tsroten:
@chrisamow Good news! This was just released in version 1.13.0. If you installed via pip you can upgrade by running `pip install -U mycli`

It's also in Homebrew as well: `brew update && brew upgrade mycli`.

chrisamow:
You guys are awesome!!  Thanks! I use mycli a lot, especially when I want to copy/show the results.

PR #501, opened 2017-08-18
Suggest file names for source command.

j-bennet:
## Description

This would add file name completion to `source` command.

File name completion is not "smart", just a simple prefix match.

Related issue: https://github.com/dbcli/mycli/issues/500.

## Checklist

### DIFF
--- a/mycli/packages/completion_engine.py
+++ b/mycli/packages/completion_engine.py
@@ -1,4 +1,5 @@
 from __future__ import print_function
+import os
 import sys
 import sqlparse
 from sqlparse.sql import Comparison, Identifier, Where
@@ -83,7 +84,7 @@ def suggest_type(full_text, text_before_cursor):
         # Be careful here because trivial whitespace is parsed as a statement,
         # but the statement won't have a first token
         tok1 = statement.token_first()
-        if tok1 and tok1.value == '\\':
+        if tok1 and tok1.value in ['\\', 'source']:
             return suggest_special(text_before_cursor)
 
     last_token = statement and statement.token_prev(len(statement.tokens))[1] or ''
@@ -115,8 +116,11 @@ def suggest_special(text):
             {'type': 'view', 'schema': []},
             {'type': 'schema'},
         ]
+    elif cmd in ['\\.', 'source']:
+        return[{'type': 'file_name'}]
 
     return [{'type': 'keyword'}, {'type': 'special'}]
+
 
 def suggest_based_on_last_token(token, text_before_cursor, full_text, identifier):
     if isinstance(token, string_types):
@@ -289,8 +293,6 @@ def suggest_based_on_last_token(token, text_before_cursor, full_text, identifier
                 prev_keyword, text_before_cursor, full_text, identifier)
         else:
             return []
-    elif token_v == 'source':
-        return[{'type': 'file_name'}]
     else:
         return [{'type': 'keyword'}]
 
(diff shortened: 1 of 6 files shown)


tsroten:
I'm excited to try this out!!! I'm traveling until Tuesday, so I probably won't be able to review it until then. But, I'm sure someone else will give it a shot :-)

j-bennet:
@tsroten @meeuw I would appreciate it if you have time to poke at it (and maybe discover some bugs).

mycli/packages/completion_engine.py
```diff
@@ -289,6 +289,8 @@ def suggest_based_on_last_token(token, text_before_cursor, full_text, identifier
                 prev_keyword, text_before_cursor, full_text, identifier)
         else:
             return []
+    elif text_before_cursor.lower().startswith('source'):
```
meeuw: I think we also should support the `\.` command

meeuw:
The completion doesn't work when combined with another query. For example:
```
mysql root@172.17.0.2:dmtest> truncate table test ; source /tmp/test.sql
```
doesn't give completions

meeuw:
This is superb, I hit tab so many times with the \. command :smile:

j-bennet:
@meeuw Weird, it actually works for me combined with another query. I added a test [here](https://github.com/dbcli/mycli/blob/608a30721a377be8470178205d72aaafdb2520d4/test/test_completion_engine.py#L500-L508) to make sure.

meeuw:
Hmm now it works; maybe I wasn't testing on the right branch, sorry.

Could you please also add the `\.` alias? I think the `source` command should be added to `suggest_special` instead of `suggest_based_on_last_token` (not sure, haven't tried that).

j-bennet:
@meeuw Yes, I'm working on `./` alias. Right now, `mycli` won't even execute it:

```
mysql root@localhost:(none)> ./zzz.sql
(1064, u"You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near './zzz.sql' at line 1")
```

so it requires more than just the autocompletion part.

tsroten:
@j-bennet Thanks for working on this!! I think it's `\.` followed by a space.

j-bennet:
@tsroten Ah ok. Thanks for the tip! It's the same with the space though.

```
mysql root@localhost:(none)> . zzz.sql
(1064, u"You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '. zzz.sql' at line 1")
mysql root@localhost:(none)> . ./zzz.sql
(1064, u"You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '. ./zzz.sql' at line 1")
```

tsroten:
@j-bennet Haha, sorry I wasn't clear, I mean like this:
```
localhost:em> \. /tmp/test.sql

```

j-bennet:
@tsroten Oh, I thought it was just a dot. That's what this says http://tldp.org/HOWTO/Bash-Prompt-HOWTO/x237.html. Is it incorrect?

j-bennet:
...but in `mycli` it's a special command `\.`, I see what you mean. Ok.

meeuw:
@j-bennet exactly, `source` / `.` are bash commands and I think these inspired the `mysql` authors (the official mysql client has `source` and `\.` as well.)

j-bennet:
File name is now completed for `\.` command also. The handling of both `\.` and `source` is moved to `suggest_special`.

meeuw:
Thanks a lot @j-bennet this is very useful! I'm going this use this a lot.

PR #503, opened 2017-08-24
Fix mysql4 version comment retrieval

### LINKED ISSUE #502: mycli remote connect fail  ，system is macOS，local is OK
log：
image

image

remote mysql version：

image

can  you help me？

fpietka:
## Description
The query fetching version comment was failing on MySQL 4.
Fix #502 

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,6 +5,7 @@ Bug Fixes:
 ----------
 
 * Fix UnicodeEncodeError when editing sql command in external editor
+* Fix MySQL4 version comment retrieval (Thanks: [François Pietka])
 
 1.12.1:
 =======
@@ -499,3 +500,4 @@ Bug Fixes:
 [Casper Langemeijer]: https://github.com/langemeijer
 [Scrappy Soft]: https://github.com/scrappysoft
 [Dick Marinus]: https://github.com/meeuw
+[François Pietka]: https://github.com/fpietka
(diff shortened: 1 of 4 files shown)


fpietka:
FYI, I used tommi2day/mysql4 docker image to test it.

meeuw:
@fpietka thanks for working on this; as a reminder, I get an error with python 2.7/mysql 4, could you please have a look? Maybe use `.encode('utf8')` instead of `str(..., 'utf8')`.

fpietka:
@meeuw thanks for the reminder. I tested it on python2/3 with mysql4 only though.

meeuw:
:+1: works! Thanks a lot. Only the `\s` (or `status`) command seems to return an error, Do you think this can/should be solved in this PR?

fpietka:
I am not sure how, there is no `GLOBAL` in mysql4. I could try to use `SHOW STATUS` and `SHOW VARIABLES` as fallback, but I'm not sure if it is the same.

fpietka:
I'm trying to get something, but it seems we're missing some keys in status.

fpietka:
Ok, I came with a workaround. Let me know what do you think about it.

j-bennet:
@fpietka It looks like your tests are failing because of pep8 checks. Run `pep8radius master -i --docformatter` to fix that.

fpietka:
I switched to pyenv to make my virtualenv, but it doesn't use system packages (hence I wasn't linting anymore on vim...)

meeuw:
Works like a charm! Thanks @fpietka!

Issue #509, opened 2017-08-30
Error When Outputting Certain Field Types


super13:
select param from  p_task_status; 
will get a red 245. 
is it not support json type? is there any log or debug mode?

table:
CREATE TABLE `p_task_status` (
  `id` bigint(20) NOT NULL AUTO_INCREMENT,
  `task_id` varchar(255) DEFAULT NULL,
  `host` varchar(255) DEFAULT NULL,
  `status` int(10) DEFAULT NULL,
  `status_msg` varchar(255) DEFAULT NULL,
  `start_time` timestamp NULL DEFAULT CURRENT_TIMESTAMP ,
  `end_time` timestamp DEFAULT '0000-00-00 00:00:00',
  `time_spends` bigint(20) DEFAULT NULL,
  `param` json DEFAULT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB AUTO_INCREMENT=1 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;

tsroten:
@super13 Hello! I'm sorry that you're having trouble with your JSON values. Mycli does support JSON data, e.g.:
```
localhost:test_db> select param from p_task_status
+--------------------------------------+
| param                                |
+--------------------------------------+
| {"key1": "value1", "key2": "value2"} |
+--------------------------------------+
1 row in set
Time: 0.016s
```

To find out more information about your error, you can check the log file. By default it is `~/.mycli.log`. Let us know what you find and we'll be happy to help you debug it.

tsroten:
Hi @super13! @fpietka has replicated the issue. It's a bug with mycli. We'll get it fixed as soon as we can. Thanks for reporting this!

LiraNuna:
For posterity, this is in the logs:

```
2017-09-12 22:22:20,606 (20851/MainThread) mycli.main ERROR - sql: u'SELECT media.metadata FROM media LIMIT 1', error: KeyError(245,)
2017-09-12 22:22:20,606 (20851/MainThread) mycli.main ERROR - traceback: 'Traceback (most recent call last):\n  File "/usr/local/lib/python2.7/dist-packages/mycli/main.py", line 541, in one_iteration\n    max_width)\n  File "/usr/local/lib/python2.7/dist-packages/mycli/main.py", line 850, in format_output\n    for col in cur.description]\nKeyError: 245\n'
```

tsroten:
[...]

PR #516, opened 2017-09-21
Fix error when displaying NULL and JSON values.

tsroten:
## Description

This makes two changes:
- Adds a `FIELD_TYPES` dict in `sqlexecute` so we can add column types as needed.
- Uses `.get()` so that column types will fall back to a string if they don't exist in the field types dict.

This addresses https://github.com/dbcli/mycli/issues/509

## Checklist

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -844,10 +844,10 @@ class MyCli(object):
         if cur:
             column_types = None
             if hasattr(cur, 'description'):
-                def sanitize(col):
-                    return col if type(col) is type else text_type
-                column_types = [sanitize(FIELD_TYPES.get(col[1], text_type))
-                                for col in cur.description]
+                def get_col_type(col):
+                    col_type = FIELD_TYPES.get(col[1], text_type)
+                    return col_type if type(col_type) is type else text_type
+                column_types = [get_col_type(col) for col in cur.description]
 
             if max_width is not None:
                 cur = list(cur)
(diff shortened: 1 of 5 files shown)


mycli/main.py
```diff
@@ -846,7 +846,7 @@ def format_output(self, title, cur, headers, expanded=False,
             if hasattr(cur, 'description'):
                 def sanitize(col):
                     return col if type(col) is type else text_type
-                column_types = [sanitize(converters.decoders[col[1]])
+                column_types = [sanitize(FIELD_TYPES.get(col[1]))
```
j-bennet: Should it be `FIELD_TYPES.get(col[1], str)` here?

tsroten: That would be more explicit which might be better. Right now line 848 will set it to `text_type`.

mycli/main.py
```diff
@@ -846,7 +846,7 @@ def format_output(self, title, cur, headers, expanded=False,
             if hasattr(cur, 'description'):
```
meeuw: I think we should move the `FIELD_TYPES.get(col[1], text_type)` to the "sanitize" function.

meeuw:
Thanks Thomas, this looks like a nice and clean fix.

mycli/main.py
tsroten: @meeuw Good thinking :) - done.

Issue #522, opened 2017-09-24
Tests should not read user config files


tsroten:
Currently our tests read a few user config files:
- `.mylogin.cnf`
- `.myclirc`
- `.my.cnf`

We need to update the unit tests and integrated tests to never read these from the user. We can use the following methods to override this behavior:
- environment variable `MYSQL_TEST_LOGIN_FILE `
- command-line argument `--myclirc`
- command-line argument `--defaults-file`

tsroten:
Fixed in version 1.13.1.

PR #523, opened 2017-09-24
Don't use user config files in tests

tsroten:
## Description

This makes the tests ignore the following user config files:
- login path file
- my.cnf
- myclirc

See https://github.com/dbcli/mycli/issues/522

## Checklist

### DIFF
--- a/test/test_main.py
+++ b/test/test_main.py
@@ -20,8 +20,13 @@ except NameError:
 default_config_file = os.path.abspath(os.path.join(
     os.path.join(os.path.dirname(os.path.dirname(__file__)), 'mycli'),
     'myclirc'))
+login_path_file = os.path.abspath(os.path.join(os.path.dirname(__file__),
+                                               'mylogin.cnf'))
+
+os.environ['MYSQL_TEST_LOGIN_FILE'] = login_path_file
 CLI_ARGS = ['--user', USER, '--host', HOST, '--port', PORT,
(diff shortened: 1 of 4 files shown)


test/test_main.py
```diff
@@ -17,8 +17,16 @@
...
+test_dir = os.path.abspath(os.path.dirname(__file__))
+project_dir = os.path.dirname(test_dir)
+default_config_file = os.path.join(project_dir, 'mycli', 'myclirc')
+login_path_file = os.path.join(test_dir, 'mylogin.cnf')
+
+os.environ['MYSQL_TEST_LOGIN_FILE'] = login_path_file
 CLI_ARGS = ['--user', USER, '--host', HOST, '--port', PORT,
-            '--password', PASSWORD, '_test_db']
+            '--password', PASSWORD, '--myclirc', default_config_file,
+            '--defaults-file', default_config_file,
```
tsroten: I have the unit tests using the default `myclirc` as the mysql `my.cnf` because we need some kind of cnf-type file to override the user's file, but we don't actually need to set any config in it. Since `myclirc` is the same format as `my.cnf` and because it doesn't use the `[mysql]` group, I figured we could use it rather than adding another file to the repo.

amjith:
Looks good. Clever idea to use myclirc as a stand in for my.cnf. :clinking_glasses:

Issue #508, opened 2017-08-28
Mycli Crashes When Unable to Read Login Path File


Peter-Sh:
I've just installed mycli in debian 9 using official repository and the tool fails to start

```Traceback (most recent call last):
  File "/usr/bin/mycli", line 11, in <module>
    load_entry_point('mycli==1.8.1', 'console_scripts', 'console')()
  File "/usr/lib/python3/dist-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3/dist-packages/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python3/dist-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "/usr/share/mycli/mycli/main.py", line 769, in cli
    auto_vertical_output=auto_vertical_output, warn=warn)
  File "/usr/share/mycli/mycli/main.py", line 158, in __init__
    mylogin_cnf = open_mylogin_cnf(mylogin_cnf_path)
  File "/usr/share/mycli/mycli/config.py", line 103, in open_mylogin_cnf
    plaintext = read_and_decrypt_mylogin_cnf(f)
  File "/usr/share/mycli/mycli/config.py", line 172, in read_and_decrypt_mylogin_cnf
    pplain = aes_cipher.decrypt(cipher)
  File "/usr/lib/python3/dist-packages/Crypto/Cipher/blockalgo.py", line 295, in decrypt
    return self._cipher.decrypt(ciphertext)
ValueError: Input strings must be a multiple of 16 in length
```
It does not start even if I supply username and password in the args.

It seems this is because mycli expects credentials in mylogin.cnf file. But the problem is that MariaDb (which is official mysql server and client for RedHat, CentOS and Debian) does not have mysql_config_editor. (And I think will [never](https://mariadb.com/resources/blog/mysql-56-security-through-complacency) has) So starting mycli tool becomes hard if ever possible.

I think mycli may read or at least fall back to plain ~/.my.cnf file just like mysql standard client does to avoid confusion with MariaDb users. Or maybe just don't try to read encrypted file if credentials are provided in command line.

Thank you!

tsroten:
[...]

PR #524, opened 2017-09-24
Catch ValueError when reading login path file.

tsroten:
## Description

We already catch quite a few errors while reading the login path file. However, if the part of the login path file that says how many bytes to read was corrupted, then mycli would crash with a `ValueError`. This fixes that.

This also removes old pycryptodome test-related code that is no long applicable.

See https://github.com/dbcli/mycli/issues/508

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,11 @@
+TBD:
+=======
+
+Bug Fixes:
+----------
+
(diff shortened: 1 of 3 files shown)


amjith:
:beetle: :hocho:

Issue #69, opened 2015-06-26
feature request: repeat command on interval


grooverdan:
While monitoring a server sometime you do commands like show slave status\G or show processlist, or show global status like ... ; where you want it repeated every few seconds or so.

is a syntax like \w{atch} [interval default 5s] :  {SQL}

possible?

scottchiefbaker:
I agree this would be a great feature. A simple `while()` loop with a `sleep()` inside would be sufficient.

amjith:
This feature is now available in master. You can install directly from master as follows: 

```
pip install -U https://github.com/dbcli/mycli/archive/master.zip
```

This will be available in the next release.

grooverdan:
Thankyou!

ShawnCorrigan:
I know this is late, but is there documentation for this? I can't seem to find any (although I could be blind)

PR #526, opened 2017-09-30
Add `watch` command - repeat a query every N seconds

Terseus:
## Description

Add a special `watch` command with the syntax `watch [seconds] query`, without shorthand.

The command executes a query every `[seconds]` seconds, 5 by default.

While it works, it have some bugs:
* The waiting time is added to the query execution time shown to the user.
* After the first execution of the query, the output have an extra newline at the beginning.

Also it probably have lots of edge cases that are not covered, like running a query that requires user intervention (e.g. a reconnect prompt).

The implementation it's... not really good 😅  but it's simple and fills the gap of a feature asked in 
#69, so maybe it's worth it for a merge. Sorry, but writing a better, more complete version would require a lot more of work.

## Checklist

- [x] Add unit tests.
- [x] Accept float as interval.
- [x] Prevent the pager to force a pause between iterations.
- [x] Add the warning on destructive commands before the first iteration.
- [x] Add flag to clear the screen between query executions.
- [x] ~Bugfix: make it work for multi-query statements.~ Too complex, maybe will be done in a separate PR.

## Side notes

Sorry, I know the branch build is failing but honestly, I don't know why.

The only clue I found is this message at the end of the builds: `fatal: Not a valid object name master`.

Maybe some change in Travis broke the tests, or did I f****d it up at some point that I don't see?

**Edit**: Ok found the cause, I described in (maybe too much) detail in the issue #527.

### DIFF
--- a/mycli/packages/special/iocommands.py
+++ b/mycli/packages/special/iocommands.py
@@ -4,6 +4,7 @@ import locale
 import logging
 import subprocess
 from io import open
+from time import sleep
 
 import click
 import sqlparse
@@ -331,3 +332,45 @@ def unset_once_if_written():
     global once_file
     if written_to_once_file:
         once_file = None
+
+
+@special_command(
+    'watch',
+    'watch [seconds] query',
+    'Executes the query every [seconds] seconds (by default 5).'
+)
+def watch_query(arg, **kwargs):
+    usage = "Syntax: watch [seconds] query.\n"
+    if not arg:
+        yield (None, None, None, usage)
+        raise StopIteration
+    try:
+        args = arg.split(' ')
+        seconds = int(args[0])
+        statement = " ".join(filter(None, args[1:])).strip()
+    except ValueError:
+        seconds = 5
+        statement = arg
+    if not statement:
+        yield (None, None, None, usage)
+        raise StopIteration
+    cur = kwargs['cur']
+    sql_list = [
+        (sql.rstrip(';'), '> %s' % (sql))
+        for sql in sqlparse.split(statement)
+    ]
+    while True:
+        try:
+            for (sql, title) in sql_list:
+                cur.execute(sql)
+                if cur.description:
+                    headers = [x[0] for x in cur.description]
+                    yield (title, cur, headers, None)
+                else:
+                    yield (title, None, None, None)
+            sleep(seconds)
+        except KeyboardInterrupt:
+            # This prints the Ctrl-C character in its own line, which prevents
+            # to print a line with the cursor positioned behind the prompt
+            click.secho("", nl=True)
+            raise StopIteration
--- a/mycli/packages/special/iocommands.py
+++ b/mycli/packages/special/iocommands.py
@@ -356,11 +356,15 @@ def watch_query(arg, **kwargs):
         raise StopIteration
     cur = kwargs['cur']
     sql_list = [
-        (sql.rstrip(';'), '> %s' % (sql))
+        (sql.rstrip(';'), "> {0!s}".format(sql))
         for sql in sqlparse.split(statement)
     ]
+    old_pager_enabled = is_pager_enabled()
     while True:
         try:
+            # Somewhere in the code the pager its activated after every yield,
+            # so we disable it in every iteration
+            set_pager_enabled(False)
             for (sql, title) in sql_list:
                 cur.execute(sql)
                 if cur.description:
@@ -374,3 +378,5 @@ def watch_query(arg, **kwargs):
             # to print a line with the cursor positioned behind the prompt
             click.secho("", nl=True)
             raise StopIteration
+        finally:
+            set_pager_enabled(old_pager_enabled)
--- a/mycli/packages/special/iocommands.py
+++ b/mycli/packages/special/iocommands.py
@@ -13,6 +13,7 @@ from . import export
 from .main import special_command, NO_QUERY, PARSED_QUERY
 from .favoritequeries import favoritequeries
 from .utils import handle_cd_command
+from ..prompt_utils import confirm_destructive_query
 
 TIMING_ENABLED = False
 use_expanded_output = False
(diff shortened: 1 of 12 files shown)


meeuw:
Pretty useful! Thanks a lot.

I usually use `watch 'mysql -Be 'select 123'` to do this.

I think this is ready to merge as-is, Travis doesn't give any errors now, it would be nice to add some unit tests for watch_query (see `test/test_special_iocommands.py`).

It would be nice (but in my opinion not required) to be able to use a float as interval (0.3 for example), clear screen, show differences, message about disabling the pager/prompting and ability to run multiple queries (`watch 1 select 1 ; select 2`).

Terseus:
> it would be nice to add some unit tests for watch_query

I didn't do it because I'm not sure how to implement them them (how can I send a Ctrl-C while in the middle of a `sleep`?) but I'll take a look.

> use a float as interval

Not a problem, that's a very small change.

> clear screen

You know, in the first version the screen was cleared but I removed it because I though it's nice being able to see the previous executions of the query, not only the last.
Maybe implement it as a flag? The only problem I see is the help, which can be too long for the `?` command (in fact I rewrote the command description several times to make it less disruptive).

> show differences

I think that's better to implement it in another PR, I prefer to make two small changes than one big.
Also I think it would be nice to have the diff as a more generalist feature so it can be reused in other places (e.g. favorite queries, or maybe its own `diff` special command somehow).

> message about disabling the pager/prompting

Sorry, I don't understand what do you mean exactly, can you explain it better?

> ability to run multiple queries

Huh, I though that would work but know that I test it it doesn't, I'll take a look.

meeuw:
Thanks a lot, this is getting better and better :smile: 

> > message about disabling the pager/prompting

> Sorry, I don't understand what do you mean exactly, can you explain it better?

It would be nice if there is a message/warning about an enabled pager/destructive prompt, it's quite annoying if you get in a loop with a pager/prompt for each iteration.

tsroten:
Thanks @Terseus for working on this! I'm taking a look now as well. Very excited for this feature 😄

mycli/packages/special/iocommands.py
```diff
@@ -331,3 +332,45 @@ def unset_once_if_written():
...
+        statement = " ".join(filter(None, args[1:])).strip()
+    except ValueError:
+        seconds = 5
+        statement = arg
+    if not statement:
+        yield (None, None, None, usage)
+        raise StopIteration
+    cur = kwargs['cur']
+    sql_list = [
+        (sql.rstrip(';'), '> %s' % (sql))
```
tsroten: Can we use the new-style string formatting (i.e. `str.format()`) instead of the old-style (i.e. `%`-formatting) here?

tsroten:
Nice job on the unit tests 👍

mycli/packages/special/iocommands.py
Terseus: Oops, that slipped in from a copy&paste 😅 of course, I'll fix it in the next commit.

Terseus:
> It would be nice if there is a message/warning about an enabled pager/destructive prompt, it's quite annoying if you get in a loop with a pager/prompt for each iteration.

Oh ok, now I got it :)

Instead of showing a warning about the pager enabled, I just disabled it in the last commit, after all it makes no sense during the `watch` command to stop every iteration because of the pager.
Of course the pager is restored when the `watch` command finish.

About the warning about a prompt... that's much harder without making a big copy&paste, unfortunately, due to the `watch` comment being implemented as a special command (which I still think was the right choice).
E.g. the destructive warning prompt is controlled in the [`mycli/main.py`](https://github.com/dbcli/mycli/blob/25291958ccec7cbd930e684dbcd0b89bdad63f72/mycli/main.py#L493) which is outside of the control of a special command.

However now I tested it, a `watch 1 DELETE FROM table;` does **not** show a warning, which is a big problem; I'll add it to the TODO list.

Terseus:
@tsroten The latest commit was way bigger than I like as it includes a (small) refactor, please review it when you can.

tsroten:
@Terseus Great job on moving some of those things around so they can be used from the io commands module.

mycli/packages/special/iocommands.py
```diff
@@ -12,6 +13,7 @@
 from .main import special_command, NO_QUERY, PARSED_QUERY
 from .favoritequeries import favoritequeries
 from .utils import handle_cd_command
+from ..prompt_utils import confirm_destructive_query
```
tsroten: I know some of the existing imports are relative, but especially with the double-dot import, can we do an absolute import?

Terseus: Of course, I did it only to follow the existing convention, I'll fix it in the next commit.

Terseus:
Ok, I've implemented the flag to clear the screen; I think that it pollutes the help a little bit, but there will be time to remove it if we don't like it.

About the multi-statement bugfix, sorry but that's quite complex.
Currently the special commands gets a single statement by design so to fix the bug a refactor is need, and I think it can turns out to be big, specially because maybe the other special commands are not prepared to support multi-statements in their argument.
I'll open an issue for it this week, probably tomorrow.

Well, this was waaayy bigger than I though, it's good to see that all of you like it :)

Oh BTW, it's the first time I use the `mock` module, please take a look if the tests that use it can be improved.

meeuw:
Wow, I think you did an amazing job in cleaning up `main.py`. I don't have a lot of experience with mock either; it looks good to me. @tsroten as this includes a refactor I'd like to hear your opinion, I'd like to merge this asap 😄

amjith:
I'll review this PR tonight. Looks like a nice substantive PR.

amjith:
Very nice work @Terseus. I love the fact that you cleaned up the files that you touched.

Thank you for taking the time to implement a useful feature in a thorough manner. 

🌮

Terseus:
Glad you like it @amjith ^_^

I think we can close #69 now.

BTW the Python 2.7 build looks broken, found this error in the Travis log from the `step_query_select_123456` test: `      () argument 1 must be unicode, not str`

PR #533, opened 2017-10-15
Make the tests more compatible with other environments

Terseus:
## Description

Some changes to make the tests work in different build environments.

Changes included:
* Tests doesn't expect the "chatty" prompt, so it's removed from the
  default config file.
* Omit the database engine name in the prompt to make them compatible
  with MariaDB.
* Force locale to en_US.utf8, as expected by the tests.
* Add a note in CONTRIBUTING.md with information about the `ex` command.

Also includes pep8 fixes in test/features/steps/wrappers.py

## Checklist

## Notes

I made those changes after some frustrating tries to make all the tests pass in my machine (Archlinux).

Some of them are opinianted (like the removing of the database engine from the prompt) so it's open to discussion.

### DIFF
--- a/test/myclirc
+++ b/test/myclirc
@@ -1,0 +1,12 @@
+# vi: ft=dosini
+
+# This file is loaded after mycli/myclirc and should override only those
+# variables needed for testing.
+# To see what every variable does see mycli/myclirc
+
+[main]
+
+log_file = ~/.mycli.test.log
+log_level = DEBUG
+prompt = '\u@\h:\d> '
+less_chatty = True
--- a/test/myclirc
+++ b/test/myclirc
@@ -8,5 +8,5 @@
 
 log_file = ~/.mycli.test.log
 log_level = DEBUG
-prompt = '\u@\h:\d> '
+prompt = '\t \u@\h:\d> '
 less_chatty = True
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -111,6 +111,19 @@ The default values are `localhost`, `root`, no password, `3306`, and `utf8`.
 You only need to set the values that differ from the defaults.
 
 
+### CLI Tests
+
+Some CLI tests expect the program `ex` to be a symbolic link to `vim`.
+
+In some systems (e.g. Archlinux) `ex` is a symbolic link to `vi`, which will
+change the output and therefore make some tests fail.
+
+You can check this by running:
+```bash
+$ readlink -f $(which ex)
+```
+
+
 ## Coding Style
 
 Mycli requires code submissions to adhere to
--- a/changelog.md
+++ b/changelog.md
@@ -10,6 +10,7 @@ Internal Changes:
 -----------------
 
 * Drop support for Python 3.3 (Thanks: [Thomas Roten]).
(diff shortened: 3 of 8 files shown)


meeuw:
@Terseus thank you! I think it would be better to support the output from archlinux ex/vi command, I'll take a look if this is possible. You've also updated myclirc but I think these are also used as the default for end users... Testing mycli with mariadb sounds like a good idea, it would be nice if we can do this from Travis as well.

Terseus:
About the myclirc changes, i though of making a `test/myclirc` file to be used in tests, this:
* have the pro of separating default config and testing config.
* have the con of maintain both files when changing/adding config variables.

Mmm, I'll see if it's possible to load incrementally a `tests/myclirc` file so it's values replace the config from the default myclirc file.

Terseus:
@meeuw I've added support for the file `test/myclirc` which contains only the config variables overriden for testing :)
The feature was already built-in in mycli so it has been easy.

amjith:
Looks good to me. @meeuw If you don't have any objections go ahead and merge it. 

One test is failing in Python 2.7. I'll look into it later tonight. But the test failure is unrelated to this PR.

meeuw:
Hmm I'm not sure, the "stock" myclirc isn't tested anymore, might not be a big problem.

We've decided (before) that we only do pep8 changes for lines that are changed, I don't think anything changed except for the pep8 formatting.

But let's go ahead and merge and keep this in mind for future PRs :smile:

test/myclirc
```diff
@@ -0,0 +1,12 @@
...
+
+# This file is loaded after mycli/myclirc and should override only those
+# variables needed for testing.
+# To see what every variable does see mycli/myclirc
+
+[main]
+
+log_file = ~/.mycli.test.log
+log_level = DEBUG
+prompt = '\u@\h:\d> '
```
meeuw: As I think of it, maybe we can use the default prompt, expect_exact doesn't mind if the prompt starts with mycli or mariadb since you've removed it...

Terseus: You're right, I though `expect_exact` wouldn't work with the product type in the prompt but it works.
In fact it works with `less_chatty = False` now 😆 

Should I revert completely the new `tests/myclirc` file and just remove the product type from the prompt expected strings in the tests?

Terseus:
@meeuw
> We've decided (before) that we only do pep8 changes for lines that are changed, I don't think anything changed except for the pep8 formatting.

There was a `sys` module use without importing it.
I could just add the `sys` module but there was two pep8 errors so it looked a good idea to left the file clean.

Also, I think the `try/except` version of the `cli_cmd` construction is much easier to understand than the previous code (although it's much larger, of course; 11 lines vs 2 in the original version).

test/myclirc
meeuw: I'm sure we'll need a test myclirc for future changes, so I think we can keep it this way.

Terseus: @meeuw Old prompt restored (but not deleted, still in `test/myclirc`), if you want more changes to the config file just tell me ^_^

meeuw:
Thanks @Terseus!

Issue #534, opened 2017-10-27
`sudo mysql` logs in fine, `sudo mycli` gives permission denied; root has no password


cmcaine:
On ubuntu 16.04, mariadb 10.0.31-0ubuntu0.16.04.2

```
$  mycli --version
Version: 1.5.2

$ sudo mycli
(1698, "Access denied for user 'root'@'localhost'")
```

tsroten:
Hi @cmcaine I'm not that familiar with how `sudo` affects database clients on Ubuntu, but I believe you need to set a password for your `root` database user.

Also, if possible you might want to install mycli via `pip` instead of `apt-get` as we're currently on version 1.13.1 instead of 1.5.2.

meeuw:
Hmm, could you please post the output from `sudo mysql` and then the `\s` command?
Maybe your `~root/.my.cnf` contains a configuration which isn't supported by `mycli` ?

cmcaine:
This could be helpful(?):

```
$ sudo mysql --silent --skip-column-names -e"show grants"              
GRANT ALL PRIVILEGES ON *.* TO 'root'@'localhost' IDENTIFIED VIA unix_socket WITH GRANT OPTION
GRANT PROXY ON ''@'%' TO 'root'@'localhost' WITH GRANT OPTION
```

As requested:

```
$ sudo mysql            
[sudo] password for colin: 
Welcome to the MariaDB monitor.  Commands end with ; or \g.
Your MariaDB connection id is 54
Server version: 10.0.31-MariaDB-0ubuntu0.16.04.2 Ubuntu 16.04

Copyright (c) 2000, 2017, Oracle, MariaDB Corporation Ab and others.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

MariaDB [(none)]> \s
--------------
mysql  Ver 15.1 Distrib 10.0.31-MariaDB, for debian-linux-gnu (x86_64) using readline 5.2

Connection id:		54
Current database:	
Current user:		root@localhost
SSL:			Not in use
Current pager:		stdout
Using outfile:		''
Using delimiter:	;
Server:			MariaDB
Server version:		10.0.31-MariaDB-0ubuntu0.16.04.2 Ubuntu 16.04
Protocol version:	10
Connection:		Localhost via UNIX socket
Server characterset:	utf8mb4
Db     characterset:	utf8mb4
Client characterset:	utf8mb4
Conn.  characterset:	utf8mb4
[...]

PR #536, opened 2017-10-29
Default to unix socket connection

cmcaine:
## Description

Default to unix socket connection if host and port are unspecified.

 - Local users can be identified by UID without passwords
 - Matches mysql behaviour

## Checklist

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -357,16 +357,9 @@ class MyCli(object):
         else:
             socket = socket or cnf['socket']
         user = user or cnf['user'] or os.getenv('USER')
-        host = host or cnf['host'] or 'localhost'
-        port = port or cnf['port'] or 3306
+        host = host or cnf['host']
+        port = port or cnf['port']
         ssl = ssl or {}
-
-        try:
-            port = int(port)
-        except ValueError as e:
-            self.echo("Error: Invalid port number: '{0}'.".format(port),
-                        err=True, fg='red')
-            exit(1)
 
         passwd = passwd or cnf['password']
         charset = charset or cnf['default-character-set'] or 'utf8'
@@ -387,25 +380,62 @@ class MyCli(object):
 
         # Connect to the database.
 
-        try:
+        def _connect():
             try:
-                sqlexecute = SQLExecute(database, user, passwd, host, port,
-                        socket, charset, local_infile, ssl)
+                self.sqlexecute = SQLExecute(database, user, passwd, host, port,
+                                             socket, charset, local_infile, ssl)
             except OperationalError as e:
                 if ('Access denied for user' in e.args[1]):
-                    passwd = click.prompt('Password', hide_input=True,
-                                          show_default=False, type=str)
-                    sqlexecute = SQLExecute(database, user, passwd, host, port,
-                            socket, charset, local_infile, ssl)
+                    new_passwd = click.prompt('Password', hide_input=True,
+                                              show_default=False, type=str)
+                    self.sqlexecute = SQLExecute(database, user, new_passwd, host, port,
+                                                 socket, charset, local_infile, ssl)
                 else:
                     raise e
+
+        try:
+            if socket is host is port is None:
+                # Try a sensible default socket first (simplifies auth)
+                # If we get a connection error, try tcp/ip localhost
+                try:
+                    socket = '/var/run/mysqld/mysqld.sock'
+                    _connect()
+                except OperationalError as e:
+                    # These are "Can't open socket" and 2x "Can't connect"
+                    if [code for code in (2001, 2002, 2003) if code == e.args[0]]:
+                        self.logger.debug('Database connection failed: %r.', e)
+                        self.logger.error(
+                            "traceback: %r", traceback.format_exc())
+                        self.logger.debug('Retrying over TCP/IP')
+                        self.echo(str(e), err=True)
+                        self.echo(
+                            'Failed to connect by socket, retrying over TCP/IP', err=True)
+
+                        # Else fall back to TCP/IP localhost
+                        socket = ""
+                        host = 'localhost'
+                        port = 3306
+                        _connect()
+                    else:
+                        raise e
+            else:
+                host = host or 'localhost'
+                port = port or 3306
+
+                # Bad ports give particularly daft error messages
+                try:
+                    port = int(port)
+                except ValueError as e:
+                    self.echo("Error: Invalid port number: '{0}'.".format(port),
+                              err=True, fg='red')
+                    exit(1)
+
+                _connect()
         except Exception as e:  # Connecting to a database could fail.
             self.logger.debug('Database connection failed: %r.', e)
             self.logger.error("traceback: %r", traceback.format_exc())
             self.echo(str(e), err=True, fg='red')
             exit(1)
-
-        self.sqlexecute = sqlexecute
 
     def handle_editor_command(self, cli, document):
         """
--- a/changelog.md
+++ b/changelog.md
@@ -5,6 +5,7 @@ Features:
 ---------
 
 * Add `watch [seconds] query` command to repeat a query every [seconds] seconds (by default 5). (Thanks: [David Caro](https://github.com/Terseus))
+* Default to unix socket connection if host and port are unspecified. This simplifies authentication on some systems and matches mysql behaviour.
 
 Internal Changes:
 -----------------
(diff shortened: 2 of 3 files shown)


cmcaine:
See #534 

@meeuw

meeuw:
Wow, you've already implemented a fix, thanks a lot!
Could you please run `./setup lint -f` to fix the PEP8 issues?

cmcaine:
The lint is failing on 2.7, but it's fine everywhere else.

meeuw:
I suppose we can keep the current default socket location like you've supplied, if someone wishes we can check all known locations of the socket using os.path.exists() (without /tmp ofcourse :smile:)

Lint isn't failing but the tests are failing on a unicode string, I've restarted the job for py27.

You've done a thorough job handling different error codes, I don't know if it's complete but it looks good to me!

meeuw:
The error for python 2.7 is also in master, I will create a seperate PR to fix it.

tsroten:
@cmcaine Thanks for your work on this!!

I have a couple of comments I'll leave on the code. 

One thing I'm wondering, is that since mycli supports Windows, shouldn't we avoid trying to connect over a unix socket on a Windows machine?

mycli/main.py
```diff
@@ -387,26 +380,63 @@ def connect(self, database='', user='', passwd='', host='', port='',
...
-                            socket, charset, local_infile, ssl)
+                    new_passwd = click.prompt('Password', hide_input=True,
+                                              show_default=False, type=str)
+                    self.sqlexecute = SQLExecute(database, user, new_passwd, host, port,
+                                                 socket, charset, local_infile, ssl)
                 else:
                     raise e
+
+        try:
+            if socket is host is port is None:
```
tsroten: It's possible that these are blank, but not `None`, e.g. an empty string `''`.

mycli/main.py
```diff
@@ -387,26 +380,63 @@ def connect(self, database='', user='', passwd='', host='', port='',
...
+                # Try a sensible default socket first (simplifies auth)
+                # If we get a connection error, try tcp/ip localhost
+                try:
+                    socket = '/var/run/mysqld/mysqld.sock'
+                    _connect()
+                except OperationalError as e:
+                    # These are "Can't open socket" and 2x "Can't connect"
+                    if [code for code in (2001, 2002, 2003) if code == e.args[0]]:
```
tsroten: I'm curious why we have a loop here. Would this be a simpler way of doing the same thing?
```
if e.args[0] in (2001, 2002, 2003):
```

mycli/main.py
cmcaine: Empty strings are never valid values for any of these options (or for any other mysql options?), so maybe the loader functions should do that validation/coercion?

Otherwise

```python
if all((opt in (None, '') for opt in (socket, host, port))):
```

might be a reasonably readable way of rewriting it.

mycli/main.py
cmcaine: Yes, that's better and is how I normally do it! I was pretty exhausted when I wrote it :/

mycli/main.py
cmcaine: https://github.com/keleshev/schema is a validation library that's pretty simple

cmcaine:
Looking for the socket on windows probably doesn't do any harm, but I haven't tested it and it is pointless.

Would you like me to edit the PR some more or would you like to finish things up yourself?

mycli/main.py
j-bennet: Why not `if not (socket or host or port)`?

cmcaine: I don't want to treat `port = 0` as 'no value', though it is treated as 'no value' by the `port = port or 3306` line, now that I think about it, so I guess it doesn't matter - problems are just being moved along to pymysql.

I don't know how good pymysql's error messages are for this kind of bad config. If they're good enough maybe there's no point doing your own validation.

j-bennet: @cmcaine True, I forgot that port may be 0. However, I just tried changing default port to 0 on my local mysql installation, and it looks like the service defaults to 3306 in this case. Here is my.cnf:
```
--- ~ » cat /etc/my.cnf
[client]
port = 0
socket = /tmp/mysql.sock

[mysqld]
port = 0
socket = /tmp/mysql.sock
key_buffer_size = 16M
max_allowed_packet = 8M

[mysqldump]
quick
```

but here is what mysql thinks the port is set to:

```
mysql root@localhost:test> show global variables like 'port%';
+-----------------+---------+
| Variable_name   | Value   |
|-----------------+---------|
| port            | 3306    |
+-----------------+---------+
```

I think we can assume that 0 is not a valid port. Also, I'd rather defer to pymysql and not add more complexity with our own validation.

tsroten:
@cmcaine Thanks! If you have time, that'd be great. Otherwise, I should be able to update it later this week or this weekend.

mycli/main.py
cmcaine: Port 0 is never a valid port in TCP or UDP/IP.

If you connect to localhost, `mysql` ignores any port value and always uses the socket. I don't know how it does this. Presumably by reading the mysqld parts of the config files. Maybe a default is baked into the build.

If you have more than one server running on localhost. I think you have to set host to 127.0.0.1 and port to \<whatever> or just set socket to db2's socket. Which is confusing: `mysql -P 3307` shouldn't silently transmute to `mysql -S /path/to/db1.sock`.

> but here is what mysql thinks the port is set to:

My understanding is that `show global variables like 'port%';` shows the server's port, not the method that the client is using to connect. Try `status` instead.

---

I think this algorithm is least surprising:

1. If port is set, use TCP/IP
2. If UNIX socket is set use that
3. If both are set, throw an error
4. If neither are set:
    1. use TCP/IP if hostname is set and not localhost or if system is windows
    2. otherwise use UNIX socket at a default location, then maybe fall back to localhost:3306.

I think specifying both port and socket should be an error and that the current behaviour is bad:

 - mycli ignores socket if host or port are set at the commandline, but ignores the port otherwise, 
 - pymysql ignores the port,
 - mysql always uses socket whether you ask it to or not for localhost, 
 - All seem to ignore socket if host is not localhost (I get [1698](https://dev.mysql.com/doc/refman/5.7/en/error-messages-server.html#error_er_access_denied_no_password_error) if user is only identified by unix_sockets for `host == <public IP>`.

Best behaviour if config and commandline disagree would be, IMHO, to emit a warning that the config value for socket is being dropped in favour of the commandline argument. Also acceptable to just throw an error and refuse to connect.

j-bennet: The above sounds sensible. But I don't think that we should throw an error in 3. I think if port is set, we should just prefer that over socket.

cmcaine: That's the opposite behaviour to mysql.

j-bennet: Really? mysql will prefer socket if port is set in the command line, as in `mysql --port 5555`?

j-bennet: @cmcaine Just checked - yes it will. That behavior does not seem very user-friendly though. I think the user would expect to be connected the way he specified when running the CLI, i.e. if port or host are specified, they should be the preferred method.

cmcaine: I already did the checks and listed the behaviour of each tool, and said that it's bad :)

Commandline taking precedence over config is fine. I'd maybe emit a warning, but whatever. If port and socket are both set on the commandline, I think that should be an error. If they're both set in the config file, I think that should be an error.

I don't want to bikeshed this any more. This is a pretty small convenience feature. I've said what I'd do, but I don't really care one way or the another.

Someone with investment in this project should just take these comments and finish them up in a reasonably sane way (please not how mysql does it).

j-bennet: I was not trying to argue with you @cmcaine - you did good work on the PR. The results of your testing the tools above are also very helpful. My comments were only suggestions.

cmcaine: I didn't feel like you were, and I didn't intend to be rude. Your suggestions were helpful and have clarified the spec :)

meeuw:
Connecting to a "dead" socket won't give long timeouts, I think it's safe to try the socket first. We also had previous reports from users who where expecting this behaviour.

meeuw:
Thanks a lot @cmcaine!

cmcaine:
You're welcome :)

It looks like the comments from @tsroten and @j-bennet were never acted upon, though, so maybe this PR should be tidied up some more?

If you want those comments enacted and no one else is willing to do it, I could potentially do it sometime in the next few days.

j-bennet:
@cmcaine I think the loop thing is worth changing - in another small PR. There's no hurry though, it's not really a bug. As for checking the socket on Windows, I understand you're not able to test this on Windows, correct?

PR #537, opened 2017-10-29
Unicode output to click for Python 2.7

meeuw:
## Description
Unicode output to click for Python 2.7 (fixes failing behave tests in master)

I'm not sure if this is the right way for Python 2/3 compatibility

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -9,6 +9,7 @@ Features:
 Internal Changes:
 -----------------
 
+* Unicode output to click for Python 2.7 (Thanks: [Dick Marinus]).
 * Drop support for Python 3.3 (Thanks: [Thomas Roten]).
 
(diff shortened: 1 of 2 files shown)


tsroten:
Hmm I wonder why `output` is not already Unicode. Doesn't CLI Helpers return Unicode?

meeuw:
Hmm good point, I'll try to figure it out

tsroten:
Thanks! My concern is that we *should* be using Unicode internally, and in the PR we are decoding without an explicit encoding.

meeuw:
Yep, found it: https://github.com/dbcli/cli_helpers/pull/32 you're the best @tsroten :smile:

tsroten:
@meeuw  Nice job!!!

PR #539, opened 2017-11-03
merge _on_completions_refreshed / _swap_completer_objects functions

meeuw:
## Description
merge _on_completions_refreshed / _swap_completer_objects functions

## Checklist

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -776,14 +776,6 @@ class MyCli(object):
                 'Auto-completion refresh started in the background.')]
 
     def _on_completions_refreshed(self, new_completer):
-        self._swap_completer_objects(new_completer)
-
-        if self.cli:
-            # After refreshing, redraw the CLI to clear the statusbar
-            # "Refreshing completions..." indicator
-            self.cli.request_redraw()
-
-    def _swap_completer_objects(self, new_completer):
         """Swap the completer object in cli with the newly created completer.
         """
         with self._completer_lock:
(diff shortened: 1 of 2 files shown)


amjith:
What's the purpose of this change? 

The _swap_completer was it's own function because I wanted the lock to be as fine grained as possible. What do we achieve by merging these functions together?

meeuw:
Hmm am I missing something? How is locking involved? `_swap_completer_objects` is only used by `_on_completions_refreshed`, by merging the functions the code base will be smaller and the code "flow" is easier to follow. But I don't know how locking is involved, could you enlighten me? :smile:

mycli/main.py
```diff
@@ -776,14 +776,6 @@ def refresh_completions(self, reset=False):
...
-
-        if self.cli:
-            # After refreshing, redraw the CLI to clear the statusbar
-            # "Refreshing completions..." indicator
-            self.cli.request_redraw()
-
-    def _swap_completer_objects(self, new_completer):
         """Swap the completer object in cli with the newly created completer.
         """
         with self._completer_lock:
```
amjith: I was talking about this lock. 

The completion refresh happens in a background thread. When we are done with the refresh we want to swap the existing completion object with the newly refreshed completion object. So it is necessary to lock the access to the completion object before we can swap it out. 

But upon further digging I now realize that you haven't changed the things that are under locks. So my original comment is not valid. 

I'll take a look at the failing test and if it is nothing bad, i'll merge this in.

Issue #518, opened 2017-09-22
Feature request: aliases with parameters


Terseus:
It would be cool if the aliases (aka. favorite queries) have support for having arguments, so you can execute them with different arguments every time.

As an example, consider the following query:
```
SELECT * FROM table WHERE foreign_key IN (SELECT id FROM other_table WHERE some_value = $1);
```

Then, you could execute it like:

```
mysql user@host:db> \f example 10
```

This would execute the previous query replacing `$1` with the value `10`.

tsroten:
@Terseus This is a great idea. Thanks for filing the issue.

I've also thought about it prompting the user to fill in arguments that aren't provided. E.g.
```
mysql user@host:db> \f example
some_value? 10
```

tsroten:
@Terseus This feature is in the latest release of myci. Thanks again for the idea!

PR #542, opened 2017-11-25
[refs #518] add support for positional arguments for favorite queries

scrappysoft:
## Description

Add support for positional parameters in favorite queries.

Variables are referenced by positional number similar to shell variables.  Example:

`\fs user_by_name select * from users where name = '$1'`

Values are passed at invocation by position, with shell word and quoting rules.  Example:

`\f user_by_name "Skelly McDermott"`

Errors are presented if too few or too many arguments are given.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 A command line client for MySQL that can do auto-completion and syntax highlighting.
 
-HomePage: [http://mycli.net](http://mycli.net)  
+HomePage: [http://mycli.net](http://mycli.net)
 Documentation: [http://mycli.net/docs](http://mycli.net/docs)
 
 ![Completion](screenshots/tables.png)
@@ -96,11 +96,8 @@ Features
     - `SELECT * FROM <tab>` will only show table names.
     - `SELECT * FROM users WHERE <tab>` will only show column names.
 * Support for multiline queries.
-* Favorite queries. Save a query using `\fs alias query` and execute it with `\f alias` whenever you need.
-    - Also supports optional positional parameters of the form `$N` in the query
-    - Arguments are given after the name, with shell word and quoting rules
-    - `\fs user_by_name select * from users where name = '$1'`
-    - `\f user_by_name "Skelly McDermott"`
+* Favorite queries with optional positional parameters. Save a query using
+  `\fs alias query` and execute it with `\f alias` whenever you need.
 * Timing of sql statments and table rendering.
 * Config file is automatically created at ``~/.myclirc`` at first launch.
 * Log every query and its results to a file (disabled by default).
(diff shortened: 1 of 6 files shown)


scrappysoft:
note: the same tests failed locally on the base fork before I made any changes.

tsroten:
@scrappysoft Thanks for the pull request! This is great! I'll take a look at this over the weekend and look at what's going on with the test.

tsroten:
The failing test is a CLI Helpers bug that has been fixed, but hasn't been released, yet: https://github.com/dbcli/cli_helpers/pull/32

So, we can ignore that here for now. It looks like there are some PEP8 errors in the Python 3 tests. If you have time could you fix those? If not, I'll push a change for that tomorrow.

I'll probably move some of your documentation from the README to http://www.mycli.net/favorites so we can keep the README more oriented around communicating what features we have and not how to use all of them.

Overall, great work, thanks again!

meeuw:
:+1: I like the shell like parameter expansion (`$1`, `$2`, etc). Although it might be better to use `?` which is commonly used: https://www.owasp.org/index.php/Query_Parameterization_Cheat_Sheet

scrappysoft:
The idea is to add primitive argument support.  

A more complete solution would allow for named, rather than positional arguments.  That would force a change to the declaration syntax to be more like a (typeless) function declaration, e.g., `\f name(firstArg, secondArg) sqlhere`, and similar change to the invocation.  That would also allow for future support of prompting for missing arguments. 

Rather than doing something that complete, I opted for the suggested positional argument approach, with a shell-like syntax both in the declaration ($N substitution variables) and invocation (whitespace delimited words with quoting support.)  80/20 rule; it gets the job done with the right ratio of effort.

There are at least a few problems that I see with using ? as a parameter.  First, it means it cannot be repeated in the query; the value would have to be passed redundantly at invocation.  Second, the order in the query needs to match the order in the invocation.  Third, and most importantly, ? implies that it is type-safely inserted into a prepared statement, not text-substituted into the query.   That difference is significant, even if only dealing in string values.  "select * from tab where name = ?" vs "select * from tab where name = '$1'", where the latter has to consider quoting et al.  OWASP prefers prepared statements with value substitutions to avoid SQL Injection.  Personal-use favorites don't suffer from this problem.

meeuw:
Oh yes, I think you're right. I'm fine with mixing shell syntax in SQL.

tsroten:
@scrappysoft Thanks again for submitting this pull request! 👍

tsroten:
@scrappysoft I think it's fine either way, but I just noticed that the parameter count starts at `1` instead of `0` like bash. Was that an intentional design decision?

scrappysoft:
bash (and most others that support $N syntax) start arguments at 1, so yes, deliberate decision.  $0 in bash and similar shells is the command name, not the first argument to it, in regex is the entire match not the first match group, etc.

PR #546, opened 2017-12-02
Add sql-update/insert output format

meeuw:
## Description
New output formats sql-update and sql-insert.
TODO:
- [x] handle NULL
- [x] support composite primary keys (multiple columsn in WHERE) for sql-update

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,11 @@
+TBD
+=======
+
+Features:
+---------
+
+* Add sql-update/insert output format. (Thanks: [Dick Marinus]).
+
 1.14.0:
 =======
 
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -31,6 +31,7 @@ from pygments.token import Token
 
 from .packages.special.main import NO_QUERY
 from .packages.prompt_utils import confirm_destructive_query
+from .packages.tabular_output import sql_format
 import mycli.packages.special as special
 from .sqlcompleter import SQLCompleter
 from .clitoolbar import create_toolbar_tokens_func
@@ -110,6 +111,8 @@ class MyCli(object):
         special.set_timing_enabled(c['main'].as_bool('timing'))
         self.formatter = TabularOutputFormatter(
             format_name=c['main']['table_format'])
+        sql_format.register_new_formatter(self.formatter)
+        self.formatter.mycli = self
         self.syntax_style = c['main']['syntax_style']
         self.less_chatty = c['main'].as_bool('less_chatty')
         self.cli_style = c['colors']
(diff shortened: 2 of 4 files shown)


amjith:
This new table format seems a bit esoteric. Can you add a description of how you use this? 

I'm reluctant to approve a PR that implements a feature that will only be used by one person (you).

meeuw:
@amjith my first plan was to make this in cli_helpers: https://github.com/dbcli/cli_helpers/issues/12
I've created an asciinema.org here: https://asciinema.org/a/ZpfxszpDAGMeOmOEqtw0brGSD

But I agree this really needs some explaination, what do you think is appropriate? A blog post or an addition to README?

j-bennet:
@meeuw Hmm... like a `pg_dump`?

meeuw:
@j-bennet I think pg_dump is way faster, but it could be used like that (only one table each time).

I sometimes have to duplicate records in a table with a single change, maybe there's an easier way but at this moment I'm using this:
Inspect table: `select * from t1;`
Make a selection to duplicate `select * from t1 where a = 33`
Then I have to manually type-out all column names `select a, b, c, d, e, f, g, h from t1` (this is where this PR is usefull to copy-paste)
And finally I'll update the query and insert into the table like `insert into t1 select null, b, c, concat(d, 'dup'), e, f, g, h from t1`

For the sql-update output format, if you'd like to update a value in a table using a CLI I would:
Inspect table: `select * from t1`
Manually type-out the update query: `update t1 set b = 123 where a = 33`

Using the sql-update output format you'll be able to exactly copy/paste that query and only update the old value with the new value.

Again, if you know an easier way to update values using a CLI I'm all ears :smile:

I've embraced copy-paste in my CLI experience for a long time, it might be a bit of a contradiction. I think it's best of both worlds...

amjith:
> Again, if you know an easier way to update values using a CLI I'm all ears :smile:

There is a project called pgcli that has a feature which we might want to copy. :wink: 

Again, if you know an easier way to update values using a CLI I'm all ears :smile:

Again, if you know an easier way to update values using a CLI I'm all ears :smile:

https://asciinema.org/a/ZpfxszpDAGMeOmOEqtw0brGSD

When you type `SELECT * FROM foo` and move the cursor back to the `*` and hit `<tab>` it will expand all the columns in that table. 

That is slightly faster than copy/pasting queries and editing them to suit your needs. 

Thoughts?

amjith:
Generating the update and insert statements are still useful. But it is a very specific use case and I wonder if it should live as a plugin that can be installed for mycli or pgcli. People can install it as an addon if they need it. Not really sure if that accomplishes anything. I'll think about it.

meeuw:
@amjith that * expansion sounds great, we should definitely port that over to mycli!

I've already isolated most of this output format in the sql_format.py "package", I don't think it has a high impact on mycli...

PR #547, opened 2017-12-03
Also complete aliases in WHERE

meeuw:
## Description
Also complete aliases in WHERE.
This was a big annoyance for me. I often use JOIN's on tables which have common columns and these columns must be addressed by `tablename`.`column` or `aliasname`.`column`. With these changes the aliasses (and tablenames!) are suggested :tada:

## Checklist

### DIFF
--- a/mycli/packages/completion_engine.py
+++ b/mycli/packages/completion_engine.py
@@ -226,7 +226,7 @@ def suggest_based_on_last_token(token, text_before_cursor, full_text, identifier
                     {'type': 'view', 'schema': parent},
                     {'type': 'function', 'schema': parent}]
         else:
-            aliases = [t[2] or t[1] for t in tables]
+            aliases = [alias or table for (schema, table, alias) in tables]
             return [{'type': 'column', 'tables': tables},
                     {'type': 'function', 'schema': []},
                     {'type': 'alias', 'aliases': aliases},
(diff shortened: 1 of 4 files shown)


mycli/packages/completion_engine.py
```diff
@@ -218,16 +218,18 @@ def suggest_based_on_last_token(token, text_before_cursor, full_text, identifier
...
         if parent:
-            tables = extract_tables(full_text)
             tables = [t for t in tables if identifies(parent, *t)]
             return [{'type': 'column', 'tables': tables},
                     {'type': 'table', 'schema': parent},
                     {'type': 'view', 'schema': parent},
                     {'type': 'function', 'schema': parent}]
         else:
-            return [{'type': 'column', 'tables': extract_tables(full_text)},
+            aliases = [t[2] or t[1] for t in tables]
```
j-bennet: How about `aliases = [alias or table for (schema, table, alias) in tables]`? It's longer, but it would be more readable than `t[1]` and `t[2]`.

meeuw: You're completely right, I've copied this from another line in this file. So I think this should be generalized and rewritten like you've suggested.

amjith:
Is it possible to list the tables above the column names? 

The suggestions are currently sorted alphabetically and this buries the tablenames and aliases down the list. It would be better if we list them higher up in the menu.

If you think that is outside the scope of this PR, let me know and I'll merge this in as is.

meeuw:
Hmm I think the completions are always sorted by name in mycli (sorted in mycli.sqlcompleter.find_matches). There is some strangeness about back quotes, I think table names / aliases always should be suggested with back quotes.

Please merge this as-is, I'd like to fix this in another PR.

amjith:
:yay:

PR #559, opened 2018-01-05
Config nopager

fraoustin:
## Description

Add parameter in section "main" : nopager.
When I want to start mycli without pager and I do not have /etc/my.cnf

## Checklist


meeuw:
It should also be possible to use `skip-pager` in `~/.my.cnf`, but I've tried this and it doesn't work.
Why would be add this option to `.myclirc` instead of `~/.my.cnf` ?

meeuw:
Thanks for this PR! It's already possible to use `skip-pager = True` in `~/.my.cnf`, why would we add this option to `.myclirc` instead of `~/.my.cnf` ?

mycli/myclirc
```diff
@@ -80,6 +80,9 @@ auto_vertical_output = False
 # keyword casing preference. Possible values "lower", "upper", "auto"
 keyword_casing = auto
 
+# disabled pager on startup
+nopager = False
```
amjith: Let's use the name `enable_pager` and use the `True` and `False` values. 

Also move this option closer to the `pager` option in the config file.

fraoustin: for @meeuw ... on my pc i haven't the .my.cnf (i have not a mysql server local)
for @amjith ... good remark ... i work

meeuw: @fraoustin you don't need a mysql server to have `.my.cnf`, `.my.cnf` is in the same directory as `.myclirc`. The advantage of `.my.cnf` is that it's also used for the official mysql client (I tend to use both mycli and the official mysql client)

mycli/myclirc
```diff
@@ -80,6 +80,9 @@ auto_vertical_output = False
 # keyword casing preference. Possible values "lower", "upper", "auto"
 keyword_casing = auto
 
+# disabled pager on startup
+enable_pager = False
```
amjith: Can you default this to `True`?

mycli/main.py
```diff
@@ -794,7 +794,7 @@ def configure_pager(self):
         else:
             self.explicit_pager = False
 
-        if cnf['skip-pager']:
+        if cnf['skip-pager'] or self.config['main'].as_bool('enable_pager'):
```
tsroten: Now that it is renamed `enable_pager`, should we use `not` in front of it? It looks like right now it will disable the pager if `enable_pager` is set to `True`.

tsroten:
@fraoustin 
```
if cnf['skip-pager'] or not self.config['main'].as_bool('enable_pager'):
    special.disable_pager()
```

Isn't that going to disable every mycli user's pager as soon as this is merged and released? Since no one will have `enable_pager` set to `True` in their config files.

Right now the default behavior is an automatic pager that turns on when the output doesn't fit on the screen.

fraoustin:
@tsroten It's possible to set the init value to false for enable_pager ... the same behaviour as today. It's mind of your remark?

tsroten:
@fraoustin You're right, I was confused. This looks good to me 👍

Issue #569, opened 2018-02-05
Mycli crashes/quits when it can't create the history file


tsroten:
```
$ mycli
Password:
localhost:(none)> use test
Traceback (most recent call last):
  File "/home/linuxbrew/.linuxbrew/bin/mycli", line 11, in <module>
    load_entry_point('mycli==1.16.0', 'console_scripts', 'mycli')()
  File "/home/linuxbrew/.linuxbrew/Cellar/mycli/1.16.0/libexec/lib/python2.7/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/home/linuxbrew/.linuxbrew/Cellar/mycli/1.16.0/libexec/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/home/linuxbrew/.linuxbrew/Cellar/mycli/1.16.0/libexec/lib/python2.7/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/linuxbrew/.linuxbrew/Cellar/mycli/1.16.0/libexec/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/home/linuxbrew/.linuxbrew/Cellar/mycli/1.16.0/libexec/lib/python2.7/site-packages/mycli/main.py", line 1053, in cli
    mycli.run_cli()
  File "/home/linuxbrew/.linuxbrew/Cellar/mycli/1.16.0/libexec/lib/python2.7/site-packages/mycli/main.py", line 696, in run_cli
    one_iteration()
  File "/home/linuxbrew/.linuxbrew/Cellar/mycli/1.16.0/libexec/lib/python2.7/site-packages/mycli/main.py", line 512, in one_iteration
    document = self.cli.run()
  File "/home/linuxbrew/.linuxbrew/Cellar/mycli/1.16.0/libexec/lib/python2.7/site-packages/prompt_toolkit/interface.py", line 415, in run
    self.eventloop.run(self.input, self.create_eventloop_callbacks())
  File "/home/linuxbrew/.linuxbrew/Cellar/mycli/1.16.0/libexec/lib/python2.7/site-packages/prompt_toolkit/eventloop/posix.py", line 159, in run
    t()
[...]

PR #575, opened 2018-02-07
Handle missing log/history file directories.

tsroten:
## Description

See https://github.com/dbcli/mycli/issues/569. If the parent logging file or history file directories don't exist, this will warn the user about them, and continue with starting up mycli.

## Checklist


amjith:
🌻

Issue #538, opened 2017-11-02
mycli conflict with pbr


lloy:
in mycli entry_points  have context :
test:
tasks:
when i run python setup.py install in my test of project 
there are something wrong:
ERROR:root:Error parsing
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/pbr/core.py", line 115, in pbr
    attrs = util.cfg_to_args(path, dist.script_args)
  File "/usr/lib/python2.7/site-packages/pbr/util.py", line 267, in cfg_to_args
    wrap_commands(kwargs)
  File "/usr/lib/python2.7/site-packages/pbr/util.py", line 576, in wrap_commands
    cmdclass = ep.resolve()
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2417, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
ImportError: No module named tasks

tsroten:
This is something we should fix. Here is a related PR in CLI Helpers: https://github.com/dbcli/cli_helpers/pull/29

PR #577, opened 2018-02-08
Move build tasks to setup.py

tsroten:
## Description

- Moves the coverage scripts to `travis.yml` since we pretty much only run coverage reports from travis. This allows the `test` command to be simpler.
- Moves the `lint` and `test` commands to `setup.py` and adds them as distutil commands, not as entry points (see #538).
- Blacklists pytest 3.3.0 from `requirements-dev.txt` since it doesn't work whenever a warning is generated during testing (see https://github.com/pytest-dev/pytest/pull/2958). Here is an example failure: https://travis-ci.org/dbcli/mycli/jobs/338815262#L633
- Adds database charset and port to the environment variables that are passed via tox.

## Checklist


requirements-dev.txt
```diff
@@ -1,5 +1,5 @@
 mock
-pytest
+pytest!=3.3.0
```
tsroten: Travis was loading 3.3.0, which has a bug that was causing tests to fail:
https://github.com/pytest-dev/pytest/pull/2958

meeuw:
Nice work Thomas, kind of killing your darlings. I'll have a look for testing this on my system.

meeuw:
Do you want to add something to the changelog? It would be nice if we could check if this fixes the pbr issue #538, do you know how to use pbr? Maybe @lloy could help us with that?

PR #580, opened 2018-02-10
add option list-dsn

fraoustin:
## Description

add option list-dsn : get all alias in aliasdsn section of config file

## Checklist

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -965,7 +965,7 @@ class MyCli(object):
 @click.option('-d', '--dsn', default='', envvar='DSN',
               help='Use DSN configured into the [alias_dsn] section of myclirc file.')
 @click.option('--list-dsn', 'list_dsn', is_flag=True,
-        help='list of DSN configured into the [alias_dsn] section of pgclirc file.')
+        help='list of DSN configured into the [alias_dsn] section of myclirc file.')
 @click.option('-R', '--prompt', 'prompt',
               help='Prompt format (Default: "{0}").'.format(
                   MyCli.default_prompt))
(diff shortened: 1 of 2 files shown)


tsroten:
Great idea!!! This is super helpful!

A couple of thoughts:
- can we mask the passwords in the output?
- can we change “myclirc” in the output to the actual config file location? This could be overrided and might not be named myclirc

mycli/main.py
```diff
@@ -964,6 +964,8 @@ def get_last_query(self):
 @click.option('-D', '--database', 'dbname', help='Database to use.')
 @click.option('-d', '--dsn', default='', envvar='DSN',
               help='Use DSN configured into the [alias_dsn] section of myclirc file.')
+@click.option('--list-dsn', 'list_dsn', is_flag=True,
+        help='list of DSN configured into the [alias_dsn] section of pgclirc file.')
```
tsroten: Looks like a copy/paste mixup - “pgclirc”

meeuw:
This looks certainly useful, for me it would be nice (for https://github.com/meeuw/wrapc) to only output the dsn names. I'm using wrapc for bash autocompletion.

mycli/main.py
```diff
@@ -1014,7 +1016,16 @@ def cli(database, user, host, port, socket, password, dbname,
                   defaults_file=defaults_file, login_path=login_path,
                   auto_vertical_output=auto_vertical_output, warn=warn,
                   myclirc=myclirc)
-
+    if list_dsn :
+        try:
+            for alias in mycli.config['alias_dsn']:
+                click.secho(alias + " : " + mycli.config['alias_dsn'][alias])
+            sys.exit(0)
+        except Exception as err:
```
meeuw: Could you please change `Exception` to the exception you're trying to catch? Now every exception in the try block will be reported as `Invalid DSN`. This `except Exception as e` is used in other places but there the exception is reported without translation (`click.secho(str(e), err=True, fg='red')`).

mycli/main.py
fraoustin: exact, i change

mycli/main.py
fraoustin: exact, i add keyerror

meeuw:
Thanks @fraoustin!

PR #589, opened 2018-03-27
Output password prompt to stderr

ushuz:
## Description

Output `Password:` prompt to stderr, so that `Password:` prompt won't end up in the result file when we redirect mycli output into a file, e.g. `mycli -h some-server-require-password -e 'xxx' > result.txt`

## Checklist


meeuw:
Makes sense! Thanks!

PR #597, opened 2018-05-15
Add git-blame-ignore revs in anticipation of hyper-blame.

amjith:
## Description
https://commondatastorage.googleapis.com/chrome-infra-docs/flat/depot_tools/docs/html/git-hyper-blame.html

This will be useful when we use hyper-blame to ignore certain revs (such as running black on the entire repo).


tsroten:
Very neat, I had no idea about this feature in git

PR #604, opened 2018-05-27
fix thanks_picker

meeuw:
## Description
Fix thanks_picker, only the last file was used (contents was assigned instead of appended).
Besides this only lines which started with `*` were used, in `mycli/AUTHORS` all names start with double spaces.

## Checklist

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -61,6 +61,7 @@ except ImportError:
 from pymysql import OperationalError
 
 from collections import namedtuple
+import re
 
 # Query tuples are used for maintaining history
 Query = namedtuple('Query', ['query', 'successful', 'mutating'])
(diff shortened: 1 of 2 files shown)


tsroten:
Nice job! Thanks @meeuw

mycli/main.py
```diff
@@ -1181,11 +1182,14 @@ def is_select(status):
 
 
 def thanks_picker(files=()):
+    contents = []
     for filename in files:
```
msztolcman: Why not [fileinput](https://docs.python.org/3/library/fileinput.html)?

belartodmarinus: I don't know, I only fixed this bug :smile: but you're right, fileinput would make sense! I'll see if I can fix this in another PR. Thanks for the suggestion!

meeuw: I don't know, I only fixed this bug :smile: but you're right, fileinput would make sense! I'll see if I can fix this in another PR. Thanks for the suggestion!

Issue #601, opened 2018-05-24
error on csv batch mode when `\n` in field


dragonkid:
there should be a quote around the value when there is a `\n` in it. and the test case should be like this.

```python
def test_batch_mode_csv(executor):
    run(executor, '''create table test(a text, b text)''')
    run(executor, '''insert into test (a, b) values('abc\nnewline', 'def'), ('ghi', 'jkl')''')

    sql = 'select * from test;'

    runner = CliRunner()
    result = runner.invoke(cli, args=CLI_ARGS + ['--csv'], input=sql)

    expected = 'a,b\n"abc\nnewline",def\nghi,jkl\n'

    assert result.exit_code == 0
    assert expected in "".join(result.output)
```

`quoting=csv.QUOTE_NONNUMERIC` may be the easiest way to fix it.

tsroten:
@dragonkid This is now fixed in the latest release of mycli (1.17.0).

If you installed via pip, you can upgrade via:
```bash
pip install -U mycli
```

PR #605, opened 2018-05-27
Use unix CSV dialect.

tsroten:
## Description

This uses the `unix` CSV dialect so we don't have to worry about breaking the CSV output with line terminators and delimiters since it quotes every field.

Addresses https://github.com/dbcli/mycli/issues/601

## Checklist


amjith:
Nice solution. :bo

PR #623, opened 2018-07-20
Print server version as part of welcome message.

j-bennet:
## Description

Print server version, unless we're being "less chatty".

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,11 @@
 TBD
 ===
 
+Features:
+---------
+
+* Display server version in welcome message (Thanks: [Irina Truong]).
+
 Internal:
 ---------
 
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -508,6 +508,7 @@ class MyCli(object):
         key_binding_manager = mycli_bindings()
 
         if not self.less_chatty:
+            print('Server:', ' '.join(sqlexecute.server_type()))
             print('Version:', __version__)
             print('Chat: https://gitter.im/dbcli/mycli')
             print('Mail: https://groups.google.com/forum/#!forum/mycli-users')
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -508,8 +508,8 @@ class MyCli(object):
         key_binding_manager = mycli_bindings()
 
         if not self.less_chatty:
-            print('Server:', ' '.join(sqlexecute.server_type()))
-            print('Version:', __version__)
+            print(' '.join(sqlexecute.server_type()))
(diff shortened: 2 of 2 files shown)


amjith:
Why? What problem does it solve?

j-bennet:
Because I keep having to look up `select version()`.

I try to keep my local postgres version in line with our production servers. If it's right there in my face I will not forget to update.

meeuw:
For mysql there is the status command `\s` isn't that easier?

tsroten:
@j-bennet When this pull request first came up, I was thinking I wouldn't like it, but now after trying it out and thinking about it some more, I think it's nice 😄 

What do you think about adding "Client" or "mycli" to the line below the server version. It's a little ambiguous now:

```
$ mycli
Server: mysql 5.7.21-0ubuntu0.17.10.1
Version: 1.17.0
```

j-bennet:
@tsroten Yes, I think saying `Client version`, or `mycli version` instead of just "version" makes sense.

I'm curious, what's your reason for disliking the version string?

tsroten:
@j-bennet At first, I thought, "The user can simply type `status` to get the information they need."

While it's true that it's easy to get the version already, I kind of like the introductory text of mycli showing a brief bit of information about both the client **and** the server it's connecting to. I think it makes sense to show some information about what server you've stumbled upon :smile:

Plus, the official client also shows the server version when you connect.

So, 👍 

By the way, I like `mycli version` as well! Or even just `mycli X.X.X`

j-bennet:
Ok. It looks like this now:

```
$ mycli -u root
mysql 10.1.26-MariaDB-0+deb9u1
mycli 1.17.0
Chat: https://gitter.im/dbcli/mycli
Mail: https://groups.google.com/forum/#!forum/mycli-users
Home: http://mycli.net
Thanks to the contributor - Martijn Engler
```

mikeerickson:
No offense, but the 3 lines for chat, mail and home are far less important than the server version.  Maybe a one liner to point to repo is sufficient (and the report could have all this info in the README that is displayed) 

Just saying, if you are concerned with all the info being displayed.

mikeerickson:
Oh, and yes, the spark startup is WAY obtrusive.

meeuw:
LGTM! We might also add a date/time of invoking mycli for in the case we'll also end up as a Hollywood prop: https://nmap.org/movies/ ... I already imagine a scene where a data engineer uses a left join!

tsroten:
@j-bennet Looks good! Thanks!

Issue #18, opened 2015-05-21
Connection via ssh


msztolcman:
I don't know this is possible for current mycli architecture and libs, but many GUI tools have an option to connect via ssh to remote host, and then connect to local mysql. This is very useful, because often admins do not allow to connect to mysql from remote hosts, or only from specified address. And often do not want to install additional software (like mycli) on this kind of host :)

gled-rs:
reported as working correctly patching mycli/sqlexecute.py

Thanks @grooverdan

gawel:
I've made a script as a workaround. Open a tunnel, read ~/.my.cnf on the remote host and connect using those infos: https://gist.github.com/gawel/190efd5b2109e4872407

gled-rs:
@gawel You just need to monkey patch mycli/sqlexecute with @grooverdan pseudocode to solve that. The necessary changes in PyMysql and Paramiko are deployed.

No need for a wrapper ;)

amjith:
You guys are welcome to send me a PR instead of monkey patching it. :smile: 

After all this is an open source project.

gled-rs:
Will do once I get it cleaned, so much hardcoded stuff for now that I would be ashamed to produce a PR ;)

seschwar:
@gled-rs: How's progress on the clean up? I'd love to see this feature included.

nick-hahner:
Any progress updates? For now I'm just using an autossh tunnel in a screen session. Not as pretty, but it works.

meeuw:
I don't know if it's useful for you but I've created a wrapper for mycli which support bash completion and ssh tunnels. https://github.com/meeuw/wrapc

tsroten:
This feature is now out in version 1.18.0. You can get the newest version via `pip` or [check out the release](https://pypi.org/project/mycli/#files).

PR #625, opened 2018-07-24
I wish you can add the "ssh forward" in this tools , although there are some bugs that I do not find. 

iubuntu:
## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,177 +1,38 @@
-# mycli
-
-[![Build Status](https://travis-ci.org/dbcli/mycli.svg?branch=master)](https://travis-ci.org/dbcli/mycli)
-[![PyPI](https://img.shields.io/pypi/v/mycli.svg?style=plastic)](https://pypi.python.org/pypi/mycli)
-[![Join the chat at https://gitter.im/dbcli/mycli](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/dbcli/mycli?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
-
-A command line client for MySQL that can do auto-completion and syntax highlighting.
-
-HomePage: [http://mycli.net](http://mycli.net)  
-Documentation: [http://mycli.net/docs](http://mycli.net/docs)
-
-![Completion](screenshots/tables.png)
-![CompletionGif](screenshots/main.gif)
-
-Postgres Equivalent: [http://pgcli.com](http://pgcli.com)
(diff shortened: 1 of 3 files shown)


tsroten:
@iubuntu Hello! Thanks for your work on this 😄 

It looks like this pull request needs a few things to make it work well with mycli:
1. The README is updated for a different project. Can you undo your changes to it?
2. It looks like there are a few `print()` statements that probably shouldn't be there. I don't think we need to output anything to the user since they'll know it's connecting over SSH because they will have passed in the necessary command-line arguments.
3. The `setup.py` file needs to be updated with the new dependencies.

Here are a few ideas that might make the user's experience better:
1. The `--jump` option seems like it isn't very self-explanatory. Maybe simply `--ssh` would work?
2. It would be nice to include the ability to pass in the SSH password.
3. Can we make the command-line options use hypens so they're easier to read? E.g. `--ssh-key`, `--ssh-user`, etc.

tsroten:
This relates to https://github.com/dbcli/mycli/issues/18

meeuw:
@iubuntu I've created https://github.com/dbcli/mycli/pull/634 which does the same as you're trying to achieve in this PR. Could you please test it (you can install it using `pip install git+http://github.com/dbcli/mycli@feature/ssh_transport`).

iubuntu:
@tsroten  Thanks for your to review the codes and give me suggestions.
I have changed my code : 
 1.  The `README`  is undone. 
 2.  I removed few `print()`  . 
 3.  I added the `sshtunnel` in `setup.py` file.

In addition, 

1.  the `--jump` option is renamed to `--ssh`
2. add `ssh-password` option to support password
3. rename `sshport` `sshusername` `sshkey` to `ssh_port` `ssh_user` `ssh_key`


I am not immediately to replay your comment,   because I  found sone problems . 
 1.   I try to add `connect_uri(` but is not work
 2.  I try to add options in configure files.

iubuntu:
@meeuw  Thanks , I viewed your code, It is super better and clear than that I wrote . 

After I tested the code, I find a small error  in sqlexecute.py:`SQLExecute.connect`.   when I  set that  `--ssh-port` is not 22,  the `ssh_port`  in function named `connection` is `22`.

meeuw:
You're right iubuntu! I've fixed it, thanks for reporting!

tsroten:
@iubuntu Thanks again for your pull request! We've already merged #634, so I'm going to close this.

Issue #628, opened 2018-07-31
use specified database in combination with --dsn


meeuw:
If you use: `mycli --dsn testdsn dbname` 
`dbname` should be used to connect to that database.
At this moment the (optional) database is used from dsn specification.

PR #630, opened 2018-08-03
when DSN is used allow overrides from mycli arguments

meeuw:
## Description
when DSN is used allow overrides from mycli arguments, fixes https://github.com/dbcli/mycli/issues/628

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,6 +5,12 @@ Features:
 ---------
 
 * Display server version in welcome message (Thanks: [Irina Truong]).
+* Set `program_name` connection attribute (Thanks: [Dick Marinus]).
+
+Bug Fixes:
+----------
+
+* When DSN is used allow overrides from mycli arguments (Thanks: [Dick Marinus]).
 
 Internal:
 ---------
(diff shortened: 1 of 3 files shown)


tsroten:
@meeuw Nice! I'll give this a test drive tonight and see how it works. 😄

tsroten:
This works well overall. One of the issues I came across is that now you can't override environment variables via a DSN. For example, if you run these, your connection will hang:

```bash
$ export MYSQL_HOST=foobar
$ mycli mysql://user@localhost
(2003, "Can't connect to MySQL server on 'foobar' ([Errno -3] Temporary failure in name resolution)")
```

It won't use the `localhost` part of the DSN even those that's the value I manually passed to mycli.

I would think the order of precedence would go:
1. Command-line arguments
2. DSN
3. Environment variables

What do you think?

meeuw:
Hi Thomas, thanks for your review!

The order of precedence is kind of tricky because:
- DSN (contains the database name) can be specified from `[alias_dsn]`, as a command-line argument  (database) and as a environment variable. What I mean is that you might call the DSN a command-line argument or even an environment variable.
- The database can be specified as a command-line argument (`--database`) or just as an unparsed argument (without any `-` / `--` flag)
- Environment variables are now handled by `@click.option(envvar=` (`click.option` preferences command line arguments above environment variables)

So I'd stick the the main order (for username, password, hostname and port):
1. Command-line arguments
2. Environment variables
3. Configuration files

And for the database to connect:
1. Command-line argument `--database`
2. DSN (from `[alias_dsn]`) or `--dsn` or envvar `DSN`
3. Configuration files

The DSN should be used in the following order:
1. Command-line argument `--dsn`
2. Environment variable `DSN`
3. Unparsed command-line argument (`database`)

Issue #631, opened 2018-08-07
Error outputting euro symbol '€' in unicode using sql-insert format


jtaylor100:
While trying to dump the contents of a table to a file made of inserts, queries fail with red errors. I ran the following to produce the error:
```console
mysql user@host:database> \T sql-insert
Changed table format to sql-insert
Time: 0.000s
mysql user@host:database> \o ~/table-dump.sql
Time: 0.000s
mysql user@host:database> select * from my_table
write() argument 1 must be unicode, not str
```

When running without setting an output file, I get the following:
```console
mysql user@host:database> \T sql-insert
Changed table format to sql-insert
Time: 0.000s
mysql user@host:database> select * from my_table
'ascii' codec can't encode character u'\u20ac' in position 207: ordinal not in range(128)
```
`'\u20ac'` is the Euro symbol which is included in a string field within `my_table`, so this suggests that unicode output with the sql-insert format currently doesn't work.

Edit: Environment info

| Thing | Version |
| :-- | --: |
| Mycli  | **1.17.0** |
| MySQL Server | **5.5.56**|
| Ubuntu | **18.04.1 LTS** |

meeuw:
Hi @jtaylor100 thanks for reporting this issue, I will fix this but in the meantime you can workaround this by using Python 3. BTW nice to see you're using two features (\once and sql-insert) I've introduced to mycli!

tsroten:
@jtaylor100 - @meeuw's fix is now released in version 1.18.0! You can get the newest version via `pip` or [check out the release](https://pypi.org/project/mycli/#files).

jtaylor100:
Awesome, thanks! :+1:

PR #632, opened 2018-08-07
Convert `sql_format` to unicode strings for py27 compatibility

meeuw:
## Description
Convert `sql_format` to unicode strings for py27 compatibility
Fix for https://github.com/dbcli/mycli/issues/631

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,10 @@ Features:
 
 * Display server version in welcome message (Thanks: [Irina Truong]).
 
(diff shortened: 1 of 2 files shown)


tsroten:
@meeuw Nice job fixing this. Since this file is pretty small and isolated, would it be better to add `from __future__ import unicode_literals` at the top instead of the manual `u`s?

meeuw:
Never thought about that, thanks, changed.

PR #633, opened 2018-08-08
prompt-toolkit-2.0

meeuw:
## Description
port promp-toolkit-2.0 changes from pgcli

## Checklist


mycli/myclirc
```diff
@@ -43,6 +43,12 @@ table_format = ascii
 # Screenshots at http://mycli.net/syntax
 syntax_style = default
 
+# Keybindings:
+# When Vi mode is enabled you can use modal editing features offered by Vi in the REPL.
+# When Vi mode is disabled emacs keybindings such as Ctrl-A for home and Ctrl-E
+# for end are available in the REPL.
+vi = False
```
tsroten: Should we use the existing `myclirc` option for this? The one below on line 55.

meeuw: Ah yes, I've searched for vi in myclirc and didn't found anything, it was called emacs :smile:. I've restored/fixed `mycli/clitoolbar.py`.

test/test_clistyle.py
```diff
@@ -1,12 +1,14 @@
...
 """Test the mycli.clistyle module."""
+import pytest
 
 from pygments.style import Style
 from pygments.token import Token
 
 from mycli.clistyle import style_factory
 
 
+@pytest.mark.skip(reason="incompatible with new prompt toolkit")
```
meeuw: @tsroten do we still want these tests? I don't know how to do these tests in promt toolkit 2.0.

mycli/clitoolbar.py
```diff
@@ -1,48 +1,48 @@
...
 
 
-def create_toolbar_tokens_func(get_is_refreshing, show_fish_help):
-    """
-    Return a function that generates the toolbar tokens.
-    """
-    token = Token.Toolbar
 
-    def get_toolbar_tokens(cli):
+def create_toolbar_tokens_func(mycli, show_fish_help):
```
meeuw: TODO: Change this to `get_app` like pgcli's branch

meeuw:
- the toolbar should hide "emacs mode" and only show "vi mode" when vi mode is enabled (like master branch)

test/test_clistyle.py
tsroten: @meeuw I don't think we need these tests any longer. Thanks for noticing that!

PR #634, opened 2018-08-12
Connect using a SSH transport

meeuw:
## Description
Connect using a SSH transport

This only works when paramiko is installed, I don't know if we want to hard depend on paramiko (with all it's dependencies).

## Checklist

### DIFF
--- a/mycli/sqlexecute.py
+++ b/mycli/sqlexecute.py
@@ -6,6 +6,10 @@ from pymysql.constants import FIELD_TYPE
 from pymysql.converters import (convert_mysql_timestamp, convert_datetime,
                                 convert_timedelta, convert_date, conversions,
                                 decoders)
+try:
+    import paramiko
+except:
+    paramiko = False
 
 _logger = logging.getLogger(__name__)
 
@@ -37,7 +41,8 @@ class SQLExecute(object):
                                     order by table_name,ordinal_position'''
 
     def __init__(self, database, user, password, host, port, socket, charset,
-                 local_infile, ssl=False):
+                 local_infile, ssl, ssh_user, ssh_host, ssh_port, ssh_password,
+                 ssh_key_filename):
         self.dbname = database
         self.user = user
         self.password = password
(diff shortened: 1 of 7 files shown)


mycli/sqlexecute.py
```diff
@@ -80,12 +98,33 @@ def connect(self, database=None, user=None, password=None, host=None,
...
+            database=db, user=user, password=password, host=host, port=port,
+            unix_socket=socket, use_unicode=True, charset=charset,
+            autocommit=True, client_flag=pymysql.constants.CLIENT.INTERACTIVE,
+            local_infile=local_infile, conv=conv, ssl=ssl, program_name="mycli",
+            defer_connect=defer_connect
+        )
+
+        if ssh_host and paramiko:
+            if not ssh_user:
+                ssh_user = getpass.getuser()
```
tsroten: @meeuw It looks like paramiko will already call `getpass.getuser()` if the username is `None`. So, maybe we don't need this on the mycli side?

meeuw: thanks, you're right.

tsroten:
@meeuw I'm not sure exactly what my issue is, but I can't get this to work on python 2.7 or 3.6. I get an error `'Channel' object has no attribute 'recv_into'`. It looks like this might fix it: https://github.com/paramiko/paramiko/pull/553

@grooverdan mentions that PR is required on a PyMySQL issue: https://github.com/PyMySQL/PyMySQL/issues/77

meeuw:
That's stange! I presume you're testing this on MacOS? Have you tried running it on Linux?

tsroten:
@meeuw Yeah, that was tested on Linux (both a Linux client and a Linux server).

I just tested a MacOS client connecting to a Linux SSH server and that worked fine. I'm going to do some more testing with some Linux boxes real quick.

tsroten:
@meeuw Okay, everything continues to work well on the MacOS computer. On the Linux servers I got a variety of errors, some of which I think are related to the ssh server configuration. I'm not too worried about that at this point.

One thing we might want to improve (not in this pull request) is that `use` opens up a new connection. So, when you are using SSH, it opens another SSH connection and then another database connection. It potentially takes a second or two. That's unnecessary since all `use` is ultimately doing is setting the `sqlexecute.dbname` instance variable and triggering the completions thread to run. A new connection isn't needed. But, that's an existing issue, not related to the SSH changes.

I added a couple of changes to the click help messages.

I do think we should add paramik` to the `setup.py` file, maybe under the optional name `ssh`. What do you think?

meeuw:
Yes, it's a good idea to add an optional name in `setup.py` for `ssh`, just committed that.
I only see this as a convenience feature, I'll definately keep using my own tunnel script to reuse the same ssh connection.

PR #645, opened 2018-09-05
remove *.swp from gitignore

meeuw:
## Description
remove `*.swp` from gitignore.

Vim uses `.swp` files as a file lock / temporary file when a file is opened, for me it's a good reminder when doing `git status` that I still have files open.

I've tried adding stuff to `.git/info/exclude` but I cannot find anything to override this in `.gitignore`. I don't know any other trick to get some alarm when files are still open in vim when juggling with git. (I'm open for suggestions here). I hope this can be merged and people will add this to their `.git/info/exclude` file when they think they need this.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,6 @@
 .vagrant
 *.pyc
 *.deb
-*.swp
 .cache/
 .coverage
 .coverage.*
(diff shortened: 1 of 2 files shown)


amjith:
Looks good to me. I typically have *.swp listed in ~/.gitignore file which is global for all git repos.

Issue #648, opened 2018-09-06
Search .myclirc in same and directories above


eirnym:
I request an ability from `mycli` tool  to search it's `.myclirc` file in local directory, directory above till `${HOME}` or `/`. 

Current alternative solutions have their downsides. For example, This behaviour can be achieved by manual specifying rc file with constant remembering to open `mycli` in a specific directory it's located.  Which is quite impractical if project path is long enough. 

Alternatively automatic behavior can be achieved by hooking `cd` which impacts shell performance as `mycli` command is on the bottom of my run list

Reasoning: For some projects I have their own myclirc and want to keep some settings per project. like history file location, DSNs and so on.

amjith:
This feature request reminds me of a vim feature where you can specify where to search for the `tags` file. 

```
set tags=tags;/
```

https://stackoverflow.com/questions/5017500/vim-difficulty-setting-up-ctags-source-in-subdirectories-dont-see-tags-file-i/5019111#5019111

I'd be wary of changing the default behavior to always search from the local folder, but I wonder if we can use a command line option to do this. 

@tsroten @meeuw Thoughts?

eirnym:
I agree to have a command line option or (better) ~/.myclirc option for it.

tsroten:
@eirnym This is now live in version 1.18.0! You can get the newest version via `pip` or [check out the release](https://pypi.org/project/mycli/#files).

eirnym:
Thank you for the release!

PR #652, opened 2018-09-11
feature(myclirc): use myclirc in local directory first if it exists.  issue #648

KeatingSmith:
## Description

Search .myclirc in same and directories above.

I request an ability from mycli tool to search it's .myclirc file in local directory, directory above till ${HOME} or /.

## Checklist

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -54,6 +54,7 @@ click.disable_unicode_literals_warning = True
 try:
     from urlparse import urlparse
     from urlparse import unquote
+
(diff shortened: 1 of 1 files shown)


meeuw:
Thanks for your pull request, I think you've accidentally closed it because I'd love to merge it.

We're using pep8radius to only do PEP8 fixes to the lines you've changed, could you please revert your last commit and use pep8radius? If you don't know how to do this I can have a look for you.

PR #653, opened 2018-09-11
feature(myclirc): use myclirc in local directory first if it exists

KeatingSmith:
## Description

resolve issue #648

Search .myclirc in same and directories above.

I request an ability from mycli tool to search it's .myclirc file in local directory, directory above till ${HOME} or /.

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -8,6 +8,7 @@ Features:
 * Set `program_name` connection attribute (Thanks: [Dick Marinus]).
 * Use `return` to terminate a generator (Thanks: [Zhongyang Guan]).
 * Add `SAVEPOINT` to SQLCompleter (Thanks: [Huachao Mao]).
+* Seach `${PWD}/.myclirc`, then `${HOMW}/.myclirc`, last `/etc/.myclirc` (Thanks: [QiaoHao Peng])
 
 Bug Fixes:
 ----------
--- a/changelog.md
+++ b/changelog.md
@@ -8,7 +8,7 @@ Features:
 * Set `program_name` connection attribute (Thanks: [Dick Marinus]).
 * Use `return` to terminate a generator (Thanks: [Zhongyang Guan]).
 * Add `SAVEPOINT` to SQLCompleter (Thanks: [Huachao Mao]).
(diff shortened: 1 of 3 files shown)


meeuw:
Nice feature! But I think it would be better to append `os.path.join(os.getcwd(), ".myclirc")` to `config_files` at https://github.com/dbcli/mycli/blob/1060644cc72ab1e0f2161044ff6c8d2ea5b83902/mycli/main.py#L111

tsroten:
> 
> 
> Nice feature! But I think it would be better to append `os.path.join(os.getcwd(), ".myclirc")` to `config_files` at
> 
> [mycli/mycli/main.py](https://github.com/dbcli/mycli/blob/1060644cc72ab1e0f2161044ff6c8d2ea5b83902/mycli/main.py#L111)
> 
> Line 111 in [1060644](/dbcli/mycli/commit/1060644cc72ab1e0f2161044ff6c8d2ea5b83902)
>  config_files = ([self.default_config_file] + self.system_config_files +

Great idea @meeuw 😄. Plus, we need to make sure that people can still specify the myclirc file on the command-line via the `--myclirc` argument.


@KeatingSmith Thanks for working on this feature! This is a neat idea. 😃

KeatingSmith:
I have change it, and test it.

mycli/main.py
```diff
@@ -89,7 +89,7 @@ class MyCli(object):
     ]
 
     default_config_file = os.path.join(PACKAGE_ROOT, 'myclirc')
-
+    pwd_config_file = os.path.join(os.getcwd(), "")
```
meeuw: I think you'll need to join the cwd and ".myclirc" instead of an empty string...

changelog.md
```diff
@@ -8,6 +8,7 @@ Features:
 * Set `program_name` connection attribute (Thanks: [Dick Marinus]).
 * Use `return` to terminate a generator (Thanks: [Zhongyang Guan]).
 * Add `SAVEPOINT` to SQLCompleter (Thanks: [Huachao Mao]).
+* Seach `${PWD}/.myclirc`, then `${HOMW}/.myclirc`, last `/etc/.myclirc` (Thanks: [QiaoHao Peng])
```
meeuw: `HOMW` should be `HOME` ?

PR #665, opened 2018-10-12
Revert "Revert "prompt-toolkit-2.0""

tsroten:
This reapplies #633 so we can fix the failing test.


tsroten:
@meeuw I just realized that my revert here will probably mess up the git history. Let me undo this real fast.

PR #674, opened 2018-10-24
Fix pytest in Fedora mock

meeuw:
## Description
Fix pytest in Fedora mock

## Checklist

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -53,7 +53,6 @@ click.disable_unicode_literals_warning = True
 try:
     from urlparse import urlparse
     from urlparse import unquote
-    FileNotFoundError = OSError
 except ImportError:
     from urllib.parse import urlparse
     from urllib.parse import unquote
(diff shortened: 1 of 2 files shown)


mycli/main.py
```diff
@@ -1160,7 +1160,7 @@ def cli(database, user, host, port, socket, password, dbname,
 
         try:
             sys.stdin = open('/dev/tty')
-        except FileNotFoundError:
+        except (FileNotFoundError, OSError):
```
tsroten: So, the way I originally implemented this probably wasn't very good. Here `FileNotFoundError` might [actually be `OSError`](https://github.com/dbcli/mycli/blob/129b579f0c2766f69abe3a9058db264b9a7480af/mycli/main.py#L53:L59). That's even more confusing in that `FileNotFoundError` is a subclass of `OSError` in Python 3.

How about we just catch `IOError` (Python2) and `OSError` (Python3)? And, get rid of the weird import thing I did with `FileNotFoundError`?

meeuw:
@tsroten something like this? There is some weirdness in Fedora mock (which runs in a systemd container) and open() fails with OSError device cannot be opened (the device node exists but it cannot be opened for some weird reason).

tsroten:
@meeuw Yeah! I think that works. Thanks! That actually matches other parts of mycli: https://github.com/dbcli/mycli/blob/45f72cb218a69fae358dbc38f401506f34f9772a/mycli/main.py#L146:L149

Issue #644, opened 2018-09-04
`use` command should not open a new connection


laoshaw:
mycli will fail to import mysql's own sample database : https://dev.mysql.com/doc/sakila/en/sakila-installation.html . You can download the schema sql at: https://dev.mysql.com/doc/index-other.html

in the sakila-schema.sql, the first three lines are:
SET @OLD_UNIQUE_CHECKS=@@UNIQUE_CHECKS, UNIQUE_CHECKS=0;
SET @OLD_FOREIGN_KEY_CHECKS=@@FOREIGN_KEY_CHECKS, FOREIGN_KEY_CHECKS=0;
SET @OLD_SQL_MODE=@@SQL_MODE, SQL_MODE='TRADITIONAL';

looks like mycli somehow is not following them? mysql's native client has no problems though.

"mycli closes the connection and opens a new one after it does the USE sakila; line. So the session variables like FOREIGN_KEY_CHECKS are reset to the defaults for a new connection.

it's actually the USE sakila; line that causes the reconnect, rather than the CREATE SCHEMA. In                 main.py, you can see that register_special_commands() causes a "use" command to call change_db(), which just makes a new connection with sqlexecute.connect()"  - found out by @snoyes on #mysql irc

laoshaw:
"To fix it, I imagine they would break the part of sqlexecute.connect() that fetches information about a connection off into a separate method, and then register "use" to a function that calls                pymysql.select_db() and then the function that gets the connection information, rather than making a new connection."  -- from @snoyes

snoyes:
In main.py, register_special_commands() causes both 'USE' and 'connect' commands to call change_db(), which just calls sqlexecute.connect(), which opens a new connection using pymsql.connect(). In the case of 'USE', it should instead call pymsql.select_db().

tsroten:
Another issue related to the `use` command is the delay when using the new SSH tunnel feature.
It potentially takes a second or two to select a different default database. Ultimately, the `use` command needs to:
1.  Set the sqlexecute.dbname instance variable
2. Trigger the completions thread to run

amjith:
This is now available in v1.19.0. Please upgrade and give it a shot. 

```
pip install -U mycli
```

PR #679, opened 2018-10-26
No connection on use

AngeloLupo:
## Description

Exposed ```pymysql.selectdb``` through sqlexecute
Added error message on ```use``` with no parameters

## Checklist

### DIFF
--- a/mycli/AUTHORS
+++ b/mycli/AUTHORS
@@ -62,6 +62,7 @@ Contributors:
   * Huachao Mao
   * QiaoHou Peng
   * Yang Zou
+  * Angelo Lupo
 
 Creator:
 --------
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -217,11 +217,14 @@ class MyCli(object):
             yield (None, None, None, msg)
 
     def change_db(self, arg, **_):
-        if arg is None:
-            self.sqlexecute.change_db()
-        else:
-            self.sqlexecute.change_db(arg)
+        if arg is '':
+            click.secho(
+                "No database selected",
+                err=True, fg="red"
+            )
+            return
 
+        self.sqlexecute.change_db(arg)
         self.refresh_completions()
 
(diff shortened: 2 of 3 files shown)


meeuw:
Hey this is nice! This will fix prevent reconnects when you change databases.

There's only some issue that the list of databases will contain duplicate entries when you switch to another database using `use`.
```
mysql root@172.17.0.2:(none)> use mysql                                             
You are now connected to database "mysql" as user "root"
Time: 0.002s
mysql root@172.17.0.2:mysql> use  
                                  _test_db            
                                  _test_db            
                                  information_schema  
                                  information_schema  
                                  mysql               
                                  mysql               
                                  performance_schema                                
```

tsroten:
@AngeloLupo Thanks for your contribution! I'm just leaving a comment to note that this pull request is addressing issue #644.

AngeloLupo:
> Hey this is nice! This will prevent reconnects when you change databases.
> 
> There's only some issue that the list of databases will contain duplicate entries when you switch to another database using `use`.
> 
> ```
> mysql root@172.17.0.2:(none)> use mysql                                             
> You are now connected to database "mysql" as user "root"
> Time: 0.002s
> mysql root@172.17.0.2:mysql> use  
>                                   _test_db            
>                                   _test_db            
>                                   information_schema  
>                                   information_schema  
>                                   mysql               
>                                   mysql               
>                                   performance_schema                                
> ```

Good spot, that should be fixed now!

meeuw:
The current behavior of using `use` without any arguments is to set None as the current database. I'm not sure why this is useful but it seems to be implemented this way on purpose. @amjith do you remember why you did this and do you like to new behavior?

Could you please add your contribution to the changelog? Many thanks!

meeuw:
I just discussed this with @amjith and this feature (use without arguments) doesn't need to exist anymore.

Thanks again, I'm merging this PR!

PR #684, opened 2018-11-07
Require prompt_toolkit>=2.0.6

meeuw:
## Description
Require `prompt_toolkit>=2.0.6`

## Checklist


changelog.md
```diff
@@ -6,6 +6,7 @@ Internal:
 
 * Add Python 3.7 trove classifier (Thanks: [Thomas Roten]).
 * Fix pytest in Fedora mock (Thanks: [Dick Marinus]).
+* Require `prompt_toolkit>=2.0.6` (Thanks: [Dick Marinus]).
```
amjith: Looks good. Just one suggestion for the future. 

In the future, it'll be good to mention why we are doing this. For instance, this is to fix a workaround for the Vi keybindings bug in the older versions of prompt-toolkit library. 

The idea is to think of the changelog as a user facing document that helps them understand whether it is worth upgrading to this version or not.

PR #697, opened 2018-12-11
Fixes #669 -- Load favorites queries from and save to --myclirc

### LINKED ISSUE #669: Favorite queries not loaded from or saved to config file
When using a non-default config file (`--myclirc`) favorite queries are neither loaded from nor saved to the config files. Instead they are loaded from and saved to the default config file.

The same issue exists in pgcli and mssql-cli.

mdsrosa:
## Description

Favorite queries were being loaded from and save to the default configuration file (`~/.myclirc`) only. This PR fixes that issue by making sure that `FavoriteQueries` instance is created with the config loaded from `--myclirc` argument.

Fixes #669.

## Checklist

### DIFF
--- a/mycli/sqlcompleter.py
+++ b/mycli/sqlcompleter.py
@@ -8,7 +8,6 @@ from prompt_toolkit.completion import Completer, Completion
 
 from .packages.completion_engine import suggest_type
 from .packages.parseutils import last_word
-from .packages.special.favoritequeries import favoritequeries
 from .packages.filepaths import parse_path, complete_path, suggest_path
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


mycli/sqlcompleter.py
```diff
@@ -351,6 +350,7 @@ def get_completions(self, document, complete_event, smart_completion=None):
                                             fuzzy=False)
                 completions.extend(special)
             elif suggestion['type'] == 'favoritequery':
+                from .packages.special.iocommands import favoritequeries
```
amjith: Can you help me understand why this import statement is inside this `elif` clause instead of at the top of the file?

mdsrosa: @amjith Leftovers of testing/debugging. I've moved it back to the top of the file where it belongs. Thanks catching that! 🙂

amjith:
Works great! 

Nice work.

PR #705, opened 2019-01-04
myclirc: Mention escaping of special characters in DSN.

apapsch:
It is possible to use URL encoding for special characters,
because unquote from urlparse is used on parts of the DSN.

One can guess this from the exception message but a notice in the sample file is more helpful.

## Checklist

(This seems a bit much for a one line change that is a comment)


meeuw:
thanks for your contribution!

Issue #716, opened 2019-02-12
auto_vertical_output always on when syntax_style is set to native


tiagocorrea:
Apparently, `auto_vertical_output` is always occurring when `syntax_style` is set to native.

### Steps to reproduce:
1. Change `syntax_style` to native
2. Set `auto_vertical_output` to True
3. Open a connection
4. Run two queries: one with a result set shorter than your current window and another with a result set larger than your current window

### Expected:
`auto_vertical_output`, when True, is respected as intended. Short result sets are displayed using normal tabular output. Large ones are displayed using vertical output.

### Actual Behavior:
Every result set is displayed using a vertical output.

tsroten:
Hello @tiagocorrea! Thanks for filing this issue. I was not able to reproduce this on the latest version of mycli.

Can you check a few things?
1. In your mycli config file, what is `table_format` set to?
2. Is your terminal window being resized after you start mycli?
3. What terminal emulator application are you using? Are you running mycli inside of any programs (e.g. tmux, screen, emacs, etc.)
4. What version of mycli are you using?

tsroten:
@tiagocorrea Actually, I left my settings like you suggested and have noticed some weird behavior related to the `native` syntax style like you've seen.

It appears to be because the ASCII color codes in the native theme are making mycli think the lines are longer than they actually are when displayed.

tsroten:
@tiagocorrea Thanks for reporting this -- it should be fixed as of yesterday's release, version 1.20.0. You can upgrade if you installed via pip:
`pip install -U mycli`

PR #717, opened 2019-02-13
Strip ANSI escape codes for auto vertical output width calculations.

tsroten:
## Description

This fixes issue #716 by stripping ANSI escape codes from the first line of the output when calculating if the output can fit in the terminal window.

## Checklist


amjith:
:ice_cream:

PR #743, opened 2019-05-21
Bugfix/config values with comma

### LINKED ISSUE #624: object supporting the buffer API required
Hi, I have installed mysql@5.7 with homebrew. I use mysql_config_editor to set default [client] login password. When I type `mysql` on terminal it can login MySQL without problem, but when I type `mycli` it will show me `object supporting the buffer API required`. 

But, when I just use regular `mycli -uUser -pPassword` to login it has no problem.

macOS version: 10.13.6
MySQL version: 5.7.22 Homebrew
mycli version: 1.17.0 pip

Below is the log in `.mycli.log`.

```
2018-07-22 01:30:33,368 (48024/MainThread) mycli.main ERROR - traceback: 'Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/mycli/main.py", line 445, in connect
    _connect()
  File "/usr/local/lib/python3.7/site-packages/mycli/main.py", line 398, in _connect
    socket, charset, local_infile, ssl)
  File "/usr/local/lib/python3.7/site-packages/mycli/sqlexecute.py", line 52, in __init__
    self.connect()
  File "/usr/local/lib/python3.7/site-packages/mycli/sqlexecute.py", line 88, in connect
    conv=conv, ssl=ssl)
  File "/usr/local/lib/python3.7/site-packages/pymysql/__init__.py", line 94, in Connect
    return Connection(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/pymysql/connections.py", line 327, in __init__
    self.connect()
  File "/usr/local/lib/python3.7/site-packages/pymysql/connections.py", line 598, in connect
    self._request_authentication()
  File "/usr/local/lib/python3.7/site-packages/pymysql/connections.py", line 808, in _request_authentication
    authresp = _auth.scramble_native_password(self.password, self.salt)
  File "/usr/local/lib/python3.7/site-packages/pymysql/_auth.py", line 31, in scramble_native_password
    stage1 = sha1_new(password).digest()
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/hashlib.py", line 149, in __hash_new
    return _hashlib.new(name, data)
TypeError: object supporting the buffer API required
'
```

dhx-mike-palandra:
## Description

When a password is read from ~/.mylogin.cnf and contains a comma (,), mycli will terminate with error message  "object supporting the buffer API required".  Please note that the official mysql CLI client handles this case properly.

More detail:
1. In mycli/main.py (`MyCLI.connect()`): if keyword parameter `passwd=''` evaluates to false (it does by default), then a password will be parsed from config:
`cnf = self.read_my_cnf_files(self.cnf_files, cnf.keys())`
...
`passwd = passwd or cnf['password']`
2. In mycli/config.py (`read_config_file()`): a `configobj.ConfigObj()` object is initialized without explicit keyword parameter `list_values=True`, so its effective value will be `True`:
`config = ConfigObj(f, interpolation=False, encoding='utf8')`
This will cause any string value containing a "," to be parsed as a list which causes an exception somewhere (I didn't trace all the way through).

## Suggested Fix
Note that this is my first time looking at the source, so my advice may not be optimal.  Initially, I thought it would be more sensible to initialize `ConfigObj` with `list_values=False` but that just caused other problems since it preserves surrounding quotes whereas `list_values=True` strips them; see also:

[list_values](https://configobj.readthedocs.io/en/latest/configobj.html#list-values)
[ISSUES](https://configobj.readthedocs.io/en/latest/configobj.html#issues)

Instead, I have modified the nested `get()` function defined inside function `read_my_cnf_files()` in mycli/main.py.

This might close #624.

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -14,6 +14,7 @@ Bug Fixes:
 * Fix automatic vertical output with native syntax style (Thanks: [Thomas Roten]).
 * Update `cli_helpers` version, this will remove quotes from batch output like the official client (Thanks: [Dick Marinus])
 * Update `setup.py` to no longer require `sqlparse` to be less than 0.3.0 as that just came out and there are no notable changes. ([VVelox])
+* workaround for ConfigObj parsing strings containing "," as lists (Thanks: [Mike Palandra])
 
 Features:
 ---------
--- a/mycli/AUTHORS
+++ b/mycli/AUTHORS
@@ -65,6 +65,7 @@ Contributors:
   * Angelo Lupo
   * Aljosha Papsch
   * Zane C. Bowers-Hadley
+  * Mike Palandra
 
 Creator:
 --------
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -322,6 +322,11 @@ class MyCli(object):
             for sect in cnf:
                 if sect in sections and key in cnf[sect]:
                     result = cnf[sect][key]
+            # HACK: if result is a list, then ConfigObj() probably decoded from
+            # string by splitting on comma, so reconstruct string by joining on
+            # comma.
+            if isinstance(result, list):
+                result = ','.join(result)
             return result
 
         return {x: get(x) for x in keys}


tsroten:
@dhx-mike-palandra Thanks for this pull request!

Good thinking on trying `list_values = False` first. It's a little disappointing that isn't the solution in the end due to the quotation mark handling.

I think this looks good -- I don't know of any places where we are expecting lists from the .cnf files.

Issue #753, opened 2019-06-20
Piping in the REPL


Aerex:
I would like to preface this issue by saying this is an amazing tool. It is far better than the mysql cli. 

In regards to this ticket, this is not a bug but I couldn't find an area where I can ask questions. I would like to pipe the out to csv then run a command to convert this csv to json using this npm cli module [csvtojson](https://www.npmjs.com/package/csvtojson#command-line-usage). Aftewards I would pipe the response to `jq`

Here is an example. 
```
mycli > \o file
mycli > select * from tb1 where id = xxx
mycli > system csvtojson file | jq '.'

[
{"id":"1","status":"on","type":"normal"}
]
```

I expected the json response to be prettified like so 
```
[
  {
    "id": "1",
    "status": "on",
    "type": "normal"
  }
]
```

Is the behaviour from mycli expected or is there something I am missing here.

tsroten:
Hello @Aerex! That's a neat idea! Unfortunately, mycli does not currently support the use of pipes in the `system` command.

However, my recommendation is to use the "pager" for this. It eliminates the need to save the file into a file. First, explicitly enable the pager (by making sure the following is in your MySQL configuration file, `~/.my.cnf`):
```
[client]
pager=on
```

Then, you can use the pager command to extend mycli quite a bit. For example, using the `csvjson` utility from [csvkit](https://csvkit.readthedocs.io/en/latest/):
```
localhost:test_db> \T csv
Changed table format to csv
Time: 0.000s
localhost:test_db> pager csvjson | jq '.'
PAGER set to csvjson | jq '.'.
Time: 0.000s
localhost:test_db> select * from foo
[
  {
    "id": 1,
    "bar": "吃饭"
  },
  {
    "id": 2,
    "bar": "data"
  },
  {
    "id": 3,
    "bar": "hello world"
  },
  {
    "id": 4,
    "bar": "foo"
  }
]
```

Additional common pagers that I use are:
- `vim -`, opens the query output immediately in a Vim buffer for editing or further processing
- `slackcat -c CHANNEL`, sends the query output to the desired Slack channel as an attachment using [slackcat](https://github.com/bcicen/slackcat) (you might need to add the argument `--filetype=text`)

Once in mycli, you can toggle the pager on/off like this:
- `nopager`, pager is disabled, output is simply printed to the screen
- `pager COMMAND`, specify a command to use as the pager
- `pager`, the previously-used pager is turned back on (previously-used in the same session)

I hope that helps!

[...]

Issue #269, opened 2016-06-05
How can i use ssh Tunnel?


huihuidaxia:
RT?

j-bennet:
Something like to set up the tunnel:

```
ssh -Nf <ssh-user>@<ssh-host> -L 3306:<remote-mysql-host>:3306
```

Then connect to mysql as if it was running on localhost:

```
mycli -h localhost -P 3306
```

It's not specific to `mycli`, you'd use the same with `mysql`.

johnksterling:
I think this issue can be closed.

andrewimeson:
Support was also added in #634

PR #757, opened 2019-07-14
Feature | Block users from using shell (bash, sh, tcsh, etc) with SYSTEM query

lfraga:
## Description
When users have setup with MyCLI and Linux user custom shell (echo '/opt/mycli-script.sh' | chsh username) for access database directly with SSH and MyCLI, shell access is expected to be blocked

Added boolean parameter **disable_system_shell** for block SQL queries like "SYSTEM id" or "SYSTEM \<shell commands\>"

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,7 @@ Features:
 * Auto find alias dsn when `://` not in `database` (Thanks: [QiaoHou Peng]).
 * Mention URL encoding as escaping technique for special characters in connection DSN (Thanks: [Aljosha Papsch]).
 * Pressing Alt-Enter will introduce a line break. This is a way to break up the query into multiple lines without switching to multi-line mode. (Thanks: [Amjith Ramanujam]).
+* Disable shell commands (bash, sh, tcsh, etc) with SYSTEM query (Thanks: [Lucas Fraga]).
 
 Bug Fixes:
 ----------
--- a/mycli/AUTHORS
+++ b/mycli/AUTHORS
@@ -66,6 +66,7 @@ Contributors:
   * Aljosha Papsch
   * Zane C. Bowers-Hadley
   * Mike Palandra
+  * Lucas Fraga
 
 Creator:
 --------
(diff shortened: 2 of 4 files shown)


tsroten:
Hello @lfraga! Thanks for the pull request! We appreciate the contribution 😄 

So, I guess I'm wondering if mycli is the best solution for a use case like this. There are other ways outside of the `system` command for mycli to affect the host machine. Removing these features would severely limit mycli users. But, without addressing them, this change wouldn't be very helpful. Four examples:
- `pager ls -la` (runs `ls -la` whenever output is sent to the pager).
- You can use `\e` to open an editor like vim that allows you to run system commands.
- You can use `\e` to save a Bash script to the host machine, then use `pager` to run it.
- Save a shell script from the query output: `\T csv; \o ~/ls.sh; select "$(ls)";`. Then, run it via `pager sh ~/ls.sh`.

One of the powerful uses for the `pager` command is how it can be used to run arbitrary commands (and even pipe output from one to another): https://github.com/dbcli/mycli/issues/753#issuecomment-504279817

What do you think?

tsroten:
@lfraga If I may propose a solution that might achieve what you want, without limiting a users ability to use mycli:

- Set up access on the SSH server with no shell at all (e.g. `/bin/false`)
- Have the user run mycli on their own machine (where they are free to run whatever commands they want).
- They can connect to the database via an SSH tunnel.



Some additional information about this approach:

SSH tunnels with no shell access: https://serverfault.com/questions/56566/ssh-tunneling-only-access  

Using `ssh` to create a tunnel with mycli: https://github.com/dbcli/mycli/issues/269

Using mycli's ssh tunnel feature, `mycli --help | grep ssh`:
```
--ssh-user TEXT               User name to connect to ssh server.
--ssh-host TEXT               Host name to connect to ssh server.
--ssh-port INTEGER            Port to connect to ssh server.
--ssh-password TEXT           Password to connect to ssh server.
--ssh-key-filename TEXT       Private key filename (identify file) for the
                              ssh connection.
```

amjith:
Closing the PR since there is no activity for a while now.

Issue #744, opened 2019-05-23
more accurate parsing of mysql configuration


dhx-mike-palandra:
Although my [pull request](https://github.com/dbcli/mycli/pull/743) does resolve issues with my particular configuration, it is not a general solution.  My original intent was to provide a general solution but I acted in haste; sincerest apologies.  At least it's a very localized change that's easy to undo. I would take no offense whatsoever if my pull request was reverted and my name thus removed from the contributor list.

To demonstrate:
```
from configobj import ConfigObj

value = "a, b,       c"
lines = ["[client]", f"key = {value}"]
parsed = ",".join(ConfigObj(lines, list_values=True)["client"]["key"])
print(f" value = {value}")
print(f"parsed = {parsed}")
print(f" same? = {parsed == value}")
```

i.e. Intermediate blanks are stripped, so the mapping between parsed value (a list) and raw value (a string) is NOT invertible.  I should have been more careful and perhaps pursued further my original intuition about initializing `ConfigObj` with `list_values=False`.  I will try to do so when time permits.

In case someone else beats me to that, I'd like to offer a few tips:

1. Be aware of this [documented issue](https://configobj.readthedocs.io/en/latest/configobj.html#issues) with `list_values=False`:
> ConfigObj doesn’t quote and unquote values if list_values=False. This means that leading or > trailing whitespace in values will be lost when writing. (Unless you manually quote).

With `list_values=False`, note that several values from `myclirc` currently are quoted, so those quotes would have to be removed somehow; e.g. `colors` section.  More tricky would be something like the `prompt` key from `main` section which quotes a trailing blank.

2. For configuration files written by official mysql client, the goal should be to implement strictly compatible parsing behavior.  Assuming configobj must remain, then one must test how the official mysql client parses at least the following:
    * Values enclosed in matching quotes (double and single)
    * Unquoted values surrounded by blanks

3. It might be better to have two different config parser implementations (or perhaps one implementation with two distinct configurations); i.e. one (read-only) for mysql configuration, and another (read/write) for mycli configuration (toml might be worth considering).

tsroten:
@dhx-mike-palandra Thanks for the detailed follow-up to your pull request. I think your pull request did just fine :thumbsup: and shouldn't have a negative side effect on the writing of configuration files.

As far as I remember, the code in your pull request was in a method that would only have affected reading. We don't use the method `read_my_cnf_files()` to read from the myclirc (which is the file we write to). We only use it to read from the following files:
[...]

PR #742, opened 2019-05-18
Sqlparse 0.3.0

meeuw:
## Description
Fixed unit tests for #737 

## Checklist

- [x] I've added this contribution to the `changelog.md`.
- [x] I've added my name to the `AUTHORS` file (or it's already there).

PR #764, opened 2019-08-10
Do not use ConfigObj's list_values.

tsroten:
## Description

This is a follow-up to #742, #744, and #624.

This disables ConfigObj's `list_values` since it converts strings with commas in them into lists. This makes it impossible to have a password that uses a comma.

Instead, this approach turns `list_values` off when reading MySQL config files. By doing so, ConfigObj will not remove surrounding quotes, so we have to do that ourselves.

This allows a user to have a password with a comma. If a user has a password that starts or ends with a quote, they can simply surround their password in the config file with quotes.

## Checklist

- [X] I've added this contribution to the `changelog.md`. (it's still there from Mike's PR)

### DIFF
--- a/mycli/config.py
+++ b/mycli/config.py
@@ -37,6 +37,7 @@ def read_config_file(f, list_values=True):
     (e.g. 'a,b,c' -> ['a', 'b', 'c']. Additionally, the config values are
     not unquoted. We are disabling list_values when reading MySQL config files
     so we can correctly interpret commas in passwords.
+
     """
 
     if isinstance(f, basestring):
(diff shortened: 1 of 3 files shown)


meeuw:
Nice fix Thomas! No comments on the implementation only that is wasn't very clear why it needs to be fixed this way. I'd suggest to add some comments that:
- (from https://configobj.readthedocs.io/en/latest/configobj.html#list-values) If set to False then values are not parsed for list values. !! In addition single line values are not unquoted. !!
- you've copied the unquote implementation from configobj (maybe we should ask upstream to fix this behavior / make the unquote function public)
- list_values is only False for my.cnf and friends and True for myclirc and friends.

tsroten:
@amjith @meeuw Thanks guys! I'd like to add some comments before this is merged (per Dick's feedback) 👍.

tsroten:
@meeuw Comments added. Thanks!

meeuw:
🎈

PR #699, opened 2018-12-13
feature(connection): auto to find dsn not -d

vastpeng:
## Description

related issue: #698 

## Checklist

- [x] I've added this contribution to the `changelog.md`.
- [x] I've added my name to the `AUTHORS` file (or it's already there).

PR #769, opened 2019-08-20
Fix login path with command-line database name.

### LINKED ISSUE #768: "Invalid DSNs found in the config file." error when using login-path option with DB name
The follow command worked in mycli 1.19.0:

`mycli --login-path=<db_alias> --auto-vertical-output <db_name>`

It no longer works in 1.20.0, with the following error:

`Invalid DSNs found in the config file. Please check the "[alias_dsn]" section in myclirc.`.

The example in the documentation (https://www.mycli.net/loginpath) also no longer works, raising the same error as above:

`mycli --login-path my_server_alias my_database_name`.

These commands work:

`mycli --login-path=<db_alias>`
`mycli --login-path=<db_alias> --auto-vertical-output`

but tacking on the database name at the end raises the error.

System info:

macOS 10.14.6
zsh 5.7.1
MariaDB 10.2.21-MariaDB-log

tsroten:
## Description

This fixes #768. 

The DSN code was checking for login credentials and if they didn't exist, it interpreted the database argument as a DSN alias (a [super handy feature](https://github.com/dbcli/mycli/pull/699)!). But, it didn't recognize the login path as part of the credentials. So, the database name was being interpreted as a DSN alias, when it shouldn't have been.

This pull request adds in a check for login path. It also only treats the database argument as the DSN alias when the user's config file defines DSN aliases.

Additionally, I reworded the error message to provide more specific information (not found vs. invalid).

I've included a version bump in the changelog with this PR so we can release this fix after it's merged.

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,11 @@
+1.20.1
+======
+
+Bug Fixes:
+----------
+
+* Fix an error when using login paths with an explicit database name (Thanks: [Thomas Roten]).
+
 1.20.0
 ======
 
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -1085,21 +1085,22 @@ def cli(database, user, host, port, socket, password, dbname,
 
     dsn_uri = None
 
-    if database and '://' not in database and not any([user, password, host, port]):
-        dsn = database
-        database = ''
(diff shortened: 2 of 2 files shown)


amjith:
:beach_umbrella:

Issue #776, opened 2019-09-30
Completions don't reload after table rename


gfrlv:
If I rename a table and then make another query with the same table, the old name shows up in the completion menu

amjith:
Fixed!

PR #775, opened 2019-09-30
Dsn alias prompt

pasenor:
## Description
Added a format specifier `\A` to include the DSN alias name into the prompt

## Checklist

- [x] I've added this contribution to the `changelog.md`.
- [x] I've added my name to the `AUTHORS` file (or it's already there).

PR #778, opened 2019-09-30
added RENAME TABLE to the lists of mutating and destructive queries

pasenor:
## Description
Added checks that `RENAME TABLE` is a destructive and mutating query, see #776 

Also, it seems we can spare the check in `is_mutating` if we already know that the query is destructive, so I added a shortcut in `one_iteration()`.

## Checklist
(done in #775 )

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -631,7 +631,7 @@ class MyCli(object):
 
                     start = time()
                     result_count += 1
-                    mutating = mutating or is_mutating(status)
+                    mutating = mutating or destroy or is_mutating(status)
                 special.unset_once_if_written()
             except EOFError as e:
                 raise e
@@ -1204,7 +1204,7 @@ def need_completion_refresh(queries):
         try:
             first_token = query.split()[0]
             if first_token.lower() in ('alter', 'create', 'use', '\\r',
-                    '\\u', 'connect', 'drop'):
+                                       '\\u', 'connect', 'drop', 'rename'):
                 return True
         except Exception:
             return False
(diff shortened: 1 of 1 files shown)


amjith:
When we mark a query as destructive `mycli` will warn you before you execute the query. 

I don't think you want to make rename as a destructive query I think you just want to trigger a refresh of completions. Let me see how we do that in other places and I'll add a comment.

mycli/packages/parseutils.py
```diff
@@ -205,7 +205,7 @@ def queries_start_with(queries, prefixes):
 
 def is_destructive(queries):
     """Returns if any of the queries in *queries* is destructive."""
-    keywords = ('drop', 'shutdown', 'delete', 'truncate', 'alter')
+    keywords = ('drop', 'shutdown', 'delete', 'truncate', 'alter', 'rename')
```
amjith: I wouldn't mark it as destructive. Otherwise this PR looks good.

amjith:
Looks like you're doing the right thing by adding it to the list of commands that need refreshing. 

But I wouldn't add it to the list of destructive queries.

pasenor:
Indeed, I just checked the default mysql shell, there is no warning before `RENAME TABLE` there too. Seems a bit counter-intuitive, since we do warn before renaming a column, but I removed it from the destructive list.

amjith:
>  we do warn before renaming a column

This has been contentious. We have received complaints about warning on ALTER commands. 

We might do something about it in the future. But for now we're leaving the behavior alone.

Issue #383, opened 2017-03-24
Add ability to set statement delimiter


jpitchardu:
Hi, I've trouble using mycli to create stored procedures but when I try `delimiter //;` I get 

> (1064, u"You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'delimiter //' at line 1")

But, if I run it on mysql default client it executes as supposed.

tsroten:
@pichardoJ Thanks for using mycli! I'm sorry you're having issues defining stored procedures.

mycli does not currently support the `delimiter` command -- this is not a SQL statement, but a MySQL client command.

I think it's a good idea. If you or anyone else wants to try to implement it, feel free to :)

tsroten:
Just a note that the sqlparse library we use internally does not currently support delimiters other than semi-colon. They have an ongoing issue tracking this request: https://github.com/andialbrecht/sqlparse/issues/139

tsroten:
A helpful sqlparse user posted a possible workaround for this that we could experiment with in mycli: https://github.com/andialbrecht/sqlparse/issues/139#issuecomment-518659069

pasenor:
There is a neat workaround since version 1.20.0: you can switch to single-line mode and break lines with `Alt-Enter`. It allows you to create multi-line stored procedures without changing delimiters:

![изображение](https://user-images.githubusercontent.com/11982173/66504585-1b167880-ead2-11e9-9c6b-778052094d53.png)

pasenor:
Released in version 1.21.0

PR #782, opened 2019-10-05
implemented DELIMITER command

pasenor:
## Description
#383 
DELIMITER command to switch the default delimiter `;` to something else. Copied the behavior from the standard mysql shell. 

Splitting lines is based on a [suggestion](https://gist.github.com/srittau/01282d7e306ae5f73c601752f2503fa6) by @srittau as a temporary workaround until custom delimiters get into the sqlparse library

One limitation is that delimiter cannot be used in a favorite query, for example if you save

    \fs foo delimiter $ select 1 $ select 2 delimiter ;

then `\f foo` will result in a syntax error, although a user may actually want to save such a query, e.g. to do some work on stored procedures. I can imagine refactoring favorite queries so that they are executed with `SQLExecute`, then all special commands will be available, but I'm not sure if it's worth the trouble.
UPD: ah, I see Thomas has been doing some work towards special commands is favorite queries in #600 so it's probably better to handle there.

Another concern is that I have changed the API of  `_multiline_exception` (it now accepts the delimiter argument), which is one more difference from pgcli and the rest of the dbcli tools.

And generally, though having delimiter switching may be useful to people accustomed to the default mysql shell, it's primary purpose is to deal with stored procedures. Maybe it would be more straightforward to implement a simple call stack, so we can track BEGIN - END statements directly? 

## Checklist


meeuw:
1. I think DELIMITER should be implemented in a class in mycli.packages.special and the delimiter should be stored there instead of passing it around though a lot of functions
1. What is the actual use case for having a delimiter other than `;` ?

Nice work though, it seems pretty thorough!

amjith:
> What is the actual use case for having a delimiter other than ; ?

The delimiter is changed from the default (;) when you want to create functions that have semi-colons in them. 

This stackoverflow answer does a better job of explaining it. 

https://stackoverflow.com/a/10259528

amjith:
> I think DELIMITER should be implemented in a class in mycli.packages.special and the delimiter should be stored there instead of passing it around though a lot of functions

@pasenor Do you want to take a stab at implementing it as part of the special commands package? 

When I try out the feature I see `DELIMITER` listed twice in the completion menu. This is because it is listed as a keyword in sqlcompleter.py and also as a special command. You can remove the entry from the keyword in sqlcompleter.py file.

pasenor:
>     1. I think DELIMITER should be implemented in a class in mycli.packages.special and the delimiter should be stored there instead of passing it around though a lot of functions

Makes sense, but we still need to somehow let the `is_multiline` function know about the new delimiter. I'll look into it, probably swapping the `prompt_app.multiline` could be an option.

pasenor:
> What is the actual use case for having a delimiter other than ; ?

As @amjith said, the only use case is declaring stored procedures with `;` in them.
In fact, I feel it's an ugly solution, and what we need instead is a proper scoping based on `BEGIN ... END`, like what `psql` does.  But a lot of scripts tailored for mysql  already use `DELIMITER` (I too have a couple of them), and it would be nice if mycli could run them.

meeuw:
mmm, I never used stored procedures in mysql, sounds like a valid use case...

amjith:
This works great! I like how thorough the tests are done for this feature. :1st_place_medal:

meeuw:
nicely done @pasenor !

Issue #758, opened 2019-07-16
Get issue when delete from table


WeitaoYan:
Hi,
It looks that I get a issue when delete from delete table.

My version is mycli 1.19.0, OS is windows, installed by pip. It is a accident, that I type 't' when confirm, and then data been deleted...
automation_test_platform> delete from test_result_tput;
You're about to run a destructive command.
Do you want to proceed? (y/n): t
Your call!
Query OK, 76 rows affected
Time: 0.063s

tsroten:
Hello @WeitaoYan! So, it looks like the library we use that provides the prompt considers the following as ["true" values](https://github.com/pallets/click/blob/2c622eed402c64e1645db3904c2410f6c05fbf19/click/types.py#L351):
`('true', 't', '1', 'yes', 'y')`

Since deleting data is not something we want to happen on accident, this should definitely be fixed in mycli. Thanks for finding this!

PR #787, opened 2019-10-13
Restrict destructive command confirmation to ('yes', 'y')

pasenor:
## Description
Issue #758 
The default boolean converter in Click evaluates "t", "true" and "1" to `True`. Added a custom one.

## Checklist

### DIFF
--- a/mycli/packages/prompt_utils.py
+++ b/mycli/packages/prompt_utils.py
@@ -7,6 +7,26 @@ import click
 from .parseutils import is_destructive
 
 
+class ConfirmBoolParamType(click.ParamType):
+    name = 'confirmation'
+
+    def convert(self, value, param, ctx):
+        if isinstance(value, bool):
+            return bool(value)
+        value = value.lower()
+        if value in ('yes', 'y'):
+            return True
+        elif value in ('no', 'n'):
+            return False
+        self.fail('%s is not a valid boolean' % value, param, ctx)
+
+    def __repr__(self):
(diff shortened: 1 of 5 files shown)


amjith:
Should we allow 't', 'true' and '1' to be truthy responses? Somehow I feel like it is not the right approach. I'd much rather have just 'yes' and 'y' as truthy and everything else as negative.

pasenor:
Yeah, but that's exactly what's happening in this PR :)
The default behavior of Click is to allow t and 1, I'm restricting it to only 'y' and 'yes'

mycli/packages/prompt_utils.py
```diff
@@ -7,6 +7,26 @@
 from .parseutils import is_destructive
 
 
+class ConfirmBoolParamType(click.ParamType):
+    name = 'confirmation'
+
+    def convert(self, value, param, ctx):
+        if isinstance(value, bool):
+            return bool(value)
```
amjith: When would we get a boolean value?

amjith:
Looks good to me. :man_juggling:

mycli/packages/prompt_utils.py
pasenor: hmm, probably never. I copied it from [here](https://github.com/pallets/click/blob/2c622eed402c64e1645db3904c2410f6c05fbf19/click/types.py#L348) without much thought, maybe we can delete this conditional branch.

PR #792, opened 2019-10-28
Added clearer error message when failing to connect to default socket

### LINKED ISSUE #572: Better error message when unable to connect via socket
Currently, on a Linux system after installing MySQL and mycli, there is a cryptic message displayed when running `mycli` with the default config:
```bash
$ mycli
(2003, "Can't connect to MySQL server on 'localhost' ([Errno 2] No such file or directory)")
Failed to connect by socket, retrying over TCP/IP
```

The issue itself will be resolved by https://github.com/dbcli/mycli/issues/571, but the error message is unhelpful because it doesn't tell you anything about the socket that was tried. It just tells you the host name, which makes you think there is a problem with your server. Compare that to the official mysql client's message if you connect to a non-existent socket:
```bash
$ mysql -S /var/run/mysqld/mysqld.sock
ERROR 2002 (HY000): Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)
```

jonathanlloyd:
Fixes #572

## Description
Added a reference to the default socket path when failing to connect, should help when trying to debug why the connection is failing

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,8 +5,8 @@ Features:
 ---------
 * Added DSN alias name as a format specifier to the prompt (Thanks: [Georgy Frolov]).
 * Mark `update` without `where`-clause as destructive query (Thanks: [Klaus Wünschel]).
-
 * Added DELIMITER command (Thanks: [Georgy Frolov])
+* Added clearer error message when failing to connect to the default socket.
 
 
 1.20.1
--- a/mycli/AUTHORS
+++ b/mycli/AUTHORS
@@ -67,6 +67,7 @@ Contributors:
   * Zane C. Bowers-Hadley
   * Mike Palandra
   * Georgy Frolov
+  * Jonathan Lloyd
 
 Creator:
 --------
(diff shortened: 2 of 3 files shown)


pasenor:
Thanks! The build is failing because of the linter: you want to run `python setup.py lint` and then `python setup.py lint --fix` before submitting. Also, it seems we had no test for this error message, so now is probably a good time to add one.

jonathanlloyd:
> Thanks! The build is failing because of the linter: you want to run `python setup.py lint` and then `python setup.py lint --fix` before submitting. Also, it seems we had no test for this error message, so now is probably a good time to add one (this is up to you, we can do it here or in a separate PR)

I'll take a look at adding the test, shouldn't be too hard with the pexpect stuff?

jonathanlloyd:
Hmm, having poked around on this for a bit I can't think of an easy way to test this - it'd have to be done as a black-box test and, by definition, would have to be run in an environment that didn't have MySQL running on any default socket/host:port.
Options are:
 - Run MySQL on a non-standard port for all tests
 - Run this specific test in a separate travis job where MySQL is not installed
 - Merge this PR as is

Thoughts @pasenor?

pasenor:
Agree, it's a little awkward right now. I think we need to get rid of the hardcoded default socket path: it works on Debian, but some distributions have different default locations. If the path was configurable (per `setup.py`?), then we could also mess with it in tests.
Anyway, it should be a different issue. Merging this PR.

Issue #794, opened 2019-11-03
Make the default socket path distribution-specific


pasenor:
When the socket path is not specified in `my.cnf`, we try [the hardcoded value](https://github.com/dbcli/mycli/blob/1480966c372ff96e2bd108aca3d0651747dc7778/mycli/main.py#L432) `/var/run/mysqld/mysqld.sock`. However, this is not a universal constant: [the docs](https://dev.mysql.com/doc/refman/8.0/en/problems-with-mysql-sock.html) specify `/tmp/mysql.sock` as the default location, and in different distribution packages of mysql it is variously overriden. Here is [an old blog post](https://ocelot.ca/blog/blog/2015/07/19/connecting-to-mysql-or-mariadb-with-sockets-on-linux/) where someone collected the following default paths:

|OS  | socket location |
|---|---|
| Non-Linux e.g. FreeBSD or Solaris | /tmp/mysql.sock |
| Debian-based e.g. Ubuntu, and archlinux | /var/run/mysqld/mysqld.sock |
| SUSE (after v11.2) |  /var/run/mysql/mysql.sock |
| Red Hat, and SUSE (before v11.2) | /var/lib/mysql/mysql.sock |
| archlinux (very old versions) | /tmp/mysqld.sock |

I suggest putting the default socket location into the `.myclirc` file. In theory, it would then be a package maintainer's job to specify the correct path for each distribution, but since many distribution packages are not maintained, we could probably do some guesswork in `setup.py`

PR #795, opened 2019-11-04
try more default socket paths

pasenor:
## Description
Issue #794 
If the socket path is not specified, try to find a file by the name mask `mysql*.socket` in different locations. 

I wonder if we should move `DEFAULT_SOCKET_DIRS` from code to `.myclirc`, so it can be configured for different distributions.

WIP: test and collect more data on where the default socket may be placed in different OS.

## Checklist


amjith:
Is this ready for a merge?

pasenor:
Not quite: it works on the distributions from [the table](https://github.com/dbcli/mycli/issues/794#issue-516884191), but I still have not gotten to inspecting other distributions, and, more importantly, other DBs like MariaDB (they may name sockets differently).

By the way, does anyone have access to a MacOS system? What is the MySQL socket path there?

mycli/packages/filepaths.py
```diff
@@ -3,11 +3,13 @@
 from mycli.encodingutils import text_type
 import os
 
+DEFAULT_SOCKET_DIRS = ('/var/run/', '/var/lib/', '/tmp')
```
meeuw: `/tmp` might not be such a good idea; on a multi-user system anyone might create a socket in this directory to intercept passwords...

meeuw:
I like this change because it will help a common problem experienced by our users.

If we support multiple socket locations we should inform the user which location was chosen, especially when there are actually multiple socket locations available.

j-bennet:
@pasenor I'm on macOs and it is `/tmp/mysql.sock`:

```
$ mysql_config --socket
/tmp/mysql.sock
```

pasenor:
OK then, let's add `/tmp` only for MacOS, to make justice to @meeuw 's security concern. After all, a mac is typically a personal system.

meeuw:
> OK then, let's add `/tmp` only for MacOS, to do justice to @meeuw 's security concern. After all, a mac is typically a personal system.

That would be fine for me, another posibility would be to check the owner of the socket (root? mysql?)

amjith:
I'm seeing an error when I launch mycli with no arguments or options. 


Stacktrace from the log file: 

```
'Traceback (most recent call last):\n  File
"/home/amjith/Dropbox/code/python/mycli/mycli/main.py", line 450, in connect\n
_connect()\n  File "/home/amjith/Dropbox/code/python/mycli/mycli/main.py", line
419, in _connect\n    ssh_password, ssh_key_filename\n  File
"/home/amjith/Dropbox/code/python/mycli/mycli/sqlexecute.py", line 62, in
__init__\n    self.connect()\n  File
"/home/amjith/Dropbox/code/python/mycli/mycli/sqlexecute.py", line 118, in
connect\n    defer_connect=defer_connect\n  File
"/home/amjith/.virtualenvs/mycli/lib/python3.7/site-packages/pymysql/__init__.py",
line 94, in Connect\n    return Connection(*args, **kwargs)\n  File
"/home/amjith/.virtualenvs/mycli/lib/python3.7/site-packages/pymysql/connections.py",
line 325, in __init__\n    self.connect()\n  File
"/home/amjith/.virtualenvs/mycli/lib/python3.7/site-packages/pymysql/connections.py",
line 598, in connect\n    self._get_server_information()\n  File
"/home/amjith/.virtualenvs/mycli/lib/python3.7/site-packages/pymysql/connections.py",
line 985, in _get_server_information\n    self.server_thread_id =
struct.unpack(\'<I\', data[i:i+4])\nstruct.error: unpack requires a buffer of 4
bytes\n'
```

pasenor:
My guess would be that we found a socket (a wrong one), connected to it, but now it is sending garbage. Is it likely to happen in your setup (i.e. do you have multiple socket candidates)? I'll try to reproduce the situation and push a workaround. Also, we only echo the socket location after a successful connect, probably better to do it earlier.

pasenor:
i looked into it again and found another problem: while reading the `.cnf` files we did follow the `!includedirs` directives. If the socket is defined in one of those included files, it was not found and mycli had to guess.

With this fixed, the guessing function should only be called as a last resort in case our user is not allowed to read configs.

@amjith can you check if your error still happens? I have not been able to reproduce it, but If the cause was mycli finding the wrong socket it may be fixed now.

amjith:
Works well. Thank you!

:racing_car:

Issue #756, opened 2019-07-10
USE does not change db if used after DELETE and CREATE in single line


danadam:
I use this sequence quite a lot:

    DROP DATABASE foobar; CREATE DATABASE foobar; USE foobar;

It works in `mysql` client, i.e., I end up with `foobar` database selected. In `mycli` it says that it connects to it but in the end the database is not selected and I have to execute `USE` again on its own:

    mariadb root@localhost:foobar> DROP DATABASE foobar; CREATE DATABASE foobar; USE foobar;
    You're about to run a destructive command.
    Do you want to proceed? (y/n): y
    Your call!
    Query OK, 0 rows affected
    Time: 0.001s

    Query OK, 1 row affected
    Time: 0.000s

    You are now connected to database "foobar" as user "root"
    Time: 0.000s
    mariadb root@localhost:(none)>
    mariadb root@localhost:(none)> USE foobar
    You are now connected to database "foobar" as user "root"
    Time: 0.000s
    mariadb root@localhost:foobar>

tsroten:
@danadam Thanks for using mycli! And, thanks for finding this bug.

My hunch is this is a bug introduced by #490.

SukiCZ:
Hi @tsroten :tada: 
I love using `mycli` and this seems like a good way to say thanks :two_hearts: 
You can find PR fixing this issue #796 :computer: 

Let me know what you think :thinking: 

Thanks! :beers:

PR #796, opened 2019-11-06
Fix/use after create delete

SukiCZ:
## Description
Is not evaluating `is_dropping_database` as `True` if `CREATE` statement follows for same db

## Checklist

- [x] I've added my name to the `AUTHORS` file (or it's already there). ( :two_hearts: )


amjith:
Hi @SukiCZ. Thank you for the PR. Can you give me a little bit more context around why this change is necessary? Is there a use case that you encountered which resulted this change? 

The idea behind this check is to ensure that we reinitialize the connection and refresh our completions when we encounter a DROP statement because that is a DDL statement that requires a completion refresh. With the change you're suggesting it could invalidate that refresh trigger. 

So I'm trying to understand if there is a specific problem you're trying to solve with this change.

pasenor:
Ah, @amjith I see what you mean. @SukiCZ here is a query that would break everything after your change:

    DROP DATABASE foobar; CREATE DATABASE foobar; DROP DATABASE foobar;

SukiCZ:
Hi @pasenor , @amjith :wave:
Thanks for your review / time :timer_clock:
Not that I have any personal use-case for this but I've noticed opened issue referencing this. I thought it might be a easy way to contribute. (Late HacktoberFest :smile: )
Maybe I was too rush thinking about this - maybe we could extend `is_dropping_database` to enforce re-initialization of new connection, yet keep using the same one - or maybe to open discussion on this topic.

`is_dropping_database('DROP DATABASE foo; CREATE DATABASE foo; DROP DATABASE foo', 'foo')` is evaluated as `True`.

What's your thoughts? :thinking: 
Cheers :+1:

amjith:
Do you have a reference to the issue that you were trying to address with this PR? We can perhaps find an alternative solution.

SukiCZ:
sure. should be #756

amjith:
@tsroten Would you mind reviewing this PR? 

I know you had some experience with a bug fix related to drop statement.

tsroten:
@SukiCZ Hi! Thanks for this contribution 😄 I'm sorry it's taken me so long to look at it. It looks great to me. I'm going to resolve the conflicts and then merge it.

Issue #798, opened 2019-12-03
changing output file prevents output


peterdutton:
Hi,

Firstly, thanks for a fantastic piece of software.

My issue is that if I direct output to a file, e.g:

`\o ~/output/file1.sql`

Then run a query (which outputs as normal), if I then direct to a different file using the same `\o` option, there is no output to that file, although there is no error in mycli itself. The only way to get the output working again is to `exit` and relaunch mycli.

Thanks for any advice you might have.

pasenor:
Thanks for pointing it out, there is indeed a bug

PR #799, opened 2019-12-03
Allow multiple invocations of \o

pasenor:
#798 
We can only use `\o` once per session.
Specifically, the global variable `written_to_once_file` is set to True after the first call of `write_once`, so that every time `once_file` is set afterwards it is immediately unset in `unset_once_if_written()`

## Checklist

- [ x] I've added my name to the `AUTHORS` file (or it's already there).


changelog.md
```diff
@@ -16,6 +16,7 @@ Bug Fixes:
 ----------
 
 * Fix an error when using login paths with an explicit database name (Thanks: [Thomas Roten]).
+* The \o command could only be used once per session (Thanks: [Georgy Frolov])
```
amjith: This PR is doing the opposite of allowing \o only once in a session, right? 

Shouldn't this say: "Allow \o command more than once per session".

pasenor: Agree, clumsy wording. I meant to say "Fixed a bug where it could only be used...".
Besides, this line was in the wrong section :)

Issue #797, opened 2019-11-29
Startup error


MrRedstoner:
I just installed MyCLI on a CentOS 7 machine, using pip.

Itermittently I get
```
Exception in thread completion_refresh:
Traceback (most recent call last):
  File "/usr/lib64/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/usr/lib64/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/lib/python3.6/site-packages/mycli/completion_refresher.py", line 77, in _bg_refresh
    callback(completer)                                                                                                                                                 
  File "/usr/local/lib/python3.6/site-packages/mycli/main.py", line 865, in _on_completions_refreshed
    self.prompt_app.app.invalidate()
  File "/usr/local/lib/python3.6/site-packages/prompt_toolkit/application/application.py", line 468, in invalidate
    schedule_redraw()
  File "/usr/local/lib/python3.6/site-packages/prompt_toolkit/application/application.py", line 451, in schedule_redraw
    redraw, max_postpone_time=self.max_render_postpone_time, loop=self.loop
  File "/usr/local/lib/python3.6/site-packages/prompt_toolkit/eventloop/utils.py", line 57, in call_soon_threadsafe
    loop2 = loop or get_event_loop()
  File "/usr/lib64/python3.6/asyncio/events.py", line 694, in get_event_loop
    return get_event_loop_policy().get_event_loop()
  File "/usr/lib64/python3.6/asyncio/events.py", line 602, in get_event_loop
    % threading.current_thread().name)
RuntimeError: There is no current event loop in thread 'completion_refresh'.
```
on startup. I can end the session using Ctrl+D, the `Goodbye!` at the end does show up.

It seems to happen <50% of the time so far.

I did some changes before the issue first happened, I set up MySQL stored credentials using mysql_config_editor and changed the .myclirc to default to vi mode. I'm starting mycli with -h localhost, otherwise it first tries and fails to connect by socket (2003, "Can't connect ...").

Versions printed after succesfull startup:
MySQL 8.0.16
MyCLI 1.20.1

tt67wq:
+1 in python 3.7

pasenor:
[...]

PR #800, opened 2019-12-03
Restrict prompt_toolkit version to <3.0.0

pasenor:
prompt_toolkit==3.0.2 is pulled during install on python >= 3.6, but it is not yet supported
Some errors are reported here: #797 
Some others can be seen in [this test run](https://travis-ci.org/dbcli/mycli/jobs/620050433?utm_medium=notification&utm_source=github_status)


carlwgeorge:
I'm the maintainer of python-prompt-toolkit in Fedora.  At some point I need to update it to version 3 in order to support other applications that require it.  Are there plans for mycli to support prompt-toolkit 3?

PR #804, opened 2019-12-07
fixed version of colorama (pep8radius dependency)

pasenor:
## Description
pep8radius just stopped building on python 3.4 after the release of [colorama 0.4.3](https://pypi.org/project/colorama/#history). 
I suggest just fixing the old version, since we don't do anything fancy with it anyway.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -8,5 +8,6 @@ pexpect
 coverage==4.3.4
 codecov==2.0.9
 autopep8==1.3.3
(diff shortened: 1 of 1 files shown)


pasenor:
Or should we better drop pep8radius and adopt the black-based procedure from pgcli?

amjith:
I think we should just use black for everything.

amjith:
Until we adopt black we can continue to use pep8radius.

PR #805, opened 2019-12-15
don't auto-execute query after an editor session

pasenor:
## Description
The default way prompt-toolkit handles external editors is to automatically execute the query after an editor session has ended. However, in mycli, it is often undesirable, for example:

- user didn't want to invoke the editor and just accidentally hit `v` in vim mode
- the editing result is not yet ready for execution, the user wanted to do something in the editor, and then continue to edit in mycli (to make use of completions etc)
- user decides to cancel the editor session

In practice it very often leads to unintentional execution of half-backed queries (and in the rare cases when they are semantically correct, there is a danger of accidentally corrupting data).

I suggest to disable auto-execution. The edited query simply appears on the prompt line and can be then run as usual.

- [ x ] I've added my name to the `AUTHORS` file (or it's already there).

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -22,6 +22,7 @@ import click
 import sqlparse
 from prompt_toolkit.completion import DynamicCompleter
 from prompt_toolkit.enums import DEFAULT_BUFFER, EditingMode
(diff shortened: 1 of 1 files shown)


amjith:
Works beautifully! Nicely done. 

:goal_net:

Issue #581, opened 2018-02-13
Config file revamp


tsroten:
In mycli, managing the config has been challenging for awhile since we have five places a user can configure their experience, three of which are config files!
1. The `~/.my.cnf` file (`[client]` section)
2. The login path file, e.g. `~/.mylogin.cnf`
3. The mycli config file, `~/.myclirc`
4. The command-line arguments
5. The commands in mycli

This issue documents that overall revamp of the config that we're trying to accomplish for version 2.

## Tasks

- [ ] Use CLI Helpers to read and validate the config files, see https://github.com/dbcli/cli_helpers/pull/39
- [ ] Store config files according to the [XDG spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html). Windows config files should be in a Windows-appropriate spot and not in the home folder.
- [ ] Make config file options consistent across pgcli and mycli, see https://github.com/dbcli/pgcli/issues/835
- [ ] Add missing configuration options to mycli's config file, e.g. ssl, pager command, etc.
- [ ] Add a migration tool that migrates config settings from `.my.cnf` and `.mylogin.cnf` to the mycli config file (allows user input to override this feature and control password migration)
- [ ] Remove support for `.my.cnf` and `.mylogin.cnf` -- "one config file to rule them all"
- [ ] Store log and audit log files according to the XDG spec.

## Missing options in mycli's config file to consider adding

These options are currently supported in mycli via the `.my.cnf` file, so we need to consider adding them to mycli's config file.

- socket
- default-character-set
- ssl-ca
- ssl-cert
- ssl-key
- ssl-cipher
- ssl-mode
- local-infile
- pager (the command itself)

rixx:
Any timeline on this? Seeing as mycli just *fails completely* if it is not allowed to create ``~/.myclirc`` is a bit annoying, especially since you can't set the location via an env variable (only via a command flag). (Same goes for log files.)

greenskeleton:
Removing support for my.cnf does not make sense to me, for a MySQL tool. I find it very useful to use the same configuration file with both mycli and the mysql command line client in scripts. This is also a nice selling point to new users who are mostly familiar with the standard client.
Without support for my.cnf, I would consider a fork or some alternative software. I would rather continue using mycli, as it is a very nice alternative to the standard client.

pasenor:
[...]

PR #808, opened 2020-02-01
search XDG_CONFIG_HOME for myclirc

iTakeshi:
## Description
Search for a config file named `$XDG_CONFIG_HOME/mycli/myclirc`, as mentioned in #581 .
If `$XDG_CONFIG_HOME` is not exported, it will be defaulted to `~/.config`.

## Checklist


mycli/main.py
```diff
@@ -91,6 +91,8 @@ class MyCli(object):
 
     system_config_files = [
         '/etc/myclirc',
+        os.path.join(
+            os.environ.get("XDG_CONFIG_HOME", "~/.config"), "mycli", "myclirc")
     ]
```
laixintao: Is it necessary to check if the `XDG_CONFIG_HOME` is valide?

For example, what if the env exist but is empty string?

iTakeshi: Yes, it has to be checked. Thanks for your comment.
I'll push a new commit.

iTakeshi: And, I don't think we don't need to check if the `XDG_CONFIG_HOME` is REALLY valid or not.
In the latest commit 8aa2466 , I added a code to check just it is empty or not.
However, when it exists and not empty, even if the referred path is NOT valid (i.e., path not exists), I think we should respect they set the very path for the envvar.

iTakeshi:
Sorry, something went wrong. Wait for an update.

iTakeshi:
↑ It was my simple misunderstanding. Everything works just fine at the latest commit 6d87744 .

amjith:
Looks good. Thank you @iTakeshi 

:cookie:

PR #818, opened 2020-03-02
Downgraded sqlparse dependency, because version 0.3.1 introduced a br…

pasenor:
…eaking change to get_name() method

## Description

The latest release of sqlparse introduced a breaking change, probably a bug: https://github.com/andialbrecht/sqlparse/issues/536

We use `Statement.get_name()` to decide whether a `DROP DATABASE` query affects the currently selected database and completions there is need to update completions. Since this method now always returns None, the `completion_refresh` thread always crashes after dropping a database :(


pasenor:
The error was on our side. Fixed in #819 , no need to downgrade

PR #827, opened 2020-03-15
pinned highest versions of requirements

pasenor:
## Description
We are too permissive with the requirements versions. It's safer to always explicitly specify the last minor version known to work. Otherwise master builds may randomly break because of dependency updates. It happened twice during the last few months (with updates of prompt-toolkit and sqlparse).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,14 +17,14 @@ with open('mycli/__init__.py', 'rb') as f:
 description = 'CLI for MySQL Database. With auto-completion and syntax highlighting.'
 
 install_requirements = [
-    'click >= 7.0',
-    'Pygments >= 1.6',
+    'click >= 7.0,<=7.1.1',
+    'Pygments >= 1.6,<=2.6.1',
     'prompt_toolkit>=2.0.6,<3.0.0',
-    'PyMySQL >= 0.9.2',
+    'PyMySQL >= 0.9.2,<=0.9.3',
     'sqlparse>=0.3.0,<0.4.0',
-    'configobj >= 5.0.5',
-    'cryptography >= 1.0.0',
-    'cli_helpers[styles] > 1.1.0',
+    'configobj >= 5.0.5,<=5.0.6',
+    'cryptography >= 1.0.0,<2.8',
+    'cli_helpers[styles] > 1.1.0,<=1.2.1',
 ]
 
 


pasenor:
Also set up [dependabot](https://app.dependabot.com) to automatically open pull requests with version bumps.

amjith:
This is only going to increase the number of support requests from users. People don't always isolate their apps in a virtualenv. So if they install mycli alongside some other python app with overlapping dependencies, mycli will refuse to start if the strict requirements defined in setup.py are not satisfied.

pasenor:
1. This will not be a problem if we keep those numbers up-to-date. With the bot it should be just a matter of merging PRs from the bot whenever they pass the CI. 
2. At the moment most support requests boil down to package versions as well. To me it just shows that the distribution model through `pip` is not ideal unless the user has experience with python. What do you think of distributing executable PyOxidizer packages by default (and leaving `pip install` as an option)?

meeuw:
Bumping would also be troublesome for the Fedora package unless we document that we support older versions than we specified in setup.py. We might not be affected by the security flaws reported by dependabot anyway, I don't think mycli is hardened to support connecting to untrusted servers anyway...

meeuw:
Oh BTW, the ranges in this PR are fine for Fedora, so feel free to merge this PR.

j-bennet:
FWIW, even if we do pin the upper version (I'm not convinced we need to), I think it would only make sense to protect ourselves from major version changes. I would not pin to a specific minor. And no need to do pinning for `dev` reqs. Those are only installed by contributors, and hopefully contributors know what they're doing.

pasenor:
Well, right now there are issues with the latest version on pypi because of the sqlparse upgrade to the minor version 3.0.1. Had we pinned `<= 3.0.0`, it would be OK. Yes, if we pin upper versions and want to avoid getting obsolete, we'll have to also release minor versions more often. But I see no harm in it, given that the master is normally quite stable.

Yes, pinning dev requirements is probably unnecessary.

amjith:
Looks like this is failing on Python 3.4 because of a missing dependency. 

We don't support that version of Python in master, so if this PR is ready we can merge it in.

andypost:
`prompt_toolkit>=2.0.6,<3.0.0` causes collision with https://github.com/dbcli/pgcli/blob/master/setup.py#L12

amjith:
@pasenor Are you still interested in pursuing this idea or should we close this?

pasenor:
well, it seems impossible to resolve to everyone's satisfaction as long as we ship through pip. I think we should first experiment with the binary deployment and then revisit the handling of dependencies. Let's close this for now.

Issue #845, opened 2020-04-17
auto-completion of favorite names not working any more?


SamCritch:
Hi, mycli favorites are extremely useful, especially when you include variables. I have quite a lot of favorites reducing complex queries to a short statement. 

I'm fairly sure auto-completion used to work for favorite names. For example I could type `\f get_` and mycli would autocomplete with a list like:

`get_customers`
`get_transactions`
`get_logins`
[...]

This has stopped working fairly recently. Is it a config option?

amjith:
Confirmed it is broken. :disappointed: 

I'll look into it. Thanks for reporting.

amjith:
Hi. This is now fixed in master. You can install directly from master using 

```
pip install -U https://github.com/dbcli/mycli/archive/master.zip
```

The change should go out in the next version which will be released in about a week or so.

amjith:
Newest version 1.21.1 released with the fix. 

Please upgrade: `pip install -U mycli`

SamCritch:
Thanks amjith

PR #846, opened 2020-04-17
Fix favorites completion

amjith:
## Description
Addresses #845. 

The FavoriteQueries implementation now closesly resembles pgcli's NamedQueries implementation. 

Updated changelog.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,16 @@
+TBD
+===
+
+Features:
+---------
+
+
+Bug Fixes:
+----------
+
+* Fix broken auto-completion for favorite queries (Thanks: [Amjith]).
+
+
 1.21.0
 ======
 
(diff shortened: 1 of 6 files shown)


amjith:
Yes. It did fix the problem. The reason was the global variable was getting created multiple times when the module was imported in different files with different names. Eg: import iocommands.favoritequeries vs from .packages import special. 

So at completion time the global variable wasn't populated correctly. With the singleton there is only one instance and it works as expected.

j-bennet:
Love it. ❤️ Very nice catch.

laixintao:
Hi @amjith I am confused by

>  Eg: import iocommands.favoritequeries vs from .packages import special.

Why would anyone import the save class from different files in the first place? :)

amjith:
@laixintao We weren't importing the save class. We were importing the favoritequeries from main.py to be able to populate the favorite queries from the config file. Then we import favoritequeries from the sqlcompleter to be able to show completion suggestions. 

But merely importing the module has the side effect of instantiating a new FavoritQueries object. So they were both dealing with different objects. That's why I made it a singleton in this PR.

laixintao:
I Get it. 👍

Issue #839, opened 2020-03-19
new line (Alt+Enter) is not the start of the line.


yahwang:
OS: Ubuntu 18.04
Python: 3.7.5
mycli : 1.20.1 pgcli: 2.2.0

I want to use multi-line with mycli like pgcli does. My situation is like below image.

mycli (left) vs pgcli (right)

![2020-03-19-22:26](https://user-images.githubusercontent.com/12381530/77072636-5f1e8980-6a31-11ea-9d8f-d7a9c54d5e3c.png)

amjith:
I've got a PR that'll let you do this once it is merged.

PR #847, opened 2020-04-17
Use the multiline continuation char from config.

amjith:
## Description
Addresses #839

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -541,7 +541,9 @@ class MyCli(object):
         def get_continuation(width, *_):
             if self.multiline_continuation_char:
                 left_padding = width - len(self.multiline_continuation_char)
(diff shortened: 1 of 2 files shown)


pasenor:
hmm, I'm getting this, both with multiline on and off :thinking: 
image

pasenor:
Ah, do you mean allowing the user to set `""` for prompt_continuation? That one works as expected once I've seen the above picture, but a priori I probably wouldn't expect it to repeat.

amjith:
You're right. I didn't think about the existing config files for all the users. 

Let me think about this a bit more. I'll amend this PR.

amjith:
I have updated this PR. 
image

pasenor:
Looks good, merging

Issue #848, opened 2020-04-19
"local variable 'destroy' referenced before assignment"


yjfguhd:
I have just updated to 1.21.0. After every command, and before the prompt, I am seeing this message. Eg.:

```
$ mycli
Failed to connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock':
(2003, "Can't connect to MySQL server on 'localhost' ([Errno 2] No such file or directory)")
Retrying over TCP/IP
(none)> \?                                                                                                             
+-------------+----------------------------+------------------------------------------------------------+
| Command     | Shortcut                   | Description                                                |
+-------------+----------------------------+------------------------------------------------------------+
| \G          | \G                         | Display current query results vertically.                  |
| \dt         | \dt[+] [table]             | List or describe tables.                                   |
| \e          | \e                         | Edit command with editor (uses $EDITOR).                   |
| \f          | \f [name [args..]]         | List or execute favorite queries.                          |
| \fd         | \fd [name]                 | Delete a favorite query.                                   |
| \fs         | \fs name query             | Save a favorite query.                                     |
| \l          | \l                         | List databases.                                            |
| \once       | \o [-o] filename           | Append next result to an output file (overwrite using -o). |
| \timing     | \t                         | Toggle timing of commands.                                 |
| connect     | \r                         | Reconnect to the database. Optional database argument.     |
| delimiter   | <null>                     | Change SQL delimiter.                                      |
| exit        | \q                         | Exit.                                                      |
| help        | \?                         | Show this help.                                            |
| nopager     | \n                         | Disable pager, print to stdout.                            |
| notee       | notee                      | Stop writing results to an output file.                    |
| pager       | \P [command]               | Set PAGER. Print the query results via PAGER.              |
| prompt      | \R                         | Change prompt format.                                      |
| quit        | \q                         | Quit.                                                      |
| rehash      | \#                         | Refresh auto-completions.                                  |
| source      | \. filename                | Execute commands from file.                                |
| status      | \s                         | Get status information from the server.                    |
| system      | system [command]           | Execute a system shell commmand.                           |
| tableformat | \T                         | Change the table format used to output results.            |
| tee         | tee [-o] filename          | Append all results to an output file (overwrite using -o). |
| use         | \u                         | Change to a new database.                                  |
| watch       | watch [seconds] [-c] query | Executes the query every [seconds] seconds (by default 5). |
+-------------+----------------------------+------------------------------------------------------------+

Time: 0.020s
[...]

PR #849, opened 2020-04-19
Undefined variable destroy

pasenor:
## Description
#848 
I also added the following logic to behave scenarios: 

1. before each test, clear the log,
2. fail the test if there are errors in the log after it finished.

It assumes that no scenario should cause log errors, and that the test log (`~/.mycli.test.log`) has no other use, so there's no harm in deleting it frequently.

- [ x ] I've added my name to the `AUTHORS` file (or it's already there).

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -9,7 +9,7 @@ Bug Fixes:
 ----------
 
 * Fix broken auto-completion for favorite queries (Thanks: [Amjith]).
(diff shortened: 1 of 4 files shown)


amjith:
Thank you for the quick turnaround on this bug fix @pasenor 

:rabbit2:

PR #855, opened 2020-04-28
Fix binary data representation in SQL formatters

pasenor:
## Description
Formatters from sql_format didn't handle binary data correctly (tried to print raw data instead of hex strings)

## Checklist

### DIFF
--- a/mycli/packages/tabular_output/sql_format.py
+++ b/mycli/packages/tabular_output/sql_format.py
@@ -9,6 +9,13 @@ supported_formats = ('sql-insert', 'sql-update', 'sql-update-1',
 preprocessors = ()
 
 
+def escape_for_sql_statement(value):
+    if isinstance(value, bytes):
+        return f"X'{value.hex()}'"
+    else:
+        return formatter.mycli.sqlexecute.conn.escape(value)
+
+
 def adapter(data, headers, table_format=None, **kwargs):
     tables = extract_tables(formatter.query)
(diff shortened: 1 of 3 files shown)


amjith:
Can you give me an example sql statement to test? I'm not sure I understand the fix.

pasenor:
Sure, we need a table with a binary column:
```
mysql root@localhost:test> create table t(x binary);                                                                                                                                                               
Query OK, 0 rows affected
Time: 0.138s
mysql root@localhost:test> insert into t (x) values (0xAA)                                                                                                                                                         
Query OK, 1 row affected
Time: 0.100s
```
Now in ascii format it displays alright:
```
mysql root@localhost:test> select * from t;                                                                                                                                                                        
+------+
| x    |
+------+
| 0xaa |
+------+
```
But not so in sql-insert:
```
mysql root@localhost:test> \T sql-insert                                                                                                                                                                           
Changed table format to sql-insert
Time: 0.000s
mysql root@localhost:test> select * from t;                                                                                                                                                                        
INSERT INTO t (`x`) VALUES
'utf-8' codec can't encode character '\udcaa' in position 4: surrogates not allowed
```

With the fix: 
```
mysql root@localhost:test> select * from t;                                                                                                                                                                        
INSERT INTO t (`x`) VALUES
  (X'aa')
;
```

amjith:
Good catch. 

I fixed the lint error. 

:icecream:

PR #863, opened 2020-05-11
pyoxidizer build

pasenor:
## Description

Building executable with PyOxidizer.

Packaging `cryptography` is such a pain that I replaced it with a pure python library [pyaes](https://github.com/ricmoo/pyaes). It's slower, but we only need it to decrypt tiny `.mylogin` files, so the speed difference is not perceptible. Otherwise it seems to work fine.

I have not figured out how to run pyoxidizer in travis, and switched the CI to github actions, which seem just as nice. Are there any objections?

One thing that needs fixing is the reliance on `__file__` in `mycli.main` (surprisingly, none of the  dependencies has this problem). Because of that we need to ship sources along with the executable, currently the layout is 
```
├── app
│   ├── mycli
│   │     ... (mycli sources)
└── mycli  (executable)
```

The behave suite for python 3.7 is run both against the built executable and against packages installed with `setup.py install`.

## Checklist


pasenor:
Will be split into several PRs

PR #866, opened 2020-05-23
Changed paramiko usage to work with older versions

pasenor:
## Description

Removed dependence of paramiko features that were introduced in version 2.7.

Plus a little refactoring: created a package to use when paramiko is not available. All it does is `sys.exit(1)` with a message. Instead of the idiom
```
if paramiko:
    paramiko.something()
else:
    exit with a message
```
 we can just code as if paramiko were always available.

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -65,7 +65,7 @@ except ImportError:
 try:
     import paramiko
 except ImportError:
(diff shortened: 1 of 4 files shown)


amjith:
Very clever. I like it. 

:chipmunk:

PR #869, opened 2020-05-25
Reuse ssh connection

gfrlv:
## Description

We were spawning a separate SSH connection for the completer thread. It seems unnecessary since we can just open a second channel on the same connection.

## Checklist


amjith:
@pasenor I don't have a good way to check the ssh connection locally, I trust you've done the necessary testing. The code looks good. 

I merged some older PRs, so now the code has some conflicts with master. Can you please resolve them?

pasenor:
Yes, I have tested the SSH connection. I wonder how to add it to the test suite. Looking at the paramiko documentation, they have a server implementation, maybe we can try to use it for testing the SSH-related stuff? (not in this PR, just thinking out loud)

mycli/main.py
```diff
@@ -387,7 +383,7 @@ def connect(self, database='', user='', passwd='', host='', port='',
 
         database = database or cnf['database']
         # Socket interface not supported for SSH connections
-        if port or host or ssh_host or ssh_port:
+        if port or host or self.ssh_client:
```
rolandwalker, line 386: On rebase this should become a little different, like
```
if port or (host and host != 'localhost') or self.ssh_client:
            socket = ''
```

mycli/main.py
```diff
@@ -1098,16 +1093,22 @@ def cli(database, user, host, port, socket, password, dbname,
             else:
                 click.secho(alias)
         sys.exit(0)
+
     if list_ssh_config:
```
rolandwalker, line 1097: Side note: the list_ssh_config feature is unreliable if the SSH config file uses certain features such as Match.  Listing the hosts is not really in the scope of a tool such as mycli, and we ought to consider removing it.

mycli/packages/ssh_client/client.py
```diff
@@ -0,0 +1,46 @@
...
+
+def read_config_file(config_path) -> paramiko.SSHConfig:
+    ssh_config = paramiko.config.SSHConfig()
+    try:
+        with open(config_path) as f:
+            ssh_config.parse(f)
+    # Paramiko prior to version 2.7 raises Exception on parse errors.
+    # In 2.7 it has become paramiko.ssh_exception.SSHException,
+    # but let's catch everything for compatibility
+    except Exception as err:
```
rolandwalker, line 40: `except FileNotFoundError as e:` should come ahead of more general Exception.

mycli/sqlexecute.py
```diff
@@ -90,15 +79,9 @@ def connect(self, database=None, user=None, password=None, host=None,
             '\tsocket: %r'
             '\tcharset: %r'
             '\tlocal_infile: %r'
-            '\tssl: %r'
-            '\tssh_user: %r'
-            '\tssh_host: %r'
-            '\tssh_port: %r'
-            '\tssh_password: %r'
-            '\tssh_key_filename: %r'
+            '\tssl: %r',
```
rolandwalker, line 82: We could still log these values, if we wanted to, right?

rolandwalker (approved):
Also tested connection and completion on the branch.  Works.

rolandwalker:
Closing per https://github.com/dbcli/mycli/issues/1464#issuecomment-3852401749 .

PR #874, opened 2020-06-10
Add --init-command option

### LINKED ISSUE #622: Execute initial command without exiting (like mysql's --init-command)
mycli supports -e/--execute to execute a SQL command, and then it exits. I'm looking for a way to execute an initial command but then keep the session open. My use case is wanting to set session variables on every startup such as "set sql_select_limit=1000, sql_max_join_size=1000000".

Would this be worth a separate command-line switch, like --init-command?

This would work as a more general solution to https://github.com/dbcli/mycli/issues/512.

kit494way:
## Description

Add --init-command option to execute SQL after connecting.
Close #622 

## Checklist


amjith:
That you for the contribution. I appreciate the thoughtful tests. 

:clinking_glasses:

Bit0r:
Can this command line option be used in `myclirc`? It would be great if it could be saved to `myclirc`.

Issue #882, opened 2020-07-18
PyMySQL version 0.10 produces ImportError crash on mycli execution


lukexorz:
After updating PyMySQL from 0.9.3 to 0.10.0 the following crash occurs when executing mycli (version 1.12.1):

```
Traceback (most recent call last):
  File "/usr/bin/mycli", line 33, in <module>
    sys.exit(load_entry_point('mycli==1.21.1', 'console_scripts', 'mycli')())
  File "/usr/bin/mycli", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/usr/lib/python3.8/importlib/metadata.py", line 77, in load
    module = import_module(match.group('module'))
  File "/usr/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 783, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/<user>/.local/lib/python3.8/site-packages/mycli/main.py", line 41, in <module>
    from .sqlexecute import FIELD_TYPES, SQLExecute
  File "/home/<user>/.local/lib/python3.8/site-packages/mycli/sqlexecute.py", line 6, in <module>
    from pymysql.converters import (convert_mysql_timestamp, convert_datetime,
ImportError: cannot import name 'convert_mysql_timestamp' from 'pymysql.converters' (/home/<user>/.local/lib/python3.8/site-packages/pymysql/converters.py)
```

I downgraded to PyMySQL 0.9.3 to continue using mycli

lukexorz:
This looks like the commit that removed the convert_mysql_timestamp function:
https://github.com/PyMySQL/PyMySQL/commit/fe0cd60e6f9e3bc0fb81f76f7e4fa30a1e1b34cc#diff-40496cc4a0f4db6907de60f7ce470b95

lunatiee:
solution 1)  Install PyMySQL 0.9.2 (this is last version that you can use with mycli)

or

solution 2)  
[...]

PR #889, opened 2020-08-01
Make pwd module optional.

amjith:
## Description
Since `pwd` module is only available Unix systems, this change makes that module an optional import.

## Checklist


pasenor:
Have you tested on windows? Otherwise let's wait for feedback on #882 
He's right, we need to figure out how to add windows builds to CI

PR #893, opened 2020-09-01
Add shortcut to login-path parameter.

shaggyz:
## Description

Thanks for this amazing tool! I use a lot the `--login-path` option. I'd like to implement this shortcut `-g` to avoid typing the full parameter. Do you think it can be added?

Thank you again!

## Checklist


shaggyz:
I think CI is failing in `master` for that reason this PR fails as well.

pasenor:
Sorry for the late response. OK, `-g` seems fine, if a bit counter-intuitive. Merging.

PR #900, opened 2020-10-08
Remove unnecessary use of python 2 compat type (Fixes sqlparse 0.4 breaking change)

mtorromeo:
Since python 2 compat has been removed and sqlparse 0.4 also doesn't ship with the compat module anymore, there is no reason to use the text_type alias for str/unicode.

## Description
Fixed compatibility with sqlparse 0.4

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,13 +6,17 @@ Features:
 
 * Add an option `--init-command` to execute SQL after connecting (Thanks: [KITAGAWA Yasutaka]).
 
+Bug Fixes:
+----------
+* Fixed compatibility with sqlparse 0.4 (Thanks: [mtorromeo]).
+
 1.22.2
 ======
 
 Bug Fixes:
 ----------
 
-*  Make the `pwd` module optional. 
+*  Make the `pwd` module optional.
 
 1.22.1
 ======
@@ -785,3 +789,4 @@ Bug Fixes:
 [Georgy Frolov]: https://github.com/pasenor
 [Zach DeCook]: https://zachdecook.com
 [laixintao]: https://github.com/laixintao
+[mtorromeo]: https://github.com/mtorromeo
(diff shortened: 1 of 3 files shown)


eigan:
Could any of the maintainers consider merging this? It should fix issues for everyone using Arch Linux. Thanks.

j-bennet (commented):
I agree that this code can be removed now, but I don't think this was causing issue https://github.com/dbcli/mycli/issues/903. Why I don't think so:

1. mycli restricts sqlparse versions to be below 0.4.0:

https://github.com/dbcli/mycli/blob/f3442f40d6378372cc0c2f9350a34ee9911336f3/setup.py#L24

2. The error says `No module named 'sqlparse.sql'`, not `sqlparse.compat`.

If you do want to fix compatibility with sqlparse 0.4.0 here, you should remove the `<0.4.0` restriction from setup.py (probably limit to `<0.5.0`), and we'd have to make sure tests are passing. It looks like tests are failing on master, perhaps we need the same fix as in this PR in pgcli: https://github.com/dbcli/pgcli/pull/1211/files.

eigan:
It was a mistake to reference #903, I removed my comment shortly after discovering the same thing. But the reference would not disappear from this pull request.

j-bennet:
Ah! Ok, the comment about updating `setup.py` is still relevant.

mtorromeo:
Arch Linux is packaging sqlparse 0.4.0 and that is what mycli is trying to use regardless of setup.py. While this can be considered a packaging bug the proposed fix still works (verified on my system) and the modification will still be required once the restriction in setup.py is lifted.

j-bennet:
@mtorromeo Any reason why we should not update `setup.py` to use 0.4.0 and up?

j-bennet:
@mtorromeo I merged your PR as-is - we can address updating `sqlparse` later. Thank you! 🍫

mtorromeo:
> @mtorromeo Any reason why we should not update `setup.py` to use 0.4.x?

No reason not to do it. I was just saying that updating setup.py is not a requirement to apply this fix.

PR #902, opened 2020-10-20
Expanduser in config file paths

ateijelo:
## Description

One of the config.py functions was not calling `expanduser`. I recommend that `expanduser` is used in the place where the string `~/.my.cnf` is defined, and make the config functions require well defined paths. This way you don't have to remember to call it in every existing or future function.

## Checklist


pasenor:
Thanks! And sorry for the long response time. The change makes sense. For completeness, I'd suggest
1. writing a test that fails when the path is not expanded (which function is it, by the way?)
2. removing invocations of expanduser that have become redundant

PR #904, opened 2020-10-23
Fix behave tests on master.

j-bennet:
## Description

Fix for failing behave tests.


j-bennet:
This does fix the tests, so I'm going to merge.

PR #913, opened 2020-12-15
Allow to pass file instead of password

jeromegit:
## Description

As suggested https://www.netmeister.org/blog/passing-passwords.html article it's a good practice to allow to pass a path to a file or FIFO containing a password vs passing the actual password
I made a small change to the code (just a few lines into 2 places)

## Checklist


pasenor:
Hi! Thanks for the contribution, although I'm not sure mycli itself should be responsible for handling password files in this way. The article you link to suggests to use standard shell magic:
```
mycli --pass=$(cat file_with_password)
```
It works just fine with mycli. Another option, specific to MySQL, is to use the encrypted `mylogin.cnf`: https://dev.mysql.com/doc/refman/8.0/en/mysql-config-editor.html.
Am I overlooking some use case?

jeromegit:
Hi @pasenor,
Thanks for getting back to me and for creating such a great tool!

As mentioned in the article https://dev.mysql.com/doc/refman/8.0/en/mysql-config-editor.html, the approach of using mycli --pass=$(cat file_with_password) is discouraged because anyone running a ps would be able to see what's after the --pass in clear.

Using the approach I'm advocating (especially when using a FIFO) is safer since the approach will be to create a wrapper around mycli to do the following:
1. start wrapper
2. wrapper pulls credentials (login/password) from the credential repository (my company uses: https://teampasswordmanager.com/)
3. wrapper creates FIFO file (/my/fifo/path) with the password to be passed to mycli
4. wrapper launches mycli with --pass /my/fifo/path
5. once mycli is done we can remove the FIFO

This is what we are using at my company and I figured that others might be interested in that safer approach.

Thanks.

pasenor:
I forgot to mention: we support login with [mylogin.cnf](https://dev.mysql.com/doc/refman/8.0/en/mysql-config-editor.html) files that can be created with mysql_config_editor. Does it work for your use case?  I'm still not convinced we should make the `--password` argument ambiguous. @amjith what's your take on this?

pasenor:
@jeromegit we just had a discussion an decided that we will accept the PR if it adds a separate keyword argument for the password file (without reusing `--password`). Otherwise there is a potential for confusion (what if the password matches the filename? what if the file doesn't exist? can a password start with `/` etc)
Would that work for you?

jeromegit:
Thank you for getting back to me, Georgy.
Sounds good to me!
How about --password-file PATH, where one can provide the path for plain
file or a FIFO?
It would only kick in when no password is provided.

How does it sound?

Jérôme  / Blog <http://provensal.com/lbb/> / FB
<https://www.facebook.com/jerome.provensal> / LinkedIn
<http://www.linkedin.com/in/jeromeprovensal> / IG
<http://instagram.com/jeromeig> / INTP
<http://www.mypersonality.info/personality-types/intp/>

pasenor:
Yeah, sounds great

jeromegit:
Great!
I'll work on it this weekend.

Jérôme  / Blog <http://provensal.com/lbb/> / FB
<https://www.facebook.com/jerome.provensal> / LinkedIn
<http://www.linkedin.com/in/jeromeprovensal> / IG
<http://instagram.com/jeromeig> / INTP
<http://www.mypersonality.info/personality-types/intp/>

jeromegit:
I think, I'm done.
Please note that I ended up using the --password-file in 2 places:
1. when not password is otherwise specified (or obtained from cnf file)
2. when the specified password is failing

Please review the changes and let me know what you think.

Thanks.

README.md
```diff
@@ -58,6 +58,7 @@ $ sudo apt-get install mycli # Only on debian or ubuntu
       -S, --socket TEXT             The socket file to use for connection.
       -p, --password TEXT           Password to connect to the database.
       --pass TEXT                   Password to connect to the database.
+      --pass PATH                   Password contained in a file or FIFO to connect to the database.
```
pasenor, line 61: it still says `--pass`

mycli/main.py
```diff
@@ -1086,14 +1106,17 @@ def get_last_query(self):
               help='Warn before running a destructive query.')
 @click.option('--local-infile', type=bool,
               help='Enable/disable LOAD DATA LOCAL INFILE.')
-@click.option('-g', '--login-path', type=str,
```
pasenor, line 1089: is that change intentional?

mycli/main.py
```diff
@@ -489,6 +497,18 @@ def _connect():
...
 
+
+    def get_password_from_file(self, password_file):
+        password_from_file = None
+        if password_file:
+            if (os.path.isfile(password_file) or \
+                stat.S_ISFIFO(os.stat(password_file).st_mode)) \
+                and os.access(password_file, os.R_OK):
+                with open(password_file) as fp:
+                    password_from_file = fp.read()
```
pasenor, line 508: Should we strip trailing `\n`, `\r\n` and space on both sides?

mycli/main.py
```diff
@@ -489,6 +497,18 @@ def _connect():
             self.echo(str(e), err=True, fg='red')
             exit(1)
 
+
+    def get_password_from_file(self, password_file):
+        password_from_file = None
+        if password_file:
+            if (os.path.isfile(password_file) or \
+                stat.S_ISFIFO(os.stat(password_file).st_mode)) \
```
pasenor, line 505: I think the linter will complain about indentations here. You can check with `python setup.py lint` (and fix with `python setup.py lint --fix`)

mycli/main.py
```diff
@@ -1086,14 +1106,17 @@ def get_last_query(self):
...
+@click.option('--login-path', type=str,
               help='Read this path from the login file.')
 @click.option('-e', '--execute',  type=str,
               help='Execute command and quit.')
 @click.option('--init-command', type=str,
               help='SQL statement to execute after connecting.')
 @click.option('--charset', type=str,
               help='Character set for MySQL session.')
+@click.option('--password-file', type=click.Path(),
+              help='File or FIFO path containing the password to connect to the db if not specified otherwise')
```
pasenor, line 1118: CI fails because there is no period at the end of the help text

jeromegit:
Hi Georgy,
Sorry, I'm not ghosting/ignoring you.
Super busy at work and it seems that I can only find time to work on cool projects like mycli on the weekend.
I'll make all the necessary changes tomorrow or Sunday.

Most of them are due to my lack of familiarity with Git (more like an SVN guy here) and the lack of good tools where I cloned the project.
I now cloned the project on my Mac and installed MySQL locally so that I can easily test it and do the git'ing much easier.

Thank you for your patience!

pasenor:
There's no hurry, take your time!

README.md
jeromegit: my bad, I didn't commit the README.md file correctly

mycli/main.py
jeromegit: excellent idea. Will do

mycli/main.py
jeromegit: fixed

mycli/main.py
jeromegit: fixed it manually but still ran lint with --fix which seems to have fixed a couple of other items

jeromegit:
@pasenor I think I have made all the changes you suggested.
Please advise if I missed anything or if changes need to be merged.
Thanks!

pasenor:
Looks good, thanks! 
I'd also add a test, but it can be addressed in a separate PR

jeromegit:
Thanks again for accepting my weird request and for your patience as I was
learning the ropes with the whole GH pull-request workflow!
Do you know when the new version will make it into pypi.org?

Thanks.

Issue #909, opened 2020-11-16
How to set mycli character encoding set?


ICLXL:
I get the data in my mysql server cli normally, but the data through mycli is garbled.

rolandwalker:
Hi!

Currently the best way is to set

```
[client]
default-character-set = utf8mb4
```

in `~/.my.cnf`.

But in #915, @seamile is also working on a `--charset` CLI option to change the character set for individual connections.

PR #915, opened 2020-12-29
Add `--charset` option for  commandline

seamile:
## Description

Add `--charset` option for command line.

The default character set of mycli is `UTF8`. But when a field contains a emoji or other special characters, the charset should be `UTF8MB4`.

So I add this option to set the default charset when open mycli. Otherwise we need to modify the config file.

### Before:
```sql
$ mycli -u seamile -h localhost
Password: 

> select id, name, comment from artist where id = 963;
+-----+-------+-----------------+
| id  | name  | comment         |
+-----+-------+-----------------+
| 963 | 大猫?  | ? Hello world ? |
+-----+-------+-----------------+

mysql > show variables like 'character_set%';
+--------------------------+----------------------------------+
| Variable_name            | Value                            |
+--------------------------+----------------------------------+
| character_set_client     | utf8                             |
| character_set_connection | utf8                             |
| character_set_database   | utf8mb4                          |
| character_set_filesystem | binary                           |
| character_set_results    | utf8                             |
| character_set_server     | utf8mb4                          |
| character_set_system     | utf8                             |
| character_sets_dir       | /usr/local/share/mysql/charsets/ |
+--------------------------+----------------------------------+
```

### After:
```sql
$ mycli -u seamile -h localhost --charset=utf8mb4
Password: 

mysql> use pix;
mysql> select id,name,comment from artist where id = 963;
+-----+--------+-------------------+
| id  | name   | comment           |
+-----+--------+-------------------+
| 963 | 大猫🐯  | 😄 Hello world 😄  |
+-----+--------+-------------------+

mysql > show variables like 'character_set%';
+--------------------------+----------------------------------+
| Variable_name            | Value                            |
+--------------------------+----------------------------------+
| character_set_client     | utf8mb4                          |
| character_set_connection | utf8mb4                          |
| character_set_database   | utf8mb4                          |
| character_set_filesystem | binary                           |
| character_set_results    | utf8mb4                          |
| character_set_server     | utf8mb4                          |
| character_set_system     | utf8                             |
| character_sets_dir       | /usr/local/share/mysql/charsets/ |
+--------------------------+----------------------------------+
```

## Checklist


pasenor:
Is there a use case when the database supports UTF8MB4, but we still want to use UTF8? If not, I think we should just use the full encoding whenever possible, why bother the user with an extra flag

seamile:
> Is there a use case when the database supports UTF8MB4, but we still want to use UTF8? If not, I think we should just use the full encoding whenever possible, why bother the user with an extra flag

This option just gives a choice for user when someone need to change the charset.
It does not change the default character set of the database.
If we want to use UTF8, we can ignore this option and use mycli as before：`mycli -u username -h localhost`.

In fact, the `UTF8MB4` is the full encoding of MySQL, not `UTF8`. It can encode the full Unicode character set.
Without UTF8MB4 we cannot display emoji correctly in mycli. And the Chinese, Japanese, Korean also need it.
The issue #909  describes the same problem encountered by others

Some links:
- [MySQL Chinese, Japanese, and Korean Character Sets](https://dev.mysql.com/doc/refman/8.0/en/faqs-cjk.html)
- [In MySQL, never use “utf8”. Use “utf8mb4”](https://medium.com/@adamhooper/in-mysql-never-use-utf8-use-utf8mb4-11761243e434)

rolandwalker:
@pasenor I support this PR as it is.

I have a legacy application which still connects to a MySQL database using the `latin1` character set.  It is useful to be able to simulate exactly what that application sees.

Separately, we should absolutely change the default connection to `utf8mb4`.  Worth noting is that a `utf8mb3` has been introduced in recent MySQLs, and that [at some point utf8 is expected to become a reference to utf8mb4](https://dev.mysql.com/doc/refman/8.0/en/charset-unicode-sets.html).  But we aren't there yet, and `utf8mb4` is currently more standards-compliant.

rolandwalker:
@seamile you do realize as well that you can set
```
[client]
default-character-set = utf8mb4
```
in `~/.my.cnf` ?

seamile:
@rolandwalker  Yes, I know this way. However, this option makes it easier to *temporarily* set the charset to a different value than the config file.

pasenor:
OK, agreed, let's merge this as is and worry about the defaults separately.

PR #916, opened 2020-12-31
Add pipe_once special command

rolandwalker:
## Description
Add a `\pipe_once` special command, alias `\|`, which is modeled on `\once` but sends output to a subprocess.  Any output from the subprocess is presented after the query results.

One usecase I had in mind is creating gists:

```
mysql root@localhost:(none)> \| gist; show databases;
Time: 0.003s

+--------------------+
| Database           |
|--------------------|
| information_schema |
| mysql              |
| performance_schema |
+--------------------+
3 rows in set
Time: 0.011s
https://gist.github.com/0b9e651d00562ad527b0d4a765dcc292
mysql root@localhost:(none)>
```

The `command` argument is parsed with `shlex`, so options may be passed as well, example `\| gist -f filename.tsv`.

If the idea is acceptable I will happily work on tests.  So far I do not have the test suite fully operational.

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,7 @@ Features:
 
 * Add an option `--init-command` to execute SQL after connecting (Thanks: [KITAGAWA Yasutaka]).
 * Use InputMode.REPLACE_SINGLE
+* Add a special command `\pipe_once` to pipe output to a subprocess.
 
 Bug Fixes:
 ----------
--- a/mycli/AUTHORS
+++ b/mycli/AUTHORS
@@ -78,6 +78,7 @@ Contributors:
   * bitkeen
   * Morgan Mitchell
   * Massimiliano Torromeo
+  * Roland Walker
 
 Creator:
 --------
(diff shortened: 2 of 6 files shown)


rolandwalker:
I can get this far on tests:
```python
def test_pipe_once_command():
    with pytest.raises(IndexError):
        mycli.packages.special.execute(None, u"\\pipe_once")

    with pytest.raises(OSError):
        mycli.packages.special.execute(None, u"\\pipe_once /proc/access-denied")

    mycli.packages.special.execute(None, u"\\pipe_once wc")
    mycli.packages.special.write_once(u"hello world")
    mycli.packages.special.unset_pipe_once_if_written()
```
But haven't figured out how to reference the output of `wc` in order to make an assertion.

`IndexError` is also not great for the case of a missing command.

amjith:
Not entirely satisfied with the test but I don't have a better suggestion at this time. So merging. 

:ice_cube:

PR #917, opened 2021-01-01
Add \clip special command

rolandwalker:
## Description

A limitation of prompt-toolkit is that it [wraps the query text, making copy/paste difficult](https://github.com/prompt-toolkit/python-prompt-toolkit/issues/493#issuecomment-398894847).

But that can be worked around in mycli with a `\clip` special command. Example usage:

```sql
\clip LONG QUERY…
```

I didn't abbreviate `\clip` to `\c`, as that might also mean `\connect`.

This adds a dependency on `pyperclip`,

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,7 @@ Features:
 
 * Add an option `--init-command` to execute SQL after connecting (Thanks: [KITAGAWA Yasutaka]).
 * Use InputMode.REPLACE_SINGLE
+* Add a `\clip` special command to copy queries to the system clipboard.
 
 Bug Fixes:
 ----------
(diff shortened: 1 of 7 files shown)


amjith:
I don't understand the intent of this special command. In what sort of workflow would this be useful? 

I'm also seeing that when I run `\clip SELECT "ABC"` the result isn't printed. So is this purely for copy/pasting the query and not the results? 

If so can this be achieved by opening the history file located in ~/.mycli-history and copying the query from there?

rolandwalker:
Whenever a line wraps in prompt-toolkit, prompt-toolkit inserts hard newlines.   Therefore, in mycli, users can't copy a wrapped query using a mouse, the usual way.  Or you can, but what lands on the clipboard is not faithful to what you typed.

Yes, the feature was intended purely for copy/pasting the query.

Yes, the same effect would be be achieved by opening ~/.mycli-history and copying the bottom-most query.

PR #919, opened 2021-01-01
Fix \once -o to overwrite output whole, instead of overwriting line-by-line

rolandwalker:
## Description
Since `\once -o filename.txt` was overwriting on a line-by-line basis, `filename.txt` would only contain the last line of the result, such as 

```
+--------------------+
```

Instead the filehandle should be kept open, and the entire result sent to the open handle.

## Checklist


amjith:
:cheese:

PR #920, opened 2021-01-01
Prefer raw strings for regular expressions

rolandwalker:
## Description
Raw strings are better for regular expressions; fewer backslashes are needed.

Arguably the whole codebase could be made clearer by replacing non-regex strings like `'\\u'` with `r'\u'`.

## Checklist
- [ ] I've added this contribution to the `changelog.md`. (so minor I did not)

### DIFF
--- a/mycli/packages/special/iocommands.py
+++ b/mycli/packages/special/iocommands.py
@@ -115,7 +115,7 @@ def get_editor_query(sql):
     # The reason we can't simply do .strip('\e') is that it strips characters,
     # not a substring. So it'll strip "e" in the end of the sql also!
     # Ex: "select * from style\e" -> "select * from styl".
-    pattern = re.compile('(^\\\e|\\\e$)')
+    pattern = re.compile(r'(^\\e|\\e$)')
     while pattern.search(sql):
         sql = pattern.sub('', sql)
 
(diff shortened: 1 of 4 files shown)


mycli/packages/special/iocommands.py
```diff
@@ -115,7 +115,7 @@ def get_editor_query(sql):
     # The reason we can't simply do .strip('\e') is that it strips characters,
     # not a substring. So it'll strip "e" in the end of the sql also!
     # Ex: "select * from style\e" -> "select * from styl".
-    pattern = re.compile('(^\\\e|\\\e$)')
+    pattern = re.compile(r'(^\\e|\\e$)')
```
amjith, line 118: If this becomes a raw string wouldn't that mean only one `\` is required to escape the `\e`  or am I mistaken?

rolandwalker: If it was `print(r'\e')` then yes, only one, but inside a regular expression Python would attempt to interpret `\e` as a (nonexistent) regular-expression escape.  I keep that straight in my head by thinking: if `\` has special semantics in a regex, then an escape is needed to quote the meaningful backslash.  (But actually it's unfortunate design that `\` has two meanings in regular expressions.)

amjith: :+1:

PR #921, opened 2021-01-01
Respect empty string value for prompt_continuation

rolandwalker:
## Description
When `.myclirc` contains
```
prompt_continuation = ''
```

then the continuation string should be the empty string, as specified, not a single space.

## Checklist


amjith:
Looks good. 

:moon_cake:

PR #922, opened 2021-01-02
Allow fine-grained customization of Pygments SQL highlighting styles

rolandwalker:
## Description
Allows individual customization of each syntax-highlighting face that the Pygments MySQL lexer can produce.

Doesn't change any behavior by default.  The user would have to add/enable entries in `~/.myclirc`.

Populates the template `~/.myclirc` with commented-out entries matching `syntax_style = default`.

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,7 @@ Features:
 
 * Add an option `--init-command` to execute SQL after connecting (Thanks: [KITAGAWA Yasutaka]).
 * Use InputMode.REPLACE_SINGLE
+* Allow customization of Pygments SQL syntax-highlighting styles.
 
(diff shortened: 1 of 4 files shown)


amjith:
I'm usually reluctant to add configuration options that don't provide value. I like the design philosophy of fish shell. https://fishshell.com/docs/current/design.html#configurability-is-the-root-of-all-evil

In this case I'm going to merge this since it doesn't take away from having sensible defaults and syntax coloring is a highly subjective feature.

rolandwalker:
Thank you for making an exception for this case!

Many of the provided styles don't provide sufficient contrast against a dark background, and certain individual colors need a little readability boost.

PR #924, opened 2021-01-02
Make Alt-Enter context-sensitive, depending on multi-line or single-line mode

rolandwalker:
## Description

xref #748

Make <kbd>Alt-Enter</kbd> context-sensitive, inserting a linefeed in single-line mode, but dispatching the command in multi-line mode.

In short, <kbd>Alt-Enter</kbd> mostly inverts the expected action of <kbd>Enter</kbd>.

Motivation: I can't find another way to dispatch a command like
```
SELECT * from TABLE WHERE \e
```
in multi-line mode.

Also, make <kbd>Control-j</kbd> unconditionally insert linefeed in Emacs mode.  Rationale: preserves an easy "unconditional insert" feature.  Happy to remove this part, which may be debatable. Edit: and, the Control-j binding may break the assumptions of the test suite.

Note that it was always possible to unconditionally insert linefeeds via quoted-insert (Emacs mode <kbd>Control-q</kbd> <kbd>Control-j</kbd>, Vi mode <kbd>Control-v</kbd> <kbd>Control-j</kbd>).

## Checklist

### DIFF
--- a/mycli/key_bindings.py
+++ b/mycli/key_bindings.py
@@ -78,8 +78,12 @@ def mycli_bindings(mycli):
 
     @kb.add('escape', 'enter')
     def _(event):
-        """Introduces a line break regardless of multi-line mode or not."""
+        """Introduces a line break in multi-line mode, or dispatches the
+        command in single-line mode."""
         _logger.debug('Detected alt-enter key.')
-        event.app.current_buffer.insert_text('\n')
+        if mycli.multi_line:
+            event.app.current_buffer.validate_and_handle()
+        else:
+            event.app.current_buffer.insert_text('\n')
 
     return kb
--- a/changelog.md
+++ b/changelog.md
@@ -9,6 +9,7 @@ Bug Fixes:
 Features:
 ---------
 * Add `-g` shortcut to option `--login-path`.
+* Alt-Enter dispatches the command in multi-line mode.
 
 Internal:
 ---------


mycli/key_bindings.py
```diff
@@ -76,10 +76,24 @@ def _(event):
...
+            event.current_buffer.auto_down(count=event.arg)
+
     @kb.add('escape', 'enter')
     def _(event):
-        """Introduces a line break regardless of multi-line mode or not."""
+        """Introduces a line break in multi-line mode, or dispatches the
+        command in single-line mode."""
         _logger.debug('Detected alt-enter key.')
-        event.app.current_buffer.insert_text('\n')
+        if mycli.multi_line:
```
j-bennet, line 94: Can this be `event.app.multi_line` (for consistency)?

mycli/key_bindings.py
```diff
@@ -76,10 +76,24 @@ def _(event):
         b = event.app.current_buffer
         b.complete_state = None
 
+    @kb.add('c-j')
+    def _(event):
+        """Introduces a line break in Emacs mode, regardless of multi-line
+        mode or not."""
+        _logger.debug('Detected control-j key.')
+        if event.app.editing_mode == EditingMode.EMACS:
+            event.app.current_buffer.insert_text('\n')
```
j-bennet, line 85: Why `\n` and not `os.linesep`?

rolandwalker: I don't understand the platform-specific subtleties well enough to be confident which way is correct.  Here I just copied the existing `'\n'`.

rolandwalker:
I just noticed [_multiline_exception()](https://github.com/dbcli/mycli/blob/4517f49efdcc167a672bce9fdb922bb7224ce5f5/mycli/clibuffer.py#L20) which is another way that ending a line with `\e` could/should be supported.

amjith:
Do we just need to kick off a new test run to verify that it passes or are we still waiting on changes in this PR?

rolandwalker:
From our Slack conversation I thought you had an idea about how to change the PR.  My only idea was to remove Control-J.

j-bennet:
@rolandwalker You could make Control + J binding conditional, and not bind it in unit tests, like this:

```
from prompt_toolkit.filters import Condition

@Condition
def no_cpr():
    """Only if not PROMPT_TOOLKIT_NO_CPR."""
    import os
    return os.environ.get('PROMPT_TOOLKIT_NO_CPR', '') == '1'
```

then:

`@kb.add('c-j', filter=~no_cpr)`

However, I don't know all the implications of overriding the default binding of `c-j`. `prompt_toolkit` handles `c-j` in a few places, example:

https://github.com/prompt-toolkit/python-prompt-toolkit/blob/4c2fb15e7f4838d46467b0ca193ae6d543b4172a/prompt_toolkit/key_binding/bindings/basic.py#L186

maybe @jonathanslenders can advise here?

rolandwalker:
In the latest push, I am just removing the <kbd>Control-J </kbd> binding.

j-bennet:
@rolandwalker are you able to rebase this on master?

rolandwalker:
@j-bennet rebased

j-bennet:
👍  Merged. Thank you!

PR #767, opened 2019-08-18
Adding mysql 5.7 and 8.0 keywords and functions for auto-completion

laoshaw:
## Description

Added keywords and json functions from mysql 5.7 and 8.0 for auto-completion

## Checklist

- [ ] I've added this contribution to the `changelog.md`.  -- not seen this file
- [ ] I've added my name to the `AUTHORS` file (or it's already there). -- change too small to be added

PR #925, opened 2021-01-04
Import MySQL completion candidates from pygments

rolandwalker:
## Description

Since the `pygments` library is already required, why not use its list of reserved words for completions?

This fixes #767 and should keep us up-to-date.  It's really great to have the new `JSON_*` functions.

Compared to the previous code, this PR
 * removes LEN and TOP, which are not MySQL reserved words
 * preserves extra completion candidates containing space, such as "ORDER BY"

Downsides and bugs:
 * `pygments.lexers._mysql_builtins` does contain a leading underscore, so we should be aware that the library reserves the right to break this usage.  We might consider pinning the library version, but I don't honestly expect there will be a problem.
 * Certain reserved words are duplicated with special commands, so if the first word of a command-line is any of the following, duplicated completions will show, as both upper- and lower-case: `exit`, `help`, `source`, `status`, `system`, `use`.  This is fixable, but should we prefer the upper- or lower-case flavor?

## Checklist


amjith:
Two things: 

1. The mysql keyword lexer is only available in newer versions of Pygments, so we should update setup.py to pin the min version of pygments to something more recent.
2. The previous way of defining the keywords allowed us to control the order of the suggestions in the completion menu. For example, the SELECT keyword is higher up in the list and so is the FROM keyword. So they show up in the list higher so it is easier for the user to select them with fewer keystrokes. Maybe we can define a list of most common keywords in a separate list and have them prepended to the lexer's keywords and remove the sorting. We can also uniqe the list to ensure we remove the duplicates. 

Thoughts?

rolandwalker:
That sounds good, will try.

_Ideally_, we'd form that most-common list dynamically from the user's own history.

amjith:
> Ideally, we'd form that most-common list dynamically from the user's own history.

I believe we have that feature, I can't tell if it is in pgcli or mycli. Let me look it up.

amjith:
https://github.com/dbcli/pgcli/blob/master/pgcli/packages/prioritization.py

rolandwalker:
Closing in favor of #1447 .

PR #929, opened 2021-01-05
Stop editor test from creating file "select"

rolandwalker:
## Description
 * `open_external_editor()` was ignoring the `$EDITOR` environment variable when `$VISUAL` was set
 * the "select" in "select 1" was interpreted as a filename when the string was passed as a positional parameter

So, when `$VISUAL` was set, the editor was opened on a file named `select`, which may create the file, depending on the editor.

## Checklist


amjith (approved):
Good catch. :1st_place_medal:

PR #934, opened 2021-01-05
Add Python 3.9 and "on: pull_request" stanza to CI spec

rolandwalker:
## Description
Add Python 3.9 to CI, ~~and originate the PR from my personal fork, to additionally find out whether CI fires correctly on that event.~~

Testing showed that CI wouldn't run without the "on: pull_request" stanza, also added in this PR.

(I don't understand why GitHub seems to require both "on-push" and "on-pull_request" events.  It seems wasteful.)

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,9 +1,7 @@
 name: mycli
 
 on:
-  push:
-    branches-ignore:
-      - 'master'
+  pull_request:
     paths-ignore:
       - '**.md'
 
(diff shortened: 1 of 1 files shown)


rolandwalker:
Let's try this with only the `on: pull_request` stanza, to avoid duplicate CI runs.

`./setup.py lint --branch=HEAD` was needed to get the lint step to run on the PR event, in which branch names are somewhat obscured.  I'm not sure why the `--branch=` argument is needed.

PR #935, opened 2021-01-07
WIP: Default to standards-compliant utf8mb4 character set

rolandwalker:
## Description

xref #915 

The `utf8` character set in current MySQL versions is not actually standards-compliant.  The standards-compliant UTF-8 character set is spelled `utf8mb4`, and that should be mycli's default.

WIP because this should be researched/tested for MariaDB and Percona.

## Checklist


pasenor:
I think this default should depend on the server version: mysql < 5.5 cannot handle utf8mb4.

rolandwalker:
@pasenor excellent point.

rolandwalker:
Wait, how would that work?  We set the charset before we make the connection.

pasenor:
Oops, indeed, that's messy. But we should be able call  [set_charset()](https://github.com/PyMySQL/PyMySQL/blob/abe83c262ea647a09e0f13587fa91d6a14a71598/pymysql/connections.py#L586) on the PyMySQL connection object inside our SQLExecute once know the version

dveeden:
Note that `utf8` is considered an alias for `utf8mb3` and both MySQL and MariaDB are actively doing work to eventually change the alias to map to `utf8mb4`.

- https://dev.mysql.com/doc/refman/8.0/en/charset-unicode-utf8.html
- https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-30.html 

> Important Change: A previous change renamed character sets having deprecated names prefixed with utf8_ to use utf8mb3_ instead. In this release, we rename the utf8_ collations as well, using the utf8mb3_ prefix; this is to make the collation names consistent with those of the character sets, not to rely any longer on the deprecated collation names, and to clarify the distinction between utf8mb3 and utf8mb4. The names using the utf8mb3_ prefix are now used exclusively for these collations in the output of SHOW statements such as [SHOW CREATE TABLE](https://dev.mysql.com/doc/refman/8.0/en/show-create-table.html), as well as in the values displayed in the columns of Information Schema tables including the [COLLATIONS](https://dev.mysql.com/doc/refman/8.0/en/information-schema-collations-table.html) and [COLUMNS](https://dev.mysql.com/doc/refman/8.0/en/information-schema-columns-table.html) tables. (Bug #33787300)


- https://jira.mariadb.org/browse/MDEV-8334
- https://jira.mariadb.org/browse/MDEV-22217

The `utf8mb4` character set was introduced in MySQL 5.5.4, this was not a G.A. release (5.5.8 was the first G.A. release). So `utf8mb4` should be used for MySQL 5.5 and newer.

- https://downloads.mysql.com/docs/mysql-5.5-relnotes-en.pdf

amjith:
@rolandwalker Is this PR still valid? Should this be merged?

rolandwalker:
@amjith yes we should do something about it.  Will review.

rolandwalker:
Reviving this old PR!

The concern earlier was mostly that MySQL 5.5 didn't support the `utf8mb4` charset.  But 5.5 was released in 2010.  At this point we can set the standards-compliant `utf8mb4` charset and document how to downgrade it for any users of 5.5 or earlier.

The PR has also been updated to add a `default_character_set` to `~/.myclirc`.  In general, we should find a path to move away from supporting `~/.my.cnf` on top of `~/.myclirc`.

scottnemes (approved):
It's an old meme, but it checks out!

Verified client/connection charset is now utf8mb4 by default when connecting; looks good.

PR #940, opened 2021-01-10
Get server version from pymysql without additional queries

pasenor:
## Description
We can get rid of two queries that ask the server about it's version: the pymysql connection object already contains everything we want to know.

Also, added a numerical representation of the server version, for future convenience if we start adding any version-specific functionality (which I think we should do).

## Checklist


test/test_sqlexecute.py
```diff
@@ -270,3 +271,19 @@ def test_multiple_results(executor):
...
     assert results == expected
+
+
+@pytest.mark.parametrize(
+    'version_string, species, parsed_version_string, version',
+    (
+        ('5.7.32-35', 'Percona', '35', 35),
+        ('5.7.32-0ubuntu0.18.04.1', 'MySQL', '5.7.32', 50732),
+        ('10.5.8-MariaDB-1:10.5.8+maria~focal', 'MariaDB', '10.5.8', 100508),
+        ('unexpected version string', None, 'unexpected version string', 0),
```
j-bennet, line 282: This needs a test case for empty string and `None`.

mycli/sqlexecute.py
```diff
@@ -18,17 +20,67 @@
...
+        self.version_str = version_str
+        if species == ServerSpecies.Percona:
+            self.version = int(version_str)
+        else:
+            self.version = self.calc_mysql_version_value(version_str)
+
+    @staticmethod
+    def calc_mysql_version_value(version_str):
+        try:
+            major, minor, patch = version_str.split('.')
```
j-bennet, line 42: Probably should handle None as well.

mycli/main.py
```diff
@@ -933,7 +933,9 @@ def get_prompt(self, string):
         string = string.replace('\\u', sqlexecute.user or '(none)')
         string = string.replace('\\h', host or '(none)')
         string = string.replace('\\d', sqlexecute.dbname or '(none)')
-        string = string.replace('\\t', sqlexecute.server_type()[0] or 'mycli')
+        string = string.replace(
+            '\\t',
+            sqlexecute.server_info.species and sqlexecute.server_info.species.name or 'mycli')
```
j-bennet, line 938: We probably want `mysql`, not `mycli` here?

mycli/sqlexecute.py
```diff
@@ -18,17 +20,67 @@
...
+                species = ServerSpecies.MariaDB
+                parsed_version_str = mysql_version
+            elif comment.isdigit():
+                species = ServerSpecies.Percona
+                parsed_version_str = comment
+            else:
+                species = ServerSpecies.MySQL
+                parsed_version_str = mysql_version
+        else:
+            species = None
```
j-bennet, line 66: How about adding `ServiceSpecies.Unknown` (or Generic or something like that)? This way, you won't need to check `species` for None.

pasenor: good idea

mycli/main.py
pasenor: well, that's the current implementation: if we don't understand, say 'mycli' (probably because it may be some renamed fork of mysql)

j-bennet: I think maybe it was a typo from the beginning :)

pasenor: hmm, I get your point, it's kind of inconsistent: if this spot is reserved for the server, we should not replace it with the client name. Would it be appropriate to spell the generic server also `MySQL`?

pasenor:
After some research it seems that special treatment of Percona DB is not warranted since dialect-wise it should be perfectly compatible with the underlying MySQL version.

mycli/main.py
j-bennet: I think so. You could set "ServerSpecies.Unknown" to map to string "MySQL", and just use `species.name` here, and never even check for None, like this:

```
string = string.replace('\\t', sqlexecute.server_info.species.name)
```

and in sqlexecute, instead of defaulting `server_info` to None, have this:

```
self.server_info = ServerInfo.from_version_string('unknown')
```

pasenor: > instead of defaulting server_info to None, have this

server_info is never really None, the code goes like
```Python
def __init__(self, ...):
    self.server_info = None
    self.connect()  # server_info is always set inside connect
```
and initializing it to anything other than None is just a completely wasteful object creation. I guess the only reason it is created in `__init__` is to appease the linter :)

PR #946, opened 2021-01-16
Fix a couple of irregular indentations

rolandwalker:
## Description
Just some irregular indentation fixes found by pylint.

## Checklist
- [ ] ~I've added this contribution to the `changelog.md`.~ no need


pasenor:
Does it make sense, given that we are going to do a full reformatting with black anyway?

pasenor:
Actually, we have such a pile of open PRs that the black commit is probably not going to happen soon, so let's merge this for now)

Issue #950, opened 2021-01-17
Create a standalone binary build


pasenor:
There are several problems with our current deployment through pypi:
1. It relies on the user's system having python3.6+. Those who are stuck with python 2 will also be stuck with an obsolete mycli version.
2. It relies on the user's system to manage dependencies. It can lead to breakages in a lot of ways that we don't control: there may be conflicts with other python applications, or dependencies may be updated in ways that break mycli. Both have happened a lot of times.
3. To address the problems in (2), we need to specify broad ranges of required versions of packages in our dependency list. But our CI always installs the newest versions, and so users may experience issues with older dependencies that we don't notice until they open an issue.

I propose to create standalone binary builds with PyOxidizer (like iredis does) and then possibly switch our main deployment strategy to just shipping these builds. PyPI packages will of course also remain available.

There are a couple of things to do in preparation:
1. Refactor code so it no longer depends on `__file__`. Otherwise, we will need to ship additional text files alongside the binary, which defies the purpose of the trivial deployment.
2. As binaries are platform-specific, we need to set up more OSes in CI. Github Actions now supports Windows and Mac OS instances, so it should be straightforward to cover the most common architectures. I don't know if we should care about weirder ones.

rolandwalker:
This no longer seems like it is an active effort.  Please re-open if that is not the case.

PR #951, opened 2021-01-17
switch to pyaes for password decoding

pasenor:
## Description

removed the Cryptography dependency in favor of [pyaes](https://github.com/ricmoo/pyaes). Cryptography was used in only one place: to decrypt the `mylogin.cnf` file. 

This is a first step towards #950: Cryptography is a pain to package because it relies on a lot of data files. It can be done, but either there will be a large useless memory footprint (those files need to be loaded into ram), or it will require a lot of hacking into the guts of Cryptography to get rid of the parts we don't need. I don't think it's worth the effort. 

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,5 @@
-TODO
-====
+TBD
+===
 
 Bug Fixes:
 ----------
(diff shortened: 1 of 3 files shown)


pasenor:
pyaes is, of course, less audited than Cryptography, if we needed it for something serious there would have been security issues to consider, but encryption of that `mylogin.cnf` file is a joke anyway, we can just take whatever works.

amjith (commented):
What is this `.eggs` folder? Do we need it?

Issue #957, opened 2021-02-18
Exception 1193, "Unknown system variable 'VERSION_COMMENT'"


jeffreytse:
When the MySQL server version is older than `5.0.22`, we will catch the error as below:

![image](https://user-images.githubusercontent.com/9413601/108316452-e4895b80-71f7-11eb-84b7-832b22e6ddc5.png)

MySQL server information:
![image](https://user-images.githubusercontent.com/9413601/108316956-9032ab80-71f8-11eb-9859-bf6fa2f93055.png)

pasenor:
Fixed in #940. Please verify that the current master branch fixes the bug.

pasenor:
released in version 1.24.1

PR #958, opened 2021-02-18
System variables cannot be SELECTed (e.g. @@version_comment)

jeffreytse:
## Description

Exactly, this MySQL bug was fixed on `5.0.22` [Bug #15684](https://bugs.mysql.com/bug.php?id=15684), otherwise we will catch an exception 1193, "Unknown system variable 'VERSION_COMMENT'")

This PR has fixed the issue #957.

## Checklist

### DIFF
--- a/mycli/sqlexecute.py
+++ b/mycli/sqlexecute.py
@@ -26,8 +26,11 @@ class SQLExecute(object):
 
     version_query = '''SELECT @@VERSION'''
 
+    # System variables cannot be SELECTed (e.g. @@version_comment)
+    # this MySQL bug was fixed on `5.0.22`
+    # Bug #15684 https://bugs.mysql.com/bug.php?id=15684
     version_comment_query = '''SELECT @@VERSION_COMMENT'''
-    version_comment_query_mysql4 = '''SHOW VARIABLES LIKE "version_comment"'''
+    version_comment_query_alter = '''SHOW VARIABLES LIKE "version_comment"'''
 
     show_candidates_query = '''SELECT name from mysql.help_topic WHERE name like "SHOW %"'''
 
@@ -280,10 +283,10 @@ class SQLExecute(object):
             _logger.debug('Version Query. sql: %r', self.version_query)
             cur.execute(self.version_query)
             version = cur.fetchone()[0]
-            if version[0] == '4':
+            if float(version[:3]) < 5.1:
                 _logger.debug('Version Comment. sql: %r',
-                              self.version_comment_query_mysql4)
-                cur.execute(self.version_comment_query_mysql4)
+                              self.version_comment_query_alter)
+                cur.execute(self.version_comment_query_alter)
                 version_comment = cur.fetchone()[1].lower()
                 if isinstance(version_comment, bytes):
                     # with python3 this query returns bytes
--- a/mycli/sqlexecute.py
+++ b/mycli/sqlexecute.py
@@ -283,6 +283,16 @@ class SQLExecute(object):
             _logger.debug('Version Query. sql: %r', self.version_query)
             cur.execute(self.version_query)
             version = cur.fetchone()[0]
+
+            # It's better to use the precise version, but we also known,
+            # there are some versions have the alpha suffix, such as
+            # 5.0.18-nt, 5.0.20a and so on, new complexity will be
+            # introduced here, so here we use an imprecise method to check
+            # and determine the way to retrieve the VERSION_COMMENT, and
+            # the alternative query is also valid for the versions like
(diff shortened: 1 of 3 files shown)


pasenor (approved):

jeffreytse (commented):
This commit reaches all the people who meet the exception about the `VERSION_COMMENT`, from now on you guys will feel happy.

pasenor:
You say yourself it was fixed in 5.0.22, but the check is `< 5.1`. Could you use the precise version?

mycli/sqlexecute.py
```diff
@@ -26,8 +26,11 @@ class SQLExecute(object):
 
     version_query = '''SELECT @@VERSION'''
 
+    # System variables cannot be SELECTed (e.g. @@version_comment)
+    # this MySQL bug was fixed on `5.0.22`
+    # Bug #15684 https://bugs.mysql.com/bug.php?id=15684
```
pasenor, line 31: It's good to have this comment, but it would probably be more useful at the line with the version check rather than here.

jeffreytse: Hi @pasenor 

It's better to use the precise version, but we also known, there are some versions have the alpha suffix, such as `5.0.18-nt`, `5.0.20a` and so on, new complexity will be introduced here, so here we use an imprecise method to check and determine the way to retrieve the `VERSION_COMMENT`, and the alternative query is also valid for the versions like `5.0.45b` and so on, only when the version is  `>=5.1`, we use the simple `SELECT @@VERSION_COMMENT`.

Thanks and regards

mycli/sqlexecute.py
```diff
@@ -26,8 +26,11 @@ class SQLExecute(object):
...
     version_comment_query = '''SELECT @@VERSION_COMMENT'''
-    version_comment_query_mysql4 = '''SHOW VARIABLES LIKE "version_comment"'''
+    version_comment_query_alter = '''SHOW VARIABLES LIKE "version_comment"'''
```
j-bennet, line 33: Does the alternative syntax work in all mysql versions, including versions `<5.0.22`? If so, why don't we always use the alternative versions, and remove the version check altogether? Seems to me like it would make things a lot simpler.

jeffreytse: Hi @j-bennet 

It's the best practice to query the version comment by selecting the system variable, that is the statement `SELECT @@VERSION_COMMENT`, but just the history issue the bug caused the problem that we need to use the `SELECT ... LIKE ...`, so nothing but also we can use the alternative statement for all situations.

Thanks and regards

jeffreytse (commented):
I'm an independent Open Source developer, just help to find out a potential and fatal bug on this awesome Open Source Project, this PR has fix all the related issues like (#957), if possible and convenient, also welcome to invite me to be a member of the organization. Any other issue welcome to get back to me, thanks and regards.

mycli/sqlexecute.py
pasenor: > best practice

but why, is there any reason it is preferred?

jeffreytse: Is it `SELECT @@VERSION_COMMENT` better than `SHOW VARABLES LIKE "VERSION_COMMENT"`? Not only for the simpler, but also no the `LIKE` comparison.

pasenor: Simpler yes, but we don't care that much about simplicity of predefined automated queries. And  `LIKE` is not a requirement, you can do `SHOW VARIABLES WHERE Variable_name = 'version_comment'`

pasenor:
@jeffreytse we are leaning towards merging #940. It removes the version queries altogether, in favor of parsing the version string as reported by pymysql. That would also circumvent the bug with `@@version_comment`. I'd be interested in your opinion on that PR too.

mycli/sqlexecute.py
jeffreytse: Yes, and the `SELECT @@VERSION_COMMENT` is also faster than `SHOW VARIABLES WHERE/LIKE ...`, but this totally depends on your choie, it just for your reference, the most important thing is to fix the fatal bug quickly.

jeffreytse:
@pasenor  I'd be happy to discuss with you, hope this can help for the awesome project, and the historical mysql bug I have mention by the code comments. Thanks and regards.

j-bennet:
Closing in favor of https://github.com/dbcli/mycli/pull/940.

Nice work on this fix @jeffreytse, thanks a lot! But I thought that 940 was  a better solution, because it also lets us reduce the number of startup queries. Quicker startup, and less chance to run into a query error.

PR #961, opened 2021-02-19
when opening a file for execution, use the same encoding as in the da…

### LINKED ISSUE #953: 'gbk' codec can't decode byte 0xa3 in position 191: illegal multibyte sequence
When I use mycli on Windows10 to load a sql script, for example, `source XXX.sql`, an error occurred: **'gbk' codec can't decode byte 0xa3 in position 191: illegal multibyte sequence.**
I thought this is a question about opening a file. According to the log, I modified the code in line 255 in `main.py`. Then it works.
image

gfrlv:
…tabase connector

## Description
Fixes #953 

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,7 @@ TBD
 Bug Fixes:
 ----------
 * Allow `FileNotFound` exception for SSH config files.
+* When opening files for execution, use the same encoding as in the database connection
 
 Features:
 ---------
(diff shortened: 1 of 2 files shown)


pasenor:
WIP: connection charsets are not the same thing as file encodings, we need to maintain a mapping between them.

rolandwalker:
> connection charsets are not the same thing as file encodings

Right!  They are really different things.  A user can even have a file they want to read which is in a different encoding than the connection type.  Therefore I think we should close this PR.

We could have `--file-charset` option, but it is better to follow the system defaults for file reads.

rolandwalker:
Closing on the theory that file and connection charsets should stay independent.  Reopen if needed!

Issue #907, opened 2020-11-03
undefined function `_parsed_is_open_quote` in file `packages/parseutils.py`


shispt:
Here is the link: https://github.com/dbcli/mycli/blob/master/mycli/packages/parseutils.py#L234

pasenor:
That's true, thanks for pointing out. But that function doesn't seem to be called, did you get this exception at runtime?

shispt:
Mine runs fine, it's just my IDE prompts the warning.

PR #962, opened 2021-02-19
remove unused function is_open_quote()

pasenor:
## Description
is_open_quote is a pgcli rudiment: there it is used to detect semicolons inside PL/pgSQL function definitions, but mycli handles it with the delimiter command. The function also did not work, see #907 :)

## Checklist

### DIFF
--- a/mycli/clibuffer.py
+++ b/mycli/clibuffer.py
@@ -1,7 +1,6 @@
 from prompt_toolkit.enums import DEFAULT_BUFFER
 from prompt_toolkit.filters import Condition
 from prompt_toolkit.application import get_app
-from .packages.parseutils import is_open_quote
 from .packages import special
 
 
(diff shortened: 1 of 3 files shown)


mycli/clibuffer.py
```diff
@@ -1,7 +1,6 @@
 from prompt_toolkit.enums import DEFAULT_BUFFER
 from prompt_toolkit.filters import Condition
 from prompt_toolkit.application import get_app
-from .packages.parseutils import is_open_quote
```
amjith, line 4: Here's how it is used in pgcli: https://github.com/dbcli/pgcli/blob/master/pgcli/pgbuffer.py#L15

Should we be copying that behavior instead?

pasenor: Yeah, I've looked at that code in pgcli, but its purpose seems to be to facilitate writing stored procedures with semicolons inside them. I don't think that's customary in MySQL, the idiom seems to be 
```
\DELIMITER $$$ 
...
<SQL code with ; replaced by $$$>
...
\DELIMITER ;
```

PR #963, opened 2021-02-21
Ignore the paramiko_stub file while collecting pytests.

amjith:
## Description
Ignore the paramiko_stub file while collecting pytests.


pasenor:
Maybe ignore the whole mycli directory? It would also speed up collection and make it more resilient to future code changes.

pasenor:
Actually, let's merge it as it is, we can fine-tune it later when we do the switch to Makefiles.

pasenor (approved):

PR #964, opened 2021-02-22
Check error code, not message for OperationalError

pasenor:
## Description
We check for ACCESS_DENIED by looking at the error message of the OperationalError, but it can be translated to other languages https://dev.mysql.com/doc/refman/8.0/en/error-message-language.html. It's safer to use the error code. 

## Checklist


j-bennet:
Nice one. Has to be rebased on master.

j-bennet (approved):
LGTM 👍

PR #965, opened 2021-02-22
fix changelog

pasenor:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,7 @@ TBD
 Bug Fixes:
 ----------
 * Allow `FileNotFound` exception for SSH config files.
+* Fix startup error on MySQL < 5.0.22
 
 Features:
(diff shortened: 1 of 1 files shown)


changelog.md
```diff
@@ -59,11 +60,6 @@ Bug Fixes:
 * Avoid opening `/dev/tty` when `--no-warn` is given.
 * Fixed some typo errors in `README.md`.
 
-Internal:
----------
-
-* Use server info from pymysql connection object.
```
j-bennet, line 65: This is not internal, it's a user-facing bug.

j-bennet (changes requested):

changelog.md
pasenor: Right. I wrote that line before I knew of the bug.

PR #970, opened 2021-03-02
Pasenor/pyoxidizer

pasenor:
## Description
Another step towards #950: building and testing the executable, for now only on Linux. It passes the behave suite, but isn't yet fully functional: as I haven't been able to bundle Cryptography, it cannot accept password from the command line, and there is no SSH. Also the file size needs optimization. But we can start using it to see if any other problems crop up.

## Checklist


.github/workflows/ci-standalone.yml
```diff
@@ -0,0 +1,75 @@
...
+
+      - name: Cache cargo registry
+        uses: actions/cache@v1
+        with:
+          path: ~/.cargo/registry
+          key: ${{ runner.os }}-cargo-registry
+
+      - name: Build with pyoxidizer
+        run: |
+          cargo install pyoxidizer
```
laixintao, line 32: just FYI, pyoxidizer can be installed via pip now: https://github.com/laixintao/iredis/pull/384
so we can drop cargo cache.

amjith:
We'll revisit this at a later time.

PR #974, opened 2021-03-31
Fix autocompletion fail for multiple JOINs

### LINKED ISSUE #567: Autocompletion fails for more than one join
If i have a query with more than one join, autocompletion won't show the name/alias of the second joined table, nor its columns.

Example:
```SQL
SELECT * FROM users u
INNER JOIN user_tags ut ON ut.`USER_ID` = u.`USER_ID`
INNER JOIN tags t ON ut.`TAG_ID` = t.`TAG_ID`;
```
In this example, it'd show the names, aliases, and columns of the tables "user_tags" and "users", but not of the table "tags", nor any of the tables that might follow.

`mycli 1.16.0, running on CPython 2.7.12`

afonsocarlos:
## Description
Fix #567 by ignoring the keyword **ON** when extracting SQL tables. 
The next item after ON is still not treated as a table, but query extraction doesn't stop at "ON" anymore (which is useful for multiple JOINs i.e.)

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -43,7 +43,7 @@ Internal:
 Bug Fixes:
 ----------
 * Ensure `--port` is always an int.
- 
+
 1.23.1
 ======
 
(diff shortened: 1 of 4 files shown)


amjith:
Thank you! I'm giggling at the fact you took the effort to alphabetized the contributors list.

Issue #981, opened 2021-04-26
threading.Thread.setDaemon has been deprecated in favor of setting daemon attribute directly in Python 3.10


tirkarthi:
Ref : python/cpython#25174

https://github.com/dbcli/mycli/blob/96e6c951f221d10be27d95fc5fbcdc8ce3710a7b/mycli/completion_refresher.py#L39

tirkarthi:
Fixed in https://github.com/dbcli/mycli/pull/982 . I incorrectly linked the PR to itself.

PR #982, opened 2021-04-26
Set daemon attribute instead of using setDaemon method that was deprecated in Python 3.10

tirkarthi:
## Description

Set daemon attribute instead of using setDaemon method that was deprecated in Python 3.10. This was deprecated since Python 2 but the message was added in Python 3.10

## Checklist

Fixes #982


Narendra-Neerukonda:
Fixes #981

Issue #897, opened 2020-09-04
No longer reading ~/.my.cnf


artfulrobot:
I have a simple `~/.my.cnf` file containing just
```ini
[client]
user=foo
password=secret
```

I've been working with mycli for a few years (thanks!) and my servers that are running v1.21.1 all work ok, but I just installed (`pip3 install mycli`) on a new server which brought in v1.22.2 and that does not appear to read the user's local ~/.my.cnf file which means it asks for a password.

Looking at `~/.mycli.log` it seems it's not looking for `~/.my.cnf`:

```...MainThread) mycli.config ERROR - Unable to parse line None of config file '/etc/mysql/my.cnf'.
...MainThread) mycli.config ERROR - Using successfully parsed config values.
...MainThread) mycli.config ERROR - Unable to parse line None of config file '/etc/mysql/conf.d/mysqldump.cnf'.
...MainThread) mycli.config ERROR - Using successfully parsed config values.
...MainThread) mycli.config ERROR - Unable to parse line None of config file '/etc/mysql/mariadb.conf.d/50-mysqld_safe.cnf'.
...MainThread) mycli.config ERROR - Using successfully parsed config values.
...MainThread) mycli.config ERROR - Unable to parse line None of config file '/etc/mysql/my.cnf'.
...MainThread) mycli.config ERROR - Using successfully parsed config values.
...MainThread) mycli.config ERROR - Unable to parse line None of config file '/etc/mysql/conf.d/mysqldump.cnf'.
...MainThread) mycli.config ERROR - Using successfully parsed config values.
...MainThread) mycli.config ERROR - Unable to parse line None of config file '/etc/mysql/mariadb.conf.d/50-mysqld_safe.cnf'.
...MainThread) mycli.config ERROR - Using successfully parsed config values.
```

Nb. `mysql` starts fine (uses client credentials in `~/.my.cnf`)

ateijelo:
I ran into the same issue. The problem seems to have been introduced by this change: https://github.com/dbcli/mycli/commit/a3cf82a4f34c8f86ea6ebc524e9aac4c08e340a1#diff-8e9c55cacc296c33014d775f5deffcb0aafeac0e0026ffebb9acb60911becc25R321

In particular, removing `mysqld` from the sections there fixes it for me.

[...]

PR #1034, opened 2022-03-02
Ignore the user of the system [mysqld] config

upekkha:
## Description
This is related to issue #897.

Any local `user` configuration gets ignored and overwritten by a `user` defined in the `[mysqld]` section of the global config. For instance Debian/Ubuntu are affected, as the [`/etc/mysql/mariadb.conf.d/50-server.conf`](https://salsa.debian.org/mariadb-team/mariadb-10.5/-/blob/bullseye/debian/additions/mariadb.conf.d/50-server.cnf) constains `user = mysql`. Thus `mycli` fails to connect,

```console
# mycli
Connecting to socket /var/run/mysqld/mysqld.sock, owned by user mysql
(1698, "Access denied for user 'mysql'@'localhost'")
```

unless the user is manually specified with `mycli -u root`. In previous versions this was not the case.

This PR extends the previous #968 and also adds `user` to the `key_transformations` dict, so that it gets ignored.

## Checklist


amjith:
Thank you!

PR #1054, opened 2022-05-15
Fix startup error with some MySQL API compatible databases.

ideal:
e.g. https://github.com/datafuselabs/databend which does not have connection_id().

## Description
Some MySQL compatible databases, may not implemented connection_id() function, on startup, mycli will give error and exit:
```
(1105, 'Code: 2602, displayText = Unknown Function connection_id (while in analyze select projection).')
```

But this is not a fatal error, so we can catch it and continue.

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,9 @@ Internal:
 ---------
 * Pin `cryptography` to suppress `paramiko` warning, helping CI complete and presumably affecting some users.
 
+Bug Fixes:
+----------
+* Support for some MySQL compatible databases, which may not implement connection_id().
 
 1.25.0 (2022/04/02)
 ===================
--- a/mycli/AUTHORS
+++ b/mycli/AUTHORS
@@ -81,6 +81,7 @@ Contributors:
   * xeron
   * Yang Zou
   * Yasuhiro Matsumoto
+  * Yuanchun Shang
   * Zach DeCook
   * Zane C. Bowers-Hadley
   * zer09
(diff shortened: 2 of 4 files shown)


rolandwalker:
Hi!  Shouldn't `self.connection_id` be set to some kind of stub value in the exception?

ideal:
> Hi! Shouldn't `self.connection_id` be set to some kind of stub value in the exception?

Haven't found an appropriate one, which do you suggest? Thanks.

rolandwalker:
@ideal if you `git rebase main` from this branch, the CI issue should be fixed.

ideal:
> @ideal if you `git rebase main` from this branch, the CI issue should be fixed.

OK, thanks.

rolandwalker:
@ideal it seems like you might have cherry-picked my "pin cryptography" commit instead of doing `git rebase main` which is what is needed to make a clean PR.

Also needed to merge:
 * add this contribution to `changelog.md`
 * add your name to the AUTHORS file

ideal:
> @ideal it seems like you might have cherry-picked my "pin cryptography" commit instead of doing `git rebase main` which is what is needed to make a clean PR.
> 

@rolandwalker Sorry, seems that I rebased on wrong remote.

> 
> * add this contribution to `changelog.md`
> * add your name to the AUTHORS file

Updated, thanks.

rolandwalker:
Thanks for this nice contribution.  We should do a release soon.

PR #1068, opened 2022-07-29
Include TiDB in the prompt

dveeden:
## Description

Include TiDB in the prompt

See also: https://github.com/dbcli/mycli/tree/tidb

## Checklist

### DIFF
--- a/mycli/myclirc
+++ b/mycli/myclirc
@@ -66,7 +66,7 @@ wider_completion_menu = False
 # \R - The current time, in 24-hour military time (0-23)
 # \r - The current time, standard 12-hour time (1-12)
 # \s - Seconds of the current time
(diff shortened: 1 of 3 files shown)


amjith:
Thank you! I'll let the CI tests finish and merge it in.

dveeden:
Thanks.

You can also get the keywords from https://github.com/pingcap/tidb/blob/master/parser/parser.y#L73

Maybe use something like this:
```
awk '/The following tokens belong/ { print; kw=1 } /builtinBitAnd/ { exit } { if (kw == 1) { print $2 }}' parser/parser.y
```

And besides the HTML version of the docs there is https://github.com/pingcap/docs/blob/master/keywords.md

PR #1072, opened 2022-08-19
Add cli flag --ssl to enable ssl

meldafert:
## Description
The vanilla mysql client has a `--ssl` flag to enable ssl/tls without any other options.
This makes it possible to connect to servers that do not allow plaintext without needing to specify other options.

I needed this because I was trying to connect to a server with TLSv1.3.
It did not allow plaintext, so it failed when I didn't pass any options.
I tried using `--ssl-cipher=TLSv1.3`, but that failed with `('No cipher can be selected.',)`.
It worked if I downgraded to `--ssl-cipher=TLSv1.2`, however that was using a more insecure cipher than necessary.

It turns out `--ssl-cipher` does not work at all with TLSv1.3. The option is passed on to [SSLContext.set_ciphers()](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.set_ciphers), which notes that "TLS 1.3 cipher suites cannot be disabled with set_ciphers().", explaining the error.

Only after reading the code I figured out that passing any of the `--ssl-...` flags to mycli will enable ssl, so in my case `--ssl-verify-server-cert` would also have worked. However, that is rather unintuitive, and might not work if the server's hostname is not configured correctly. Having a separate flag like the vanilla mysql client does makes more sense here.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,8 @@ $ sudo apt-get install mycli # Only on debian or ubuntu
       --ssh-config-host TEXT        Host to connect to ssh server reading from ssh
                                     configuration.
 
+      --ssl                         Enable SSL for connection (automatically
+                                    enabled with other flags).
       --ssl-ca PATH                 CA file in PEM format.
       --ssl-capath TEXT             CA directory.
       --ssl-cert PATH               X509 cert in PEM format.
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,10 @@
 TBD
 ===
 
+Features:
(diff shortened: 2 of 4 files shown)


meldafert:
It may make sense to add some note about `--ssl-ciphers` not making sense with TLSv1.3 to the cli help. Alternatively, we could also give a more detailed error message if someone tries to use that option with TLSv1.3.

amjith:
Great addition. Thank you!

:+1:

PR #1073, opened 2022-08-19
Add cli option --tls-version to control tls version used

### LINKED ISSUE #1053: Adding a new "TLS version" argument to mycli command line options ?
Hello,

Is it possible to add a new argument to provide a specific version of TLS while connecting to database via mycli command line options ?

Thanks in advance!

Regards,
Pavan.

meldafert:
## Description
Fixes #1053.

This PR adds a `--tls-version` option, parallel to the option of the vanilla mysql client.
It was implemented by bypassing PyMySQL's ssl logic and creating the SSLContext ourselves.

This PR is currently a draft, as it would require raising the minimum python version to 3.7, as 3.6 has no way to force TLSv1.3.
Specifically, `SSLContext.minimum_version` was only added in 3.7.
However, PyMySQL's next update will require 3.7 as well.

This could alternatively be partially implemented inside PyMySQL as well. However, it is a documented option to pass a SSLContext to PyMySQL directly, so this seemed like the easier option.
One downside is that the `_create_ssl_context` method was strongly influenced by the one in PyMySQL - does this need to be noted somewhere for copyright reasons?

## Checklist


mycli/sqlexecute.py
```diff
@@ -354,3 +358,40 @@ def reset_connection_id(self):
...
         self.conn.select_db(db)
         self.dbname = db
+
+    def _create_ssl_ctx(self, sslp):
+        import ssl
+
+        ca = sslp.get("ca")
+        capath = sslp.get("capath")
+        hasnoca = ca is None and capath is None
+        ctx = ssl.create_default_context(cafile=ca, capath=capath)
```
amjith, line 368: If the values for cafile or capath are None, is it still safe to pass them into this function?

mycli/sqlexecute.py
```diff
@@ -354,3 +358,40 @@ def reset_connection_id(self):
...
+                ctx.minimum_version = ssl.TLSVersion.TLSv1_1
+                ctx.maximum_version = ssl.TLSVersion.TLSv1_1
+            elif tls_version == "TLSv1.2":
+                ctx.minimum_version = ssl.TLSVersion.TLSv1_2
+                ctx.maximum_version = ssl.TLSVersion.TLSv1_2
+            elif tls_version == "TLSv1.3":
+                ctx.minimum_version = ssl.TLSVersion.TLSv1_3
+                ctx.maximum_version = ssl.TLSVersion.TLSv1_3
+            else:
+                _logger.error('Invalid tls version: %s', tls_version)
```
amjith, line 395: Should this else section raise an exception so the user knows they've passed in an invalid version?

mycli/sqlexecute.py
```diff
@@ -354,3 +358,40 @@ def reset_connection_id(self):
     def change_db(self, db):
         self.conn.select_db(db)
         self.dbname = db
+
+    def _create_ssl_ctx(self, sslp):
+        import ssl
+
+        ca = sslp.get("ca")
+        capath = sslp.get("capath")
+        hasnoca = ca is None and capath is None
```
amjith, line 367: Should this use an `or` condition instead of `and` ?

amjith (commented):
I like the feature. I don't personally use ssl context so pardon my questions if they're too basic. I've left some questions inline for own clarification.

mycli/sqlexecute.py
meldafert: Yes, they are optional and default to `None`: https://docs.python.org/3/library/ssl.html#ssl.create_default_context

mycli/sqlexecute.py
meldafert: This should never happen when using the cli, since it is checked by click:
https://github.com/dbcli/mycli/pull/1073/files#diff-8e9c55cacc296c33014d775f5deffcb0aafeac0e0026ffebb9acb60911becc25R1101
If anything other than those four values is passed, it will error at that point already.

mycli/sqlexecute.py
meldafert: ca and capath don't need to be combined - if either of them is given, a custom certificate is being used. So only if neither are passed we use the system certificates (and don't verify).
As I said, the logic in this is taken from PyMySQL's [Connection._create_ssl_context](https://github.com/PyMySQL/PyMySQL/blob/dd47caae95011e79b9e2ee12549d23f05a7f839d/pymysql/connections.py#L361), so this is already how it worked before my changes.
I could try to contribute the changes inside `_create_ssl_context` to PyMySQL instead, but since it allows passing a SSLContext directly, this seemed easier.

dveeden:
Might be good to consider lists/ranges of TLS versions.

- `--tls-version TLSv1.2,TLSv1.3` Allowing these two versions only
- `--min-tls-version TLSv1.2` Allowing TLSv1.2 and up

Might be good to by default only allow TLSv1.2 and up. See [RFC 8996](https://www.rfc-editor.org/rfc/rfc8996.html)

meldafert:
Thanks!

PR #1075, opened 2022-08-28
Prettify / unprettify current statement using sqlglot

rolandwalker:
## Description
Using https://github.com/tobymao/sqlglot, prettify the current statement, almost always on to multiple lines, or un-prettify the statement, collapsing onto a single line.

This is something I would use *all the time*!  I knew how to take it this far.  Some things I didn't know how to do:
 * ~give an informative feedback message on failure~ Solved
 * ~create vi bindings~ Solved
 * ~test coverage (got lost here on my most recent PR and gave up)~ Solved

~There's no documentation yet pending approval of this direction.~ Solved

If there is more than one statement in the buffer, we neither prettify nor unprettify but ~give a beep.~ give a transient message in the toolbar.

There is some attempt to find an appropriate cursor position in the reformatted buffer but of course that is difficult to do perfectly.  One would need the cooperation of the parser.

This introduces a dependency on sqlglot, when we already have a dependency on sqlparse.   But sqlglot is a more capable reader and is being more actively maintained.  We should not be bothered by the dependency; we should look into shifting more functionality to sqlglot.

Edit: In order to bring in the sqlglot dependency, we need to drop support for Python 3.6, per discussion below.  We should also add 3.10 to CI in a separate PR.

Demo

https://user-images.githubusercontent.com/727482/187200145-4b2d745a-dc38-469a-8323-276e820a268a.mov

## Checklist

### DIFF
--- a/mycli/key_bindings.py
+++ b/mycli/key_bindings.py
@@ -1,6 +1,6 @@
 import logging
 from prompt_toolkit.enums import EditingMode
-from prompt_toolkit.filters import completion_is_selected
+from prompt_toolkit.filters import completion_is_selected, emacs_mode, vi_mode
 from prompt_toolkit.key_binding import KeyBindings
 
 _logger = logging.getLogger(__name__)
@@ -60,6 +60,48 @@ def mycli_bindings(mycli):
             b.complete_next()
         else:
             b.start_completion(select_first=False)
+
+    @kb.add('>', filter=vi_mode)
+    @kb.add('c-x', 'p', filter=emacs_mode)
+    def _(event):
+        """
+        Prettify and indent current statement, usually into multiple lines.
+
+        Only accepts buffers containing single SQL statements.
+        """
+        _logger.debug('Detected <C-x p>/> key.')
+
+        b = event.app.current_buffer
+        cursorpos_relative = b.cursor_position / len(b.text)
+        pretty_text = mycli.handle_prettify_binding(b.text)
+        if len(pretty_text) > 0:
+            b.text = pretty_text
+            cursorpos_abs = int(round(cursorpos_relative * len(b.text)))
+            while 0 < cursorpos_abs < len(b.text) \
+                  and b.text[cursorpos_abs] in (' ', '\n'):
+                cursorpos_abs -= 1
+            b.cursor_position = min(cursorpos_abs, len(b.text))
+
+    @kb.add('<', filter=vi_mode)
+    @kb.add('c-x', 'u', filter=emacs_mode)
+    def _(event):
+        """
+        Unprettify and dedent current statement, usually into one line.
+
+        Only accepts buffers containing single SQL statements.
+        """
+        _logger.debug('Detected <C-x u>/< key.')
+
+        b = event.app.current_buffer
+        cursorpos_relative = b.cursor_position / len(b.text)
+        unpretty_text = mycli.handle_unprettify_binding(b.text)
+        if len(unpretty_text) > 0:
+            b.text = unpretty_text
+            cursorpos_abs = int(round(cursorpos_relative * len(b.text)))
+            while 0 < cursorpos_abs < len(b.text) \
+                  and b.text[cursorpos_abs] in (' ', '\n'):
+                cursorpos_abs -= 1
+            b.cursor_position = min(cursorpos_abs, len(b.text))
 
     @kb.add('enter', filter=completion_is_selected)
     def _(event):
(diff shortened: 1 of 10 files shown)


rolandwalker:
CI is failing on Python 3.6 since it doesn't have `dataclasses`.  Looks like there is a backport.

Tagging @amjith for general opinions before I put any more time in.

mycli/main.py
```diff
@@ -582,6 +583,34 @@ def handle_clip_command(self, text):
             return True
         return False
 
+    def handle_prettify_binding(self, text):
+        try:
+            statements = sqlglot.parse(text)
```
tobymao, line 588: i don't know much about this library, but i assume it's mysql based, so you'll want to specify read="mysql"

and when you call sql(), write="mysql"

mycli/key_bindings.py
```diff
@@ -61,6 +61,45 @@ def _(event):
...
+        pretty_text = mycli.handle_prettify_binding(b.text)
+        if len(pretty_text) > 0:
+            b.text = pretty_text
+            cursorpos_abs = int(round(cursorpos_relative * len(b.text)))
+            while 0 < cursorpos_abs < len(b.text) \
+                  and b.text[cursorpos_abs] in (' ', '\n'):
+                cursorpos_abs -= 1
+            b.cursor_position = min(cursorpos_abs, len(b.text))
+        # todo
+        # * informative feedback message on failure
```
amjith, line 82: We could use the CLI toolbar at the bottom to show the error message. Here's an example: 

https://github.com/dbcli/mycli/blob/main/mycli/clitoolbar.py#L35

mycli/main.py
```diff
@@ -582,6 +583,34 @@ def handle_clip_command(self, text):
...
+        except:
+            statements = []
+        if len(statements) == 1:
+            pretty_text = statements[0].sql(pretty=True, pad=4)
+        else:
+            pretty_text = ''
+            self.bell()
```
amjith, line 595: Dinging a bell for this seems wee bit intrusive. I can be convinced otherwise.

amjith: Perhaps we can add a config option to ding the bell or not and read that in the self.bell() function so it will be centralized?

amjith (commented):
I've left some comments inline.

amjith:
Looks like Python 3.6 has already been EOL'ed. I'm happy to drop support for 3.6 since it is breaking the dataclass dependency.

mycli/main.py
rolandwalker: I agree it is too intrusive.  It was just the only thing I could figure out in a draft.  A message would be better.

mycli/main.py
rolandwalker: Hey thanks!  Added, substituting `dialect="mysql"` for `write=`.

Although actually `read="mysql"` makes `parse()` fail on the example I was using, simplified:

```
python -c 'import sqlglot; print(sqlglot.parse_one("""SELECT "2021-01-01" + interval 1 month""", read="mysql").sql(dialect="mysql"))'
…
sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 1, Col: 21.
  SELECT "2021-01-01" + interval 1 month
  ```

`python-dateutil` is installed if that matters.  Should I file an issue?

Comment (tobymao) [reply mycli/main.py]: yea you can file an issue, it might be because you're using double quotes around the date

File: mycli/key_bindings.py (hunk 2 above)
Comment (rolandwalker) [reply mycli/key_bindings.py]: resolved

File: mycli/main.py (hunk 3 above)
Comment (rolandwalker) [reply mycli/main.py]: resolved

File: setup.py [hunk 4]
```diff
@@ -26,6 +26,8 @@
     'prompt_toolkit>=3.0.6,<4.0.0',
     'PyMySQL >= 0.9.2',
     'sqlparse>=0.3.0,<0.5.0',
+    'sqlglot>=5.1.3',
+    'python-dateutil>=2.8.2',
```
amjith, line 30: What is the need for the `python-dateutil` package?

amjith (commented):
I didn't understand the python-dateutil requirement but otherwise it looks good to me.

setup.py
tobymao: sqlglot does not require it, although it's an option dependency for the optimizer. it's not really needed for a cli tool

rolandwalker: Thanks, removed.

amjith:
Nicely done! :1st_place_medal:

amjith:
@tobymao  Thank you for your help in reviewing this PR. Your library sqlglot looks very interesting! I like the duckdb examples :smile:

PR #1089, opened 2022-09-17
Remove vi-mode bindings for prettify/unprettify functionality

### LINKED ISSUE #1087: In vi editing mode '<' and '>' are not inserted
mycli version 1.26.1 MacOS 12.5 installed via homebrew

In vi editing mode I can't insert < and > characters.  Also hitting one of them as the first character of a line causes a divide by zero error.

I fixed this by disabling the following 2 lines in key_bindings.py:

    --- key_bindings.py~	2022-09-01 14:50:41.000000000 -0600
    +++ key_bindings.py	2022-09-15 12:58:25.000000000 -0600
    @@ -61,7 +61,7 @@
             else:
                 b.start_completion(select_first=False)
     
    -    @kb.add('>', filter=vi_mode)
    +    # @kb.add('>', filter=vi_mode)
         @kb.add('c-x', 'p', filter=emacs_mode)
         def _(event):
             """
    @@ -82,7 +82,7 @@
                     cursorpos_abs -= 1
                 b.cursor_position = min(cursorpos_abs, len(b.text))
     
    -    @kb.add('<', filter=vi_mode)
    +    # @kb.add('<', filter=vi_mode)
         @kb.add('c-x', 'u', filter=emacs_mode)
         def _(event):
             """
If editing mode is emacs, there is no problem.

rolandwalker:
## Description

Fixes #1087 .

I thought this implementation was correct, but I don't use vi-mode, and we have an issue from a user who does.  Best to remove the vi bindings.  Perhaps someone who uses vi-mode can PR better bindings.

## Checklist


mycli/key_bindings.py
```diff
@@ -61,7 +61,6 @@ def _(event):
         else:
             b.start_completion(select_first=False)
 
-    @kb.add('>', filter=vi_mode)
```
brianphillips, line 64: I think the filter you want here is `vi_navigation_mode`. And then I would suggest using `]` here and `[` down below (since `>` and `<` are already bound by the default VI key bindings as part of `>>` indent and `<<` de-indent commands)

PR #1114, opened 2023-05-04
Correct version parsing of TiDB

dveeden:
## Description

Update the version reporting for TiDB.

Examples:
- `5.7.25-TiDB-v7.0.0` (v7.0.0 release)
- `8.0.11-TiDB-v7.2.0-alpha-69-g96e9e68daa` (build from git branch)

Without this change this would parse as `5.7.25` and `8.0.11`.

With this change this parses as `7.0.0` and `7.2.0`.

The first version in the version string (5.7.25, 8.0.11) is the version of MySQL that TiDB claims to be compatible with. The second version is the actual version of TiDB.

## Checklist


amjith:
Thank you!

Issue #1144, opened 2024-01-15
optimize the default cli option `--prompt` with splitline for long prompt


hzc989:
## Background

a default prompt (`\\t \\u@\\h:\\d> `) is already defined inside MyCli class. however it does not work properly when the total length exceeds 45 characters.

## how to optimize it

simply define another default prompt with splitline 

I made a PR which closes this issue

rolandwalker:
Resolved by #1145

PR #1145, opened 2024-01-15
fix(main.py): makes default prompt correctly

hzc989:
## Description

default prompt option not working when exceeds max limit, which is 45 chars.

in this circumstance, only the databasename is show, which is a bit inconvenient and unexpected for information display.

image

## Optimization PR

I made a PR to optimize it.

as image shown below, add a default prompt with `\n` for line splitting, when it exceeds the max length.

image

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,7 @@ Upcoming
 Bug Fixes:
 ----------
 * Don't install tests.
+* let the `--prompt` option act normally with its predefined default value
 
 1.27.0 (2023/08/11)
 ===================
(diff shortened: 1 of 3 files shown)


hzc989:
this fixes the issue https://github.com/dbcli/mycli/issues/1144

amjith:
Thank you!

your-diary:
@hzc989 @amjith 
Can you consider reverting this PR?
I think this PR is not tested well; it breaks an existing important feature.
See #1155.

PR #1148, opened 2024-02-19
tests now work or are skipped on Windows.

ERYoung11:
## Description

I updated the tests so that they now run on Windows.  Some techniques employed:
- NamedTemporaryFile is problematic on Windows because it keeps the file locked.  Setting delete=False means it isn't locked but it must be cleaned up manually.
- several checks with EOL = \n.  Now check  \r\n for Windows only.
- a few things skipped that just don't work quite right on Windows for now.
    - favorite queries - something is not working with this code.  I need to research more to understand why it breaks but it complains about the item used to make a Singleton in the class.
    - watch_query - apparently watching with interrupts just isn't working and this is somewhat known for Windows. Tends to loop more times than expected.  Need to investigate to see if we can get it to work at all.
    - 
## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,6 @@
 .cache/
 .coverage
 .coverage.*
+
+.venv/
+venv/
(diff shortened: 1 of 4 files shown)


amjith:
Thanks a lot for the improvements.

Issue #1001, opened 2021-08-19
CLI argument for socket path is ignored unless port is specified


weirdan:
Version: 1.24.1

### Steps to reproduce
Connect via a socket (with a non-default path), e.g.
```console
$ MYSQL_PWD="$(gcloud auth print-access-token)" mycli  -u "username" -S ~/.local/run/cloud-sql/connection-name
```

### Expected

`mycli` can connect

### Actual

```console
$ MYSQL_PWD="$(gcloud auth print-access-token)" mycli  -u "username" -S ~/.local/run/cloud-sql/connection-name
Connecting to socket /var/run/mysqld/mysqld.sock, owned by user mysql
(1698, "Access denied for user 'username'@'localhost'")
```
Note that the path to socket specified via cli argument is ignored and default path is used.

### Workaround

If I specify the port, it works, even though port makes no sense for socket connections:
```console
$ MYSQL_PWD="$(gcloud auth print-access-token)" mycli  -u "username" -P1 -S ~/.local/run/cloud-sql/connection-name
Connecting to socket /home/weirdan/.local/run/cloud-sql/connection-name, owned by user weirdan
MySQL username@(none):(none)>
```

### Cause

https://github.com/dbcli/mycli/blob/d62eefdc819a11ecdb97d93dd7ad1922d28a3795/mycli/main.py#L420-L428

chriscroome:
On Debian Buster, using the Debian packaged version:

```bash
mycli --version
Version: 1.23.2
```
The only work-around for this error:

```bash
mycli 
Connecting to socket /var/run/mysqld/mysqld.sock, owned by user mysql
(1698, "Access denied for user 'mysql'@'localhost'")
```
That I have found that works, is to create a `~/.my.cnf` like this:

```conf
[client]
user="foo"
password="SOME_LONG_RANDOM_STRING"
socket="/run/mysqld/mysqld.sock"
```

And a `~/.myclirc` like this:

```conf
[alias_dsn]
bar="mysql://foo@localhost/foo"
baz="mysql://foo@localhost/baz"
```

And then connect to the databases like this:

```bash
mycli -d bar
mycli -d baz
```

However this isn't ideal as running the client without arguments still generates an error:

```bash
mycli
Connecting to socket /var/run/mysqld/mysqld.sock, owned by user mysql
(1698, "Access denied for user 'mysql'@'localhost'")
```

[...]

PR #1151, opened 2024-03-23
Do not ignore the socket passed with the -S option, even when no port…

fabzzap:
… is passed

## Description
When `mycli -S <socket>`  is used, mycli tries to connect to localhost anyway instead of using the provided socket. The problem seems to be there since 51b8ee3c.

The problem has been reported as #1001 

## Checklist


amjith:
Looks good. Thanks!

PR #1152, opened 2024-03-26
Tests: monkeypatch shutil.get_terminal_size properly

hroncok:
The original way led to very confusing internal pytest errors when the test failed.

When the test failed, the original function was not restored, and pytest uses it:

    INTERNALERROR> Traceback (most recent call last):
    INTERNALERROR>   File "/usr/lib/python3.13/site-packages/_pytest/main.py", line 271, in wrap_session
    INTERNALERROR>     session.exitstatus = doit(config, session) or 0
    INTERNALERROR>                          ~~~~^^^^^^^^^^^^^^^^^
    INTERNALERROR>   File "/usr/lib/python3.13/site-packages/_pytest/main.py", line 325, in _main
    INTERNALERROR>     config.hook.pytest_runtestloop(session=session)
    INTERNALERROR>     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
    INTERNALERROR>   File "/usr/lib/python3.13/site-packages/pluggy/_hooks.py", line 493, in __call__
    INTERNALERROR>     return self._hookexec(self.name, self._hookimpls, kwargs, firstresult)
    INTERNALERROR>            ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    INTERNALERROR>   File "/usr/lib/python3.13/site-packages/pluggy/_manager.py", line 115, in _hookexec
    INTERNALERROR>     return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
    INTERNALERROR>            ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    INTERNALERROR>   File "/usr/lib/python3.13/site-packages/pluggy/_callers.py", line 152, in _multicall
    INTERNALERROR>     return outcome.get_result()
    INTERNALERROR>            ~~~~~~~~~~~~~~~~~~^^
    INTERNALERROR>   File "/usr/lib/python3.13/site-packages/pluggy/_result.py", line 114, in get_result
    INTERNALERROR>     raise exc.with_traceback(exc.__traceback__)
    INTERNALERROR>   File "/usr/lib/python3.13/site-packages/pluggy/_callers.py", line 77, in _multicall
    INTERNALERROR>     res = hook_impl.function(*args)
[...]

PR #1153, opened 2024-03-29
Tests: Use monkeypatch.context to avoid side effects on pytest

parona-source:
* Otherwise this mocked attribute would affect pytest negatively when running tests with --verbose.

See-Also: https://github.com/dbcli/mycli/pull/1152#issuecomment-2027036256

## Description

https://github.com/dbcli/mycli/pull/1152 had a side effect of breaking pytest when invoked with --verbose, this pr use monkeypatch.context so that setattr doesnt affect pytest itself.

## Checklist

### DIFF
--- a/test/test_main.py
+++ b/test/test_main.py
@@ -259,9 +259,10 @@ def test_reserved_space_is_integer(monkeypatch):
     def stub_terminal_size():
         return (5, 5)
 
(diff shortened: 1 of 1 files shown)


parona-source:
Test fixes aren't included in the changelog I presume?

Don't see any explicit note about what contribution should be listed in the changelog but it appears test fixes aren't included.

PR #1154, opened 2024-03-29
Fixup CI

parona-source:
## Description
Fixes CI so that it runs and passes.

See https://github.com/dbcli/mycli/pull/1153

## Checklist


parona-source:
A future task would be to move to PEP517 and stop using setup.py.

PR #1165, opened 2024-06-13
Add better sorting of completions

schmeic:
## Description
Sort completions by match point, then match length, then item text.

Note that match length is equal to the length of the search text for non fuzzy matches, and equal to the length of the match group for fuzzy matches.

This means that matches that start with the search text will be displayed first, followed by matches that contain the search text, and then fuzzy matches.  

Sorting by item text when both match point and match length are the same for matched completions ensures that completions that start with the search text will be in alphabetical order. For example, searching for _fr_ will display _FROM_ before _FROM_UNIXTIME_.

## Checklist

### DIFF
--- a/mycli/sqlcompleter.py
+++ b/mycli/sqlcompleter.py
@@ -370,6 +370,9 @@ class SQLCompleter(Completer):
         in the collection of available completions.
         """
 
+        if not text:
+            casing = None 
+        
         if casing == 'auto':
             casing = 'lower' if text and text[-1].islower() else 'upper'
 
--- a/mycli/sqlcompleter.py
+++ b/mycli/sqlcompleter.py
@@ -372,7 +372,17 @@ class SQLCompleter(Completer):
         last = last_word(text, include='most_punctuations')
         text = last.lower()
 
-        completions = []
+        if casing == 'auto':
+            casing = 'lower' if last and last[-1].islower() else 'upper'
+
+        def apply_case(kw):
+            if casing is None:
+                return kw
+            if casing == 'upper':
+                return kw.upper()
+            return kw.lower()
+
+        matches = []
 
         if fuzzy:
             regex = '.*?'.join(map(escape, text))
@@ -380,37 +390,37 @@ class SQLCompleter(Completer):
             for item in collection:
                 r = pat.search(item.lower())
                 if r:
-                    completions.append((len(r.group()), r.start(), item))
+                    matches.append(
+                        (len(r.group()), r.start(), apply_case(item)))
         else:
             match_end_limit = len(text) if start_only else None
             for item in collection:
                 match_point = item.lower().find(text, 0, match_end_limit)
                 if match_point >= 0:
-                    completions.append((len(text), match_point, item))
+                    matches.append((len(text), match_point, apply_case(item)))
 
-        if casing == 'auto':
-            casing = 'lower' if last and last[-1].islower() else 'upper'
-
-        def apply_case(kw):
-            if casing == 'upper':
-                return kw.upper()
-            return kw.lower()
-
-        return (Completion(z if casing is None else apply_case(z), -len(text))
-                for x, y, z in completions)
+        return matches
 
     def get_completions(self, document, complete_event, smart_completion=None):
         word_before_cursor = document.get_word_before_cursor(WORD=True)
+
+        def sorted_completions(matches):
+            # sort by match point, then match length, then item text
+            matches = sorted(matches, key=lambda m: (m[1], m[0], m[2].lower()))
+            return (Completion(z, -len(word_before_cursor))
+                    for x, y, z in matches)
+
         if smart_completion is None:
             smart_completion = self.smart_completion
 
         # If smart_completion is off then match any word that starts with
         # 'word_before_cursor'.
         if not smart_completion:
-            return self.find_matches(word_before_cursor, self.all_completions,
-                                     start_only=True, fuzzy=False)
+            matches = self.find_matches(word_before_cursor, self.all_completions,
+                                        start_only=True, fuzzy=False)
+            return sorted_completions(matches)
 
-        completions = []
+        matches = []
         suggestions = suggest_type(document.text, document.text_before_cursor)
 
         for suggestion in suggestions:
@@ -431,14 +441,14 @@ class SQLCompleter(Completer):
                     ]
 
                 cols = self.find_matches(word_before_cursor, scoped_cols)
-                completions.extend(cols)
+                matches.extend(cols)
 
             elif suggestion['type'] == 'function':
                 # suggest user-defined functions using substring matching
                 funcs = self.populate_schema_objects(suggestion['schema'],
                                                      'functions')
                 user_funcs = self.find_matches(word_before_cursor, funcs)
-                completions.extend(user_funcs)
+                matches.extend(user_funcs)
 
                 # suggest hardcoded functions using startswith matching only if
                 # there is no schema qualifier. If a schema qualifier is
@@ -450,35 +460,35 @@ class SQLCompleter(Completer):
                                                          start_only=True,
                                                          fuzzy=False,
                                                          casing=self.keyword_casing)
-                    completions.extend(predefined_funcs)
+                    matches.extend(predefined_funcs)
 
             elif suggestion['type'] == 'table':
                 tables = self.populate_schema_objects(suggestion['schema'],
                                                       'tables')
                 tables = self.find_matches(word_before_cursor, tables)
-                completions.extend(tables)
+                matches.extend(tables)
 
             elif suggestion['type'] == 'view':
                 views = self.populate_schema_objects(suggestion['schema'],
                                                      'views')
                 views = self.find_matches(word_before_cursor, views)
-                completions.extend(views)
+                matches.extend(views)
 
             elif suggestion['type'] == 'alias':
                 aliases = suggestion['aliases']
                 aliases = self.find_matches(word_before_cursor, aliases)
-                completions.extend(aliases)
+                matches.extend(aliases)
 
             elif suggestion['type'] == 'database':
                 dbs = self.find_matches(word_before_cursor, self.databases)
-                completions.extend(dbs)
+                matches.extend(dbs)
 
             elif suggestion['type'] == 'keyword':
                 keywords = self.find_matches(word_before_cursor, self.keywords,
                                              start_only=True,
                                              fuzzy=False,
                                              casing=self.keyword_casing)
-                completions.extend(keywords)
+                matches.extend(keywords)
 
             elif suggestion['type'] == 'show':
                 show_items = self.find_matches(word_before_cursor,
@@ -486,41 +496,41 @@ class SQLCompleter(Completer):
                                                start_only=False,
                                                fuzzy=True,
                                                casing=self.keyword_casing)
-                completions.extend(show_items)
+                matches.extend(show_items)
 
             elif suggestion['type'] == 'change':
                 change_items = self.find_matches(word_before_cursor,
                                                  self.change_items,
                                                  start_only=False,
                                                  fuzzy=True)
-                completions.extend(change_items)
+                matches.extend(change_items)
             elif suggestion['type'] == 'user':
                 users = self.find_matches(word_before_cursor, self.users,
                                           start_only=False,
                                           fuzzy=True)
-                completions.extend(users)
+                matches.extend(users)
 
             elif suggestion['type'] == 'special':
                 special = self.find_matches(word_before_cursor,
                                             self.special_commands,
                                             start_only=True,
                                             fuzzy=False)
-                completions.extend(special)
+                matches.extend(special)
             elif suggestion['type'] == 'favoritequery':
                 queries = self.find_matches(word_before_cursor,
                                             FavoriteQueries.instance.list(),
                                             start_only=False, fuzzy=True)
-                completions.extend(queries)
+                matches.extend(queries)
             elif suggestion['type'] == 'table_format':
                 formats = self.find_matches(word_before_cursor,
                                             self.table_formats,
                                             start_only=True, fuzzy=False)
-                completions.extend(formats)
+                matches.extend(formats)
             elif suggestion['type'] == 'file_name':
                 file_names = self.find_files(word_before_cursor)
-                completions.extend(file_names)
+                matches.extend(file_names)
 
-        return completions
+        return sorted_completions(matches)
 
     def find_files(self, word):
         """Yield matching directory or file names.
--- a/mycli/sqlcompleter.py
+++ b/mycli/sqlcompleter.py
@@ -371,7 +371,9 @@ class SQLCompleter(Completer):
         """
 
         if casing == 'auto':
-            casing = 'lower' if last and last[-1].islower() else 'upper'
+            casing = 'lower' if text and text[-1].islower() else 'upper'
+
+        text = text.lower()
 
         def apply_case(kw):
             if casing is None:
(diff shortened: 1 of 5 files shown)


rolandwalker:
Wow, you weren't joking about the tests.

From the English description, the algorithm sounds great.

I may have time to experiment with this on the weekend.

schmeic:
Yeah, that's why I wanted to make sure the algorithm would be accepted before working on the tests. From my experiments with this new sorting, I've found that it has always put what I consider the most relevant results at the top.

The funny thing was that the find_matches(...) function was already storing all 3 of the parameters I used for sorting, and then  ignoring everything but the item text at the end when creating the completions generator. That's what gave me the idea for the sorting algorithm.

rolandwalker:
Let's let the PR hang around for some days and see if it attracts any objections.

schmeic:
Ok. Maybe you'll have some time to experiment with it as well. I'm hoping it will attract some "likes"...

rolandwalker:
Should the proposed change only apply when `if not smart_completion:` ?  It would see that the point of `smart_completion` is to give categories in category order.

mycli/sqlcompleter.py
```diff
@@ -372,45 +372,55 @@ def find_matches(text, collection, start_only=False, fuzzy=True, casing=None):
...
-        return (Completion(z if casing is None else apply_case(z), -len(text))
-                for x, y, z in completions)
+        return matches
 
     def get_completions(self, document, complete_event, smart_completion=None):
         word_before_cursor = document.get_word_before_cursor(WORD=True)
+
+        def sorted_completions(matches):
+            # sort by match point, then match length, then item text
+            matches = sorted(matches, key=lambda m: (m[1], m[0], m[2].lower()))
```
rolandwalker, line 409: As long as the sort is casefolding it might as well also do

```
m[2].lower().strip('`')
```

schmeic:
I don't think so, I personally would rather have a completion at the top that best matches my search text regardless of the category.  Have to tried it out yet?  Maybe you will agree after some experimenting.

mycli/sqlcompleter.py
schmeic: Yeah, I can certainly add that. Pretty sure I've seen someone else complain in another issue about the back tick messing with sorting.

rolandwalker:
I'm trying it now.  It would seem that the config file lies when it claims

```
# Enables context sensitive auto-completion. If this is disabled then all
# possible completions will be listed.
smart_completion = True
```

because when set to `False`, table names and such are simply not returned.

It _is_ rather helpful to see the completions in the "smart" ordering when nothing has yet been typed, like maybe

```
            if len(word_before_cursor) != 0 and word_before_cursor[-1] != '(':
                matches = sorted(matches, key=lambda m: (m[1], m[0], m[2].lower().strip('`')))
```

schmeic:
> I'm trying it now. It would seem that the config file lies when it claims
> 
> ```
> # Enables context sensitive auto-completion. If this is disabled then all
> # possible completions will be listed.
> smart_completion = True
> ```
> 
> because when set to `False`, table names and such are simply not returned.
> 
> It _is_ rather helpful to see the completions in the "smart" ordering when nothing has yet been typed, like maybe
> 
> ```
>             if len(word_before_cursor) != 0 and word_before_cursor[-1] != '(':
>                 matches = sorted(matches, key=lambda m: (m[1], m[0], m[2].lower().strip('`')))
> ```

schmeic:
Regarding the `smart_completion = False` setting, I think what you're seeing is a bug. You need to first do:

`use <db name>`

and then you should see table and column names. It doesn't work if you use a DSN to select the db.

rolandwalker:
Why closed?

schmeic:
Sorry, didn't realize I did that. I deleted a comment and maybe that closed it?

How can I reopen?

schmeic:
Accidentally closed.

rolandwalker:
> Regarding the smart_completion = False setting, I think what you're seeing is a bug. You need to first do:
> use <db name>
> and then you should see table and column names. It doesn't work if you use a DSN to select the db.

Your suggestion works!  But we really need to fix that bug.

> By the way, I think it would be really useful to do fuzzy matching when not doing smart completion. Maybe this can be implemented as a setting?

Yes, though one issue is that I adore settings, and @amjith really does not.  For example for the entire functionality in this PR, I think it would be fantastic if `~/.myclirc` added

```
completion_aggresive_sort = True
```

or any other property name that makes sense.

> I personally would rather have a completion at the top that best matches my search text regardless of the category. 

Right.  I _personally_ would also prefer to use your algorithm.  But we have to think of the most general usecase and the default expectation.  That is why I proposed to enable it outside of the `smart_completion` codepath.

amjith:
Let me catch up on the PR. 

I'm not opposed to adding a config option but if the new completion behavior feels logical we can switch to it by default. Let me read through the proposal and try it out locally.

schmeic:
I think this new sort is much better than the current default, so I can't see anyone preferring the current method. Also, there are other open issues I've seen where people complain about the fuzzy matching and want an option to turn it off because the fuzzy matches are often at the top of the results. My algorithm will also fix this since the fuzzy matches will always be below exact matches.

amjith:
I'm still testing things out. Notes so far, I like that `FROM` is listed about `FROM_UNIXTIME`. There is a crash that happens when I type `SOURCE `. Haven't dug into it yet, if you have the time please take a look.

schmeic:
> I'm still testing things out. Notes so far, I like that `FROM` is listed about `FROM_UNIXTIME`. There is a crash that happens when I type `SOURCE `. Haven't dug into it yet, if you have the time please take a look.

Ok, I see the issue. The find_files() method needs to return an array of matches not a Completion generator.

Any idea what happened to the sorting in the latest version (1.27.1) ?  Why is it now alphabetical?

schmeic:
Still waiting on feedback for this one...

amjith:
I've tested it locally and I'm fine with the ordering. Happy to merge if you can fix the tests.

schmeic:
Cool, I'll work on fixing the tests.

schmeic:
Is there a way to run tests individually?  I'm having a bit of trouble understanding the test setup.

amjith:
We just use `pytest`. You can run `pytest -k <name of the test>` to run just a single test. 

For example: 

`pytest -k test_suggested_multiple_column_names_with_dot` will run only that one single test. 

You can also specify partial names to select multiple tests. 

`pytest -k test_suggested` will select all the tests that start with that name.

schmeic:
Cool, thanks!  I don't use python that much, so this is all pretty new to me.  First test fixed...

schmeic:
I've fixed all of the unit tests, except when I run `setup.py test` I have one fail. 

Running the same test like this works:
`pytest -k test_empty_string_completion test/test_smart_completion_public_schema_only.py`

Any idea what's causing this? Or how I can get more details when running the setup test?

schmeic:
Looks like the error is related to the test command, which is not directly related to this pull request.

_"The test command is disabled and references to it are deprecated."_

What's the status of this pull request?

rolandwalker:
What happens if you change the line `setuptools` in `requirements-dev.txt` to `setuptools<=71.1.0` ?

schmeic:
Better, but why are most of these checks getting cancelled?  Also, the one that failed had no errors:

8 features passed, 0 failed, 0 skipped
32 scenarios passed, 0 failed, 0 skipped
139 steps passed, 0 failed, 0 skipped, 0 undefined
Took 0m16.305s
Error: Process completed with exit code 1.

rolandwalker:
First pytest is run, then behave is run.  Then the CI run exits with a failure if either one failed.  In this case, pytest is showing failures:
```
test/test_clistyle.py ss                                                 [  0%]
test/test_completion_engine.py ......................................... [ 15%]
...................x...................................................  [ 41%]
test/test_completion_refresher.py ....                                   [ 42%]
test/test_config.py ..............                                       [ 47%]
test/test_dbspecial.py ....                                              [ 49%]
test/test_main.py ...................                                    [ 56%]
test/test_naive_completion.py .....                                      [ 57%]
test/test_parseutils.py .............X........................           [ 71%]
test/test_prompt_utils.py .                                              [ 72%]
test/test_smart_completion_public_schema_only.py .................F....  [ 80%]
test/test_special_iocommands.py ...................                      [ 86%]
test/test_sqlexecute.py ...................................              [ 99%]
test/test_tabular_output.py F                                            [100%]
```

I believe the other CI runs in the matrix were canceled as soon as one failed.

schmeic:
I realized that the reason one of the tests is failing is because my sort algorithm is comparing lower cased matches, which doesn't work for the special commands starting with `\`, since these are case sensitive. How should I handle this?

It's also strange that when I run `./setup.py test` I get an error, but when I run the following command, it passes:
`pytest -vv -k test_empty_string_completion  test/test_smart_completion_public_schema_only.py`

schmeic:
Still waiting to hear back on how to handle this one failing unit test. Doesn't seem worth complicating the sort method just to support these special commands, and fix a single unit test.

schmeic:
Pretty disappointed not to get any feedback on this, I spent a lot of time trying to fix all these unit tests and I think it's really close.

I wanted to contribute to this project, but I guess I can just use my own branch.

amjith:
Sorry @schmeic. Life got ridiculous for the past few months. I'm taking a look now.

mycli/sqlcompleter.py
```diff
@@ -369,48 +369,61 @@ def find_matches(text, collection, start_only=False, fuzzy=True, casing=None):
...
+        return matches
 
     def get_completions(self, document, complete_event, smart_completion=None):
         word_before_cursor = document.get_word_before_cursor(WORD=True)
+        text = last_word(word_before_cursor, include='most_punctuations')
+
+        def sorted_completions(matches):
+            # sort by match point, then match length, then item text
+            matches = sorted(list(matches), key=lambda m:
+                             (m[1], m[0], m[2].lower().strip('`'), m[2].startswith('`')))
```
amjith, line 411: I think this `strip` is causing the tests to fail. The backticks are used to escape the table name if it happens to be a keyword. For instance if someone names the table `table` then it going to conflict with the keyword since `table` is a keyword. This adds backticks around it to escape the string to avoid the conflict.

amjith:
I noticed that the sort order seems to have messed up the ordering when it completes column names in a table. 

This is the completion from your branch. The columns are not listed at the top but instead pushed down below the keywords:
Cursor_and_0_0_python3_11_-__mycli_-u_root__Users_amjith_Dropbox_code_python_mycli_

This is the completion order from the main branch, the columns from the table are listed at the top. 

Cursor_and_0_0_python3_11_-__mycli_-u_root__Users_amjith_Dropbox_code_python_mycli_

amjith:
@schmeic I went down a rabbit hole here. The reason we don't sort the completion candidates is because we have the keywords listed in a curated order that is based on popularity instead of lexicography. 

But I went back to your original motivation for this PR. When you type `SELECT * FR` it should list `FROM` instead of `FROM_UNIXTIMESTAMP`. I modified the completion engine a little bit to allow only keywords (not functions) which fixes the issue superficially. 

I'll open a new PR and tag you on it so you could take it for a spin and let me know what you think.

amjith:
https://github.com/dbcli/mycli/pull/1175

mycli/sqlcompleter.py
schmeic: See the comment by @rolandwalker on June 13. The idea is that a table/field surrounded by backticks should match the same one without backticks. I've tried the change the tests to work with this.

schmeic:
> @schmeic I went down a rabbit hole here. The reason we don't sort the completion candidates is because we have the keywords listed in a curated order that is based on popularity instead of lexicography.
> 
> But I went back to your original motivation for this PR. When you type `SELECT * FR` it should list `FROM` instead of `FROM_UNIXTIMESTAMP`. I modified the completion engine a little bit to allow only keywords (not functions) which fixes the issue superficially.
> 
> I'll open a new PR and tag you on it so you could take it for a spin and let me know what you think.

The FROM coming after FROM_UNIXTIMESTAMP was just one example of what I thought was wrong with the completions. Why should columns be at the top? There are plenty of times when I'm looking for a keyword, and don't want to see all those unsorted column names at the top. I think the completions are a lot more intuitive when everything is sorted and works the same.

amjith:
> The FROM coming after FROM_UNIXTIMESTAMP was just one example of what I thought was wrong with the completions. 

I understand that. That's why I said my PR is a superficial fix. 

> Why should columns be at the top? 

Imagine this scenario: 
You just connected to a fresh DB and you don't know the tables or the columns in each table. You start typing `SELECT * FROM ` and you get the list of tables at the top so you navigate using arrow keys to get an idea of the tables. Then you pick the table from the suggestion and continue with your query `SELECT * FROM users WHERE `, you don't know the columns in the table but after the WHERE clause you get a list of column names from the `users` table. You can find all the columns using the arrow keys and pick the one that makes the most sense and so on. If we take the approach you've suggested the user has to know the columns in the table in order to start typing a few letters to get the desired completion. It thwarts the ability discover the schema.

I think that's a more human centric approach with UX. In fact, the keywords are ordered (I think I mentioned this in my previous comment as well) to make sure the most common keywords are listed at the top. This is also meant to facilitate a user to pick things from the completion menu rather than continue to type to get to their desired keyword.

My goal was to ensure we can provide a user friendly UX for people who pick from a menu without compromising the efficiency of someone who likes to type (or knows the schema full well). But you pointed out a very valid flaw with the FROM and FROM_UNIXTIMESTAMP that was annoying for a user who prefers to type rather than use arrows. So I decided to fix that tactically.

I hope my counterpoint convinces you of the reason for sticking with the current approach. As always I'm open to revisiting the design if there are compelling reasons to do so.

schmeic:
Ok, I guess I see your point about "discovering the schema", but I would argue that if you want to do that, use:
`DESC <table name>`

Once you start typing any characters, the current approach will typically provide worse results than my approach, especially when you want a keyword instead of a column. So, I still feel like my approach is better, since it will provide better results the majority of the time. It seems pretty inefficient to try and discover the schema using autocomplete.

As far as your point about keyword ordering, they may be listed with more common ones at the top, but what if I'm looking for a less common one?  My approach will find it much more efficiently. However, I think the keyword search should be doing fuzzy searching when using my approach. Currently, when I search for `WHERE`, I need to type the first 4 chars to get past `WHEN`. If we enable fuzzy searching, I would be able to just type `wr`.

amjith:
> Ok, I guess I see your point about "discovering the schema", but I would argue that if you want to do that, use:
DESC <table name>

The whole point of mycli is to not break your flow. When I am typing a query I don't want to stop in the middle and run a `DESC table` to continue my query. This becomes especially relevant when writing JOIN queries. 

Adding fuzzy matching for keywords is not a bad idea. Let me see if I can prototype that in a PR.

amjith:
I've updated #1175 to enable fuzzy matching for keywords. Can you give it a try? 

You can install it as follows:

```
pip install -U https://github.com/dbcli/mycli/archive/refs/heads/fix-select-star.zip
```

schmeic:
Sounds like you prefer good discovery over relevant results when text is typed. I want to find the completion I'm looking for with the fewest keystrokes. I'm not sure if it's possible to have both, let me know if you have any ideas on how to make it work.

I actually added fuzzy keyword matching on my branch, it works better.

amjith:
> Sounds like you prefer good discovery over relevant results when text is typed.


I'm not doing a good job of explaining myself. I value both I think we can achieve the outcome you're looking for without compromising discoverability. Let me explain it slightly differently. I have a users table with columns that have similar names `user_name` and `user_name_normalized`. 

When I type `SELECT * FROM users WHERE u` I want to see `user_name` on the top and `user_name_normalized` at the bottom because the first one is shorter and if the user wants to select the second one they can type the suffix that is unique to the longer name (like `zed`). 

But what I don't want to see is `SELECT * FROM users WHERE u` listing the functions `ucase` and `unix_timestamp` before listing the column names. Because that interrupts my flow to abandon my query and go run a `DESC users` to find the column names. 

I think enabling `fuzzy` on keywords has fixed the root of your complaint (listing `FROM_UNIXTIME` before `FROM`). For that I'm grateful to you for proposing a solution and forcing me to think hard about the issue. But unfortunately I don't agree that we should sort all the entries without prioritization. That hampers discoverability. 

I am sorry this PR is not getting merged but I want to thank you for prompting me to do better with the UX.

schmeic:
Just to be clear, the root of my complaint was not listing FROM_UNIXTIME before FROM, that was just one example of what I considered to be undesirable completion results.

I want to move the completion term that I'm looking for to the top with the minimum amount of keystrokes, and prioritizing column names over other completions hinders that.

I wish we could have had this discussion before you told me to fix all the unit tests and you would accept the pull request. I wasted a lot of time working on them.

Issue #1146, opened 2024-01-17
Database connection failed: error('unpack requires a buffer of 4 bytes').


bkysela:
When connecting using `mycli` I see the following on the console.

```
$ mycli
Connecting to socket /var/run/mysqld/mysqlx.sock, owned by user mysql
unpack requires a buffer of 4 bytes
$ 
```

Running with DEBUG, this is in the log.

```
2024-01-17 12:16:16,402 (2797336/MainThread) mycli.sqlexecute DEBUG - Connection DB Params: 
        database: 'xxxxxxx'     user: 'xxxxxxxx'    host: xxxxx      port: xxxxx      socket: '/var/run/mysqld/mysqlx.sock'   charset: 'utf8' local_infile: False     ssl: None             ssh_user: None  ssh_host: None  ssh_port: 22    ssh_password: None      ssh_key_filename: None  init_command: None
2024-01-17 12:16:16,402 (2797336/MainThread) mycli.main DEBUG - Database connection failed: error('unpack requires a buffer of 4 bytes').
2024-01-17 12:16:16,402 (2797336/MainThread) mycli.main ERROR - traceback: 'Traceback (most recent call last):\n  File "/usr/lib/python3/dist-packages/mycli/main.py", line 482, in connect\n    _connect()\n  File "/usr/lib/python3/dist-packages/mycli/main.py", line 456, in _connect\n    self.sqlexecute = SQLExecute(\n  File "/usr/lib/python3/dist-packages/mycli/sqlexecute.py", line 119, in __init__\n    self.connect()\n  File "/usr/lib/python3/dist-packages/mycli/sqlexecute.py", line 177, in connect\n    conn = pymysql.connect(\n  File "/usr/lib/python3/dist-packages/pymysql/connections.py", line 353, in __init__\n    self.connect()\n  File "/usr/lib/python3/dist-packages/pymysql/connections.py", line 632, in connect\n    self._get_server_information()\n  File "/usr/lib/python3/dist-packages/pymysql/connections.py", line 1065, in _get_server_information\n    self.server_thread_id = struct.unpack("<I", data[i : i + 4])\nstruct.error: unpack requires a buffer of 4 bytes\n'
```

```
$ mycli --version
Version: 1.27.0
$ mysqld --version
/usr/sbin/mysqld  Ver 8.0.35-0ubuntu0.22.04.1 for Linux on x86_64 ((Ubuntu))
```

The regular MySQL client connects without issue.

[...]

Issue #726, opened 2019-03-30
Can support fzf select history in mycli?


kevinhwang91:
It's very  convenient like in shell to select history.
I think is a very awesome feature in mycli!

tsroten:
This is a neat idea! At this time, prompt_toolkit, the library we use for searching the history, doesn't support customizing the search. But, I think this is definitely do-able.

@jonathanslenders (author of prompt_toolkit), has said this is feasible for the future :thumbsup:.

JunkoPF:
Is there any progress?

PR #1170, opened 2024-08-03
Add fzf history search feature

lazmond3:
### Background
Previously, the history function of mycli only allowed viewing history one line at a time, making it feel like peering through a keyhole. Users had to navigate through the history by typing on the keyboard to see the previous and next entries, making it difficult to find the desired history entry. A request for a feature to enable fzf-like history search was made at https://github.com/dbcli/mycli/issues/726.

### Achievements
Implemented the ability to perform fzf search on mycli history. fzf is a software written in Go, and its binary file is also named fzf. (See: [junegunn/fzf: :cherry_blossom: A command-line fuzzy finder](https://github.com/junegunn/fzf))
The mycli history fzf search feature is designed to switch between the old implementation and the new one based on the presence of the fzf binary. (It might be better to have this feature turned off by default and enabled via an option. Feedback is welcome.)

The history format is `<timestamp>  <stmt>`. (Placing the timestamp at the end makes formatting difficult and searching by date harder.) Currently, the timestamp is handled as a string and displayed with seconds. Using `FormattedText` from `prompt_toolkit` to gray out the timestamp could improve visibility, but to keep the scope of this PR focused, this enhancement is not included.

Since fzf provides line-based selection functionality, newlines in the history are replaced with spaces for the suggestion list. However, when setting data to the buffer, the original text is retrieved and set.

Regarding the fzf search score, the `tiebreak=index` option is used. This is because I believe that when searching for SQL queries, users often prefer to use the most recent ones. If the default `length` option is used, searching for a common term like `select` would bring up entries where `select` was mistakenly entered by itself, which can be annoying. For more details, see the [fzf tiebreak option documentation](https://man.archlinux.org/man/fzf.1.en#tiebreak=).

This implementation is an ad-hoc override of some `prompt_toolkit` behavior. Therefore, the implementation is gathered under a `toolkit` subpackage within the `packages` directory.

### Alternative Approach
Ideally, we should contribute to `prompt_toolkit` and use the history functionality implemented there in mycli. However, `prompt_toolkit` is pure Python, and fzf cannot be integrated directly. Rewriting the Go-based program in Python and adapting it to the `prompt_toolkit` Completion mechanism would be a significant undertaking.

For this reason, I decided to override some history-related implementations and keybindings in mycli.

### Comments
mycli is a great tool, and I have greatly benefited from it. I am pleased to have the opportunity to contribute to this excellent software and hope that mycli will become an even more convenient tool.

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -11,6 +11,8 @@ Internal:
 Features:
 ---------
 
+* Added fzf history search functionality. The feature can switch between the old implementation and the new one based on the presence of the fzf binary.
+
 
 1.27.2 (2024/04/03)
 ===================
@@ -23,7 +25,6 @@ Bug Fixes:
 
 1.27.1 (2024/03/28)
 ===================
-* Added fzf history search functionality. The feature can switch between the old implementation and the new one based on the presence of the fzf binary.
 
 Bug Fixes:
 ----------
--- a/changelog.md
+++ b/changelog.md
@@ -23,7 +23,7 @@ Bug Fixes:
 
 1.27.1 (2024/03/28)
 ===================
-* Added fzf-like history search functionality. The feature can switch between the old implementation and the new one based on the presence of the fzf binary.
+* Added fzf history search functionality. The feature can switch between the old implementation and the new one based on the presence of the fzf binary.
 
 Bug Fixes:
 ----------
--- a/changelog.md
+++ b/changelog.md
@@ -23,7 +23,7 @@ Bug Fixes:
 
 1.27.1 (2024/03/28)
 ===================
-
+* Added fzf-like history search functionality. The feature can switch between the old implementation and the new one based on the presence of the fzf binary.
 
 Bug Fixes:
 ----------
--- a/mycli/packages/filepaths.py
+++ b/mycli/packages/filepaths.py
@@ -100,7 +100,7 @@ def guess_socket_location():
         for r, dirs, files in os.walk(directory, topdown=True):
             for filename in files:
                 name, ext = os.path.splitext(filename)
-                if name.startswith("mysql") and ext in ('.socket', '.sock'):
+                if name.startswith("mysql") and name != "mysqlx" and ext in ('.socket', '.sock'):
                     return os.path.join(r, filename)
             dirs[:] = [d for d in dirs if d.startswith("mysql")]
     return None
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -14,4 +14,4 @@ pyperclip>=1.8.1
 importlib_resources>=5.0.0
 pyaes>=1.6.1
 sqlglot>=5.1.3
-setuptools
+setuptools<=71.1.0
--- a/mycli/AUTHORS
+++ b/mycli/AUTHORS
@@ -97,6 +97,7 @@ Contributors:
   * Zhanze Wang
   * Houston Wong
   * Mohamed Rezk
+  * Ryosuke Kazami
 
 
 Created by:
--- a/mycli/key_bindings.py
+++ b/mycli/key_bindings.py
@@ -3,6 +3,8 @@ from prompt_toolkit.enums import EditingMode
 from prompt_toolkit.filters import completion_is_selected, emacs_mode
 from prompt_toolkit.key_binding import KeyBindings
 
+from .packages.toolkit.fzf import search_history
+
 _logger = logging.getLogger(__name__)
 
 
@@ -100,6 +102,12 @@ def mycli_bindings(mycli):
                   and b.text[cursorpos_abs] in (' ', '\n'):
                 cursorpos_abs -= 1
             b.cursor_position = min(cursorpos_abs, len(b.text))
+
+    @kb.add('c-r', filter=emacs_mode)
+    def _(event):
+        """Search history using fzf or default reverse incremental search."""
+        _logger.debug('Detected <C-r> key.')
+        search_history(event)
 
     @kb.add('enter', filter=completion_is_selected)
     def _(event):
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -36,7 +36,6 @@ from prompt_toolkit.formatted_text import ANSI
 from prompt_toolkit.layout.processors import (HighlightMatchingBracketProcessor,
                                               ConditionalProcessor)
 from prompt_toolkit.lexers import PygmentsLexer
-from prompt_toolkit.history import FileHistory
 from prompt_toolkit.auto_suggest import AutoSuggestFromHistory
 
 from .packages.special.main import NO_QUERY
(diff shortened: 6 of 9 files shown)


amjith:
Sorry about the delay in response. I'll give it shot later today and provide feedback. Thank you for taking the time to contribute.

changelog.md
```diff
@@ -23,7 +23,7 @@ Bug Fixes:
 
 1.27.1 (2024/03/28)
 ===================
-
+* Added fzf-like history search functionality. The feature can switch between the old implementation and the new one based on the presence of the fzf binary.
```
lazmond3, line 26: Not fzf-like, but fzf itself.

lazmond3 (commented):
It seems that the pytest for python 3.12 failed. I'll check it later.

lazmond3:
@amjith 

It appears that the failure in the CI tests is unrelated to the changes introduced in this PR. The error message indicates that the use of setuptools for running pytest is deprecated and no longer supported:

```
RuntimeError: Support for the test command was removed in Setuptools 72
```

To resolve this issue, I propose modifying the command in the CI step as follows:

```
- name: Pytest / behave
  env:
    PYTEST_PASSWORD: root
    PYTEST_HOST: 127.0.0.1
  run: |
    pytest test --cov-report --cov=mycli 
```

Would it be acceptable for me to make this change to ensure the tests run successfully?

rolandwalker:
What happens if you change the line `setuptools` in `requirements-dev.txt` to `setuptools<=71.1.0` ?

amjith:
This is nice. I have not used fzf before (fishshell user so never really had a need for fzf). I haven't had a chance to review the implementation itself. But wanted to chime in and say nicely done. I like the fact that it uses fzf if it is installed otherwise falls back to the original behavior. I think that is the right approach. I don't want to hide this behind a config option.

amjith:
Feel free to modify the CI steps to make the tests pass. I'm planning on modernizing the codebase to switch to pyproject.toml, ruff and uv. So don't worry too much about making it perfect. Just make whatever edits necessary to get the CI to pass.

lazmond3:
@amjith @rolandwalker

Thank you for your guidance. Following @rolandwalker 's suggestion to specify `setuptools <= 71.1.0` in the CI configuration has proven successful. Considering that the required changes are minimal, I intend to proceed with this approach to ensure that the CI passes. I appreciate your assistance and look forward to your feedback.

lazmond3:
@amjith 
Thank you for your patience. We have identified a solution to address the CI failure.

## Cause of CI Failure

**CI Link:** [GitHub Actions](https://github.com/dbcli/mycli/actions/runs/10784615615/job/30492417249)

In this CI, the following two steps are experiencing failures:

- `test/features/connection.feature:4`: Execute mycli on localhost without specifying a port
- `test/features/connection.feature:20`: Execute mycli without specifying a host and port

These tests fail when mycli attempts to connect to the MySQL server using a socket file. The error message encountered is as follows:

```
error('unpack requires a buffer of 4 bytes')
```

## Problem Investigation

The following issue provided insights that were instrumental in resolving the problem:

- [Database connection failed: error('unpack requires a buffer of 4 bytes'). · Issue #1146 · dbcli/mycli](https://github.com/dbcli/mycli/issues/1146)

It has been reported that on Ubuntu 22.04, connections via socket files fail, resulting in the error message "unpack requires a buffer of 4 bytes."

Upon examining the behavior of mycli in an Ubuntu 22.04 environment using Docker on a local machine, it was confirmed that the error occurs only when the socket file is ***not explicitly specified.*** This causes mycli to connect to the MySQL server via the socket file. The logic used when the socket file is "not explicitly specified" is [guess_socket_location](https://github.com/dbcli/mycli/blob/df44b068ae10dfa37eac4682c62ddff932966cc3/mycli/packages/filepaths.py#L96), which often returns "mysqlx.sock" before "mysqld.sock" depending on conditions.

Additionally, experiments using mycli within the Docker environment revealed that attempting to specify the socket file `/var/run/mysqld/mysqlx.sock`, which is intended for a different protocol X server, results in the error "unpack requires a buffer of 4 bytes." In other words, it seems that the root cause of error #1146 might have been an attempt to connect to the MySQL server using the socket file intended for the X plugin, possibly due to incorrect socket file guess results.


## Fix Details

The issue can be resolved by filtering out `mysqlx.sock` in `guess_socket_location`. I pushed this fix to the following PR in the repository I forked, and confirmed that the CI passed.

- fix guess_socket_location, exclude mysqlx.sock by lazmond3 · Pull Request #3 · lazmond3/mycli

Additionally, I want to clarify that the CI failure is not caused by my code changes, as shown in [the test CI fails by lazmond3 · Pull Request #5 · lazmond3/mycli](https://github.com/lazmond3/mycli/pull/5). In this PR, I only fixed the version of `setuptools` against the main branch, yet it failed with the same error as mentioned above.

--- 

In PR #1170, I intend to include the fix for `guess_socket_location` to address the CI issue and "Close #1146" simultaneously. Is that acceptable?

changelog.md
```diff
@@ -23,7 +23,7 @@ Bug Fixes:
 
 1.27.1 (2024/03/28)
 ===================
-
+* Added fzf history search functionality. The feature can switch between the old implementation and the new one based on the presence of the fzf binary.
```
lazmond3, line 26: I'm sorry, the release notes for the implemented feature should be written in the upcoming release section, not in the already released 1.27.1. I will correct this along with the CI.

lazmond3: Fixed in [e766643](https://github.com/dbcli/mycli/pull/1170/commits/e766643982a6062043767904afdf6e893050a0cb)

lazmond3:
@amjith @rolandwalker 

I implemented the solution I mentioned the other day, fixed the failing CI, and simultaneously resolved issue #1146.

mycli/packages/filepaths.py
```diff
@@ -100,7 +100,7 @@ def guess_socket_location():
         for r, dirs, files in os.walk(directory, topdown=True):
             for filename in files:
                 name, ext = os.path.splitext(filename)
-                if name.startswith("mysql") and ext in ('.socket', '.sock'):
+                if name.startswith("mysql") and name != "mysqlx" and ext in ('.socket', '.sock'):
```
amjith, line 103: Can you elaborate a little bit on what this `mysqlx` is meant to do?

requirements-dev.txt
```diff
@@ -14,4 +14,4 @@ pyperclip>=1.8.1
 importlib_resources>=5.0.0
 pyaes>=1.6.1
 sqlglot>=5.1.3
-setuptools
+setuptools<=71.1.0
```
amjith, line 17: Why is this version restricted?

amjith:
Thank you for the PR. I appreciate your patience. Sorry it took a while to merge.

PR #1177, opened 2024-11-16
Fix the test for \pipe_once command.

amjith:
## Description
Fix a test for the `\pipe_once` special command.


amjith:
Thanks to @danp for finding this oversight while adding the feature to [litecli](https://github.com/dbcli/litecli/pull/193)

rolandwalker (approved):

Issue #1176, opened 2024-11-08
'Channel' object has no attribute 'getsockopt' when connecting with SSL through a jump server


amne:
I'm trying to connect to an Amazon RDS MySQL 8 instance through a jump server with IAM ROLE.
```
$ docker run -it -v ~:/home/amne ccruceru/mycli \
 --ssh-user ccruceru\
 --ssh-host jump-server-example.net\
 --ssh-key-filename ~/.ssh/id_rsa\
 mysql://iam_usr@engine8-db-databasecluster.cluster-blabla.us-east-1.rds.amazonaws.com\
 -p $TOKEN\
 --ssl-verify-server-cert
```

But this command fails with a simple error message (no stack trace is printed):
```
'Channel' object has no attribute 'getsockopt'
```

My understanding is that with "--ssl-verify-server-cert" it should try to pass the token in cleartext which is required for IAM auth (source https://github.com/dbcli/mycli/issues/550#issuecomment-642516893 .)

Connecting without SSL but with a traditional user that exists in mysql.user and native password works fine.

Here is the docker file:

```
FROM python:3.9.15-bullseye

RUN pip3 install paramiko PyMySQL==1.0.2 mycli==1.27.2 \
    && apt-get update \
    && apt-get install -y less locales \
    && locale-gen en_US.UTF-8 \
    && apt-get clean autoclean \
    && apt-get autoremove --yes \
    && rm -rf /var/lib/{apt,dpkg,cache,log}/

ADD docker-entrypoint.sh /usr/local/bin/

ENTRYPOINT ["docker-entrypoint.sh"]

CMD ["mycli"]
```

```
docker build -t ccruceru/mycli .
```

amne:
I followed CONTRIBUTING.md and got a venv where I can try to debug this and I managed to trace the error.

It appears that the `paramiko.Channel` object instance is reaching python lib `ssl.py` code which expects something else that implements `getsockopt()`?

```
got sock <paramiko.Channel 0 (open) window=2097152 in-buffer=78 -> <paramiko.Transport at 0x35916e30 (cipher aes128-ctr, 128 bits) (active; 1 open channel(s))>>
Traceback (most recent call last):
  File "/home/amne/work/lab/mycli/mycli_dev/lib/python3.10/site-packages/pymysql/connections.py", line 670, in connect
    self._request_authentication()
[...]

PR #1178, opened 2024-11-20
fix SSL through SSH jump

amne:
## Description
Attempt to fix https://github.com/dbcli/mycli/issues/1176
The error is caused by paramiko.channel.Channel class that implements a subset of python's socket API and cannot be properly wrapped in SSL because of this.

These changes replace the use of Channel with a straightforward approach of opening a python socket to local bound host and port of a tunnel to the jump host.

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,7 @@ Upcoming Release (TBD)
 Bug Fixes:
 ----------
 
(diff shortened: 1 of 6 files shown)


amne:
owned by git! again

PR #1180, opened 2024-11-25
Modernize

amjith:
## Description

Switch from setup.py to pyproject.toml. 
Use tox.ini to run tests.
Use GH actions to release new versions.
Use setuptools-scm to get version from git tags.

## Checklist


amjith:
Only because I didn’t realize 3.8 had gone EOL.
😊

PR #1198, opened 2025-03-28
Add collapsed output special command

FatBoyXPC:
## Description
It's sometimes annoying to have auto_vertical_output enabled in myclirc and end up wanting to have the output not be expanded. The current workaround for this is to modify the setting in the rc file and restart mycli (or start a new session), which is obviously not super desirable. Now you can add `\g` at the end of the query akin to `\G` if you're not using auto vertical output.

I would add this to changelog but I don't see anywhere for unreleased features.

## Checklist


amjith:
Thank you!

Issue #1195, opened 2025-03-19
Is it possible to specify `--init-command` in `~/.myclirc`?


your-diary:
Ideally I want to specify different `--init-command` for different DSNs defined in `[alias_dsn]`.

For example, I want to always make the session read only via `SET SESSION TRANSACTION READ ONLY` when I connect to production database.

amjith:
That's a good idea. We don't yet have that feature. Let me see if I can prototype something.

serijn441:
This fix in another issue and it might help you [click](https://newrides.es/captha/)

amjith:
Try out this PR #1200 

You can install directly from the branch using: 

```
pip install https://github.com/dbcli/mycli/archive/refs/heads/amjith/init-per-dsn.zip
```

your-diary:
@amjith 
Thank you for your great work!

It worked.

Only one problem I have is, it seems `[alias_dsn.init-commands]` is behaving as if it was a required parameter.
When I omit the parameter, I get this error:

```
/ # mycli -d main_3
(1065, 'Query was empty')
```

Perhaps this means an empty string is passed to MySQL as an init command rather than not passing an init command?

amjith:
Thank you for trying it out and catching that bug. I have now merged the PR. I'll release a new version later today. 

I see you've filed an issue with pgcli to port this feature. I'll get that going soon. I will need your help with testing that as well. 😄

your-diary:
@amjith 
Thank you!
Tried the fix and confirmed it is now working as expected.

---

## Testing Env

### `Dockerfile`

```dockerfile
FROM alpine:3.21

RUN apk update
RUN apk add python3 py3-pip
RUN apk add postgresql-dev
RUN apk add bash vim

RUN adduser -D user
USER user

WORKDIR /home/user
```

### `compose.yaml`

```yaml
services:
    test_service_alpine:
        build: .
        image: test_image_alpine
        container_name: test_container_alpine
        command:
            - sh
            - -c
            - |
                export SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 &&
                python3 -m pip install --break-system-packages 'https://github.com/dbcli/mycli/archive/refs/heads/amjith/init-per-dsn.zip' &&
                python3 -m pip install --break-system-packages 'https://github.com/dbcli/pgcli/archive/refs/heads/amjith/init-command.zip' &&
                export PATH="${PATH}:/home/user/.local/bin" &&

                #creates the config files
[...]

PR #1200, opened 2025-04-19
Amjith/init per dsn

amjith:
## Description

#1195 

Add dsn specific init-commands and a global init-command to myclirc. 

## Checklist

### DIFF
--- a/mycli/sqlexecute.py
+++ b/mycli/sqlexecute.py
@@ -233,7 +233,7 @@ class SQLExecute(object):
             ssl=ssl_context,
             program_name="mycli",
             defer_connect=defer_connect,
-            init_command=init_command,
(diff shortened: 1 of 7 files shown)


mycli/sqlexecute.py
```diff
@@ -236,6 +236,9 @@ def connect(
             init_command=init_command,
         )
 
+        if init_command:
+            print("Running init commands:\n", init_command)
```
rolandwalker, line 240: Should such output be to sys.stderr?

rolandwalker (approved):

mycli/sqlexecute.py
amjith: Yup. Changed it.

Issue #1202, opened 2025-04-21
Improve Handling of password-file Option


allrob23:
The current `get_password_from_file` function uses `os.access()` to check file permissions before opening, which creates a security risk due to race conditions. The Python documentation recommends using EAFP (Easier to Ask for Forgiveness than Permission) [link](https://docs.python.org/3.10/library/os.html#os.access) to avoid this.

### Proposed Changes

- Replace os.access() and other pre-checks with EAFP by directly attempting to open the file and catching exceptions.
- Improve error handling by raising specific exceptions (e.g., FileNotFoundError, PermissionError) with clear messages, possibly using a custom exception class for the library.

This will make the function more secure and user-friendly.

rolandwalker:
Resolved in [#1203](https://github.com/dbcli/mycli/pull/1203)

PR #1203, opened 2025-04-21
refactor: adopt EAFP and explicit error handling for password file

allrob23:
## Description

#1202

This PR enhances `get_password_from_file` by removing pre-check `if` statements and adding explicit error handling for cases like file not found or permission denied, e.g., `mycli.main.PasswordFileError: Password file '/tmp/a.txt' is empty or contains only whitespace`, improving security and clarity.

## Checklist


mycli/main.py
```diff
@@ -536,14 +539,21 @@ def _connect():
...
         if password_file:
-            if (os.path.isfile(password_file) or stat.S_ISFIFO(os.stat(password_file).st_mode)) and os.access(password_file, os.R_OK):
+            try:
                 with open(password_file) as fp:
-                    password_from_file = fp.readline()
-                    password_from_file = password_from_file.rstrip().lstrip()
-
-        return password_from_file
+                    password = fp.readline().strip()
+                    if not password:
```
rolandwalker, line 546: Doesn't this introduce a new constraint?

mycli/main.py
```diff
@@ -536,14 +539,21 @@ def _connect():
...
+                        raise PasswordFileError(f"Password file '{password_file}' is empty or contains only whitespace")
+                    return password
+            except FileNotFoundError:
+                raise PasswordFileError(f"Password file '{password_file}' not found") from None
+            except PermissionError:
+                raise PasswordFileError(f"Permission denied reading password file '{password_file}'") from None
+            except IsADirectoryError:
+                raise PasswordFileError(f"Path '{password_file}' is a directory, not a file") from None
+            except OSError as e:
```
rolandwalker, line 555: Also advisable to catch `RuntimeError` ? I'm unsure.

mycli/main.py
allrob23: Are you referring to `if not`? If so, I thought it makes sense to raise an error when the content is empty

mycli/main.py
allrob23: I don't think it's useful, maybe I can leave this last exception generic to handle all other types of exceptions? what do you think?

mycli/main.py
rolandwalker: Yes, raising an error when the password is empty changes the functionality, which I suppose we should hold constant for this PR.  Looking closer, I would even `strip()` the newline more precisely but again that is out of scope.

Because the empty string is a valid password, as are passwords containing whitespace.  Whether they are advisable is a separate Q.

mycli/main.py
rolandwalker: Your decision!

mycli/main.py
allrob23: Makes sense, I updated the code.

rolandwalker:
@allrob23 I see this PR now has a merge conflict.  Do you want me to fix that and merge, or take that on yourself?

PR #1207, opened 2025-04-22
Add `ruff format` linting

rolandwalker:
## Description
The tox lint check at the end of ci.yml doesn't seem to be showing lint errors.

Add a separate lint action, a ruff configuration, and ruff-format the codebase.  The changes are all whitespace, if joining strings can be considered whitespace.

Passing "ruff check" would be more difficult.

## Checklist
- [x] ~I've added this contribution to the `changelog.md`.~ there is already an entry about linting, which is sufficient.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,3 @@ jobs:
           TERM: xterm
         run: |
           uv run tox -e py${{ matrix.python-version }}
-
-      - name: Run Style Checks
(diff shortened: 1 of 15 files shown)


amjith:
This is great. We should also look into adding a pre-commit integration to make sure we stay clean going forward. Litecli has a pre-commit config https://github.com/dbcli/litecli/blob/main/.pre-commit-config.yaml

PR #1209, opened 2025-04-24
Incrementally improve linting setup

rolandwalker:
## Description

 * let lint target the oldest supported Python version
 * disable lint rules which may conflict with the formatter, per Astral recommendations, and document
 * establish rules for sorting imports (but don't activate this yet in CI)
 * enable preview mode for "ruff format", and reformat code, making some whitespace tighter

Enabling `preview = true` for ruff might be controversial, but I've had only good luck with it.

The ignored lint rules wouldn't be needed by default, but the "E" and "W" series are explicitly activated.

## Checklist
- [x] ~I've added this contribution to the `changelog.md`.~ already covered

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,6 +56,7 @@ mycli = ["myclirc", "AUTHORS", "SPONSORS"]
 include = ["mycli*"]
 
 [tool.ruff]
+target-version = 'py39'
 line-length = 140
 
 [tool.ruff.lint]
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -72,12 +73,24 @@ select = [
 ignore = [
     'E401',   # Multiple imports on one line
     'E402',   # Module level import not at top of file
-    'E501',   # Line too long
     'F541',   # f-string without placeholders
     'PIE808', # range() starting with 0
+    # https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules
+    'E501',   # Line too long
```
amjith, line 79: I don't think E501 is needed.

pyproject.toml
```diff
@@ -72,12 +73,24 @@ select = [
 ignore = [
     'E401',   # Multiple imports on one line
     'E402',   # Module level import not at top of file
-    'E501',   # Line too long
     'F541',   # f-string without placeholders
```
amjith, line 76: We can remove F541 as well and fix all the places that violate this rule.

amjith (approved):
I'm a fan of not bothering with exceptions and going with the defaults and let `ruff check --fix` do the work. I realize we need a minimal set of exceptions for conflicting rules. Other than that we can go all in.

pyproject.toml
rolandwalker: I'll remove it in this PR, but the point is that sometimes `ruff check` and `ruff format` have different opinions about the line length.  We can always add it back if there is friction.

pyproject.toml
rolandwalker: Removed.  F541 was an intentional style choice.  I think there is nothing wrong with f-strings lacking placeholder.

PR #1210, opened 2025-04-25
Add a preview window to fuzzy history search

rolandwalker:
## Description
Add a preview window to fuzzy history search, compressing whitespace, and wrapping text.

This is highly desirable since many queries are longer than one line, and can't be distinguished in the search.  That can be considered a regression over non-fuzzy search.

It would be even nicer to
 * somehow avoid compressing quoted whitespace
 * detect when the user has already set a preview location preference in $FZF_DEFAULT_OPTS

## Checklist


amjith (approved):

amjith:
Looks really nice.

PR #1218, opened 2025-04-26
Fix repeated `endswith()`

rolandwalker:
## Description

Missed in #1213, which I merged without review!

## Checklist


rolandwalker:
Also merging this without review, since it's an outright bug/blocker.

Issue #1066, opened 2022-07-20
sql-insert formatter sometimes calls tables `DUAL` if the table name is not enclosed in back-ticks


cricalix:
Context
==
Queries written without back-ticks on the table names will sometimes result in a table name of `DUAL` in the generated sql-insert formatted text.

```
mysql 10.5.15-MariaDB-0+deb11u1
mycli 1.23.2

cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
```

To rule out debian packaging, I ran `pip install -U mycli`, which installed `1.25.0` and executed `~/.local/bin/mycli`. The behaviour persisted.

Reproduction
==

The behaviour appears to be tied to table names, and the use (or lack thereof) of back-ticks when writing the query. I presume this is reserved keyword business of some sort, though the query executes just fine so I don't think it's a DB engine problem.

```
mysql root@(none):bar> \T sql-insert
Changed table format to sql-insert
Time: 0.000s

mysql root@(none):bar> create table alias (a int);
Query OK, 0 rows affected
Time: 0.027s
```
Query without back-ticks on table name
--
```
mysql root@(none):bar> select * from alias;
INSERT INTO `DUAL` (`a`) VALUES
;
0 rows in set
```
Query with back-ticks on table name
--
```
mysql root@(none):bar> select * from `alias`;
INSERT INTO `alias` (`a`) VALUES
;
0 rows in set
Time: 0.002s
```

So far I can get it to do this for the names
* admin
* alias
* domain
* quota

All these names are table names created by [Postfixadmin.](https://github.com/postfixadmin/postfixadmin).

I note that mycli syntax-highlights `admin` in the same colour as SELECT/FROM/LIMIT, but the other three do not get the same colour applied.

PR #1227, opened 2025-04-30
Working extract-tables for multi-statement inputs

### LINKED ISSUE #1122: why table name will be `DUAL` using sql-insert tableformat
image
image

I found that <tableformat> must be used as a single command to be effective, otherwise the table name will be `DUAL`, is it a bug?

rolandwalker:
## Description
Working extract-tables for multi-statement inputs, using an unholy combination of sqlparse and sqlglot.

sqlparse is more lenient and can chop up multiple statements even though it doesn't understand `\T`.  sqlglot is stricter, and chokes when it sees `\T`, but is better at the simple task of extracting table names from a valid statement, and provides a much better interface.

We could also use `sqlglot.tokenize()` and split on the semicolon token, which would be more verbose.

Closes #1122, in which the table name was always given as `DUAL` when multiple statements were in the same input leading with `\T sql-insert`.

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,8 +1,15 @@
 Upcoming Release (TBD)
 ======================
 
+Bug Fixes
+----------
+
+* Let table-name extraction work on multi-statement inputs.
+
+
 Internal
 --------
+
 * Work on passing `ruff check` linting.
 * Remove backward-compatibility hacks.
 * Pin more GitHub Actions and add Dependabot support.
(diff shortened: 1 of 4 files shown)


rolandwalker:
Appears to close https://github.com/dbcli/mycli/issues/1066 as well.

amjith (approved):
Wow. That is an interesting approach.

rolandwalker:
I think a tree-sitter grammar could be even better at covering all the cases with one tool.  In my experience tree-sitter is especially good at best-effort parsing of erroneous or partial data, which makes sense since it was developed for syntax highlighting in an editor.

PR #1229, opened 2025-05-03
Implement \llm command.

amjith:
## Description
WIP. Looking for initial feedback. 

## Checklist

### DIFF
--- a/mycli/packages/special/llm.py
+++ b/mycli/packages/special/llm.py
@@ -3,11 +3,11 @@ import io
 import logging
 import os
 import re
+from runpy import run_module
 import shlex
 import sys
-from runpy import run_module
-from typing import Optional, Tuple
 from time import time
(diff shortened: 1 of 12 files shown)


mycli/packages/special/llm.py
```diff
@@ -0,0 +1,284 @@
...
+
+```sql
+SELECT count(*) FROM table_name;
+```"""
+
+
+def initialize_llm():
+    if click.confirm("This feature requires additional libraries. Install LLM library?", default=True):
+        click.echo("Installing LLM library. Please wait...")
+        run_external_cmd("pip", "install", "--quiet", "llm", restart_cli=True)
```
rolandwalker, line 157: Running `pip install` from within the application is surprising and may have unintended effects.  Can it instead recommend reinstalling with extras?

rolandwalker (approved):

PR #1240, opened 2025-06-13
Remove click<8.1.8 version limit

wrmsr:
## Description

`click` was set to `<8.1.8` in bb18b0c2f2ed7375efe31d379e616a11c82b1299 in #1200 due to [this test run](https://github.com/dbcli/mycli/actions/runs/14551337225/job/40822870505) failing.

Looking at the click diff, `_pipepager` [was changed](https://github.com/pallets/click/commit/299efb82e1a3d34d870129dc0c677c6efb42d811#diff-f25fea0c0c214556563712a06f29270750a9b484a56d6a4df07ad5ae181b3eabR408-L417) to `shutil.which` the cmd it's given, where it previously just dumped it into `subprocess.Popen` with `shell=True`. Test code [tries to set](https://github.com/dbcli/mycli/blob/4aa159861d71f1dfeeef1aeb17300f83b9ad0fbf/test/features/environment.py#L70) the pager to a multi-arg shell invocation, not a `which`-able path to a single file.

I've changed it to write out an executable temp file to pass as the pager, and that seems to work on my box.

I'm not familiar with the project's supported environments and how well this fits into them, but this seems to fix the problem on my box.

## Checklist

### DIFF
--- a/test/features/environment.py
+++ b/test/features/environment.py
@@ -85,12 +85,13 @@ def before_all(context):
         "pager_boundary": "---boundary---",
     }
 
-    _, wrappager = mkstemp()
+    wrappager_fd, wrappager = mkstemp()
     with open(wrappager, "w") as f:
         f.write(WRAPPAGER_TEMPLATE.format(
             sys_executable=sys.executable,
             pager_boundary=repr(context.conf["pager_boundary"]),
         ))
+    os.close(wrappager_fd)
     os.chmod(wrappager, 0o755)
 
     _, my_cnf = mkstemp()


rolandwalker:
It would be great to have some bisecting done to narrow down the problem.  I took a look at the most likely parts of the `click` source and didn't see recent suspect changes.

wrmsr:
Reproducing it locally, narrowed it down to [299efb82e1a3d34d870129dc0c677c6efb42d811](https://github.com/pallets/click/commit/299efb82e1a3d34d870129dc0c677c6efb42d811) but haven't investigated more as I'm unfamiliar with all this machinery lol.

rolandwalker:
Ah, that's very helpful!

rolandwalker:
@wrmsr your analysis is great, thanks!  I tried out a different idea in #1305, and the tests pass.  My bumbling with env variables in #1241  was never going to work.

However, the tests were telling us something!  We allow setting a pager in the config file with arguments:

```
pager = 'less -iRSMwF --use-color'
```

and that is going to break with the new `click`, right?

wrmsr:
@rolandwalker >_< forgot to close the mkstemp fd, seems like linux won't exec an open-for-write file. Tested locally on 🍎 *and* 🐧 and it passes. Mind rerunning to confirm?

Not saying this is the proper fix if a configurable multi-token 'shelly' pager command is supported. At least for the very first breaking click version - the one with the diff I linked - I don't think that's doable without moving this kind of hack into mycli since the whole thing is `shlex.which`'d. However it seems on modern clicks - since [3/27 (8.2.0?)](https://github.com/pallets/click/commit/43a7d70f1e2a4a929af99b848612e0fd478e6e50#diff-f25fea0c0c214556563712a06f29270750a9b484a56d6a4df07ad5ae181b3eabR383) - it tries to `shlex.split` the pager env var and `shutil.which` only the first word, which looking at it *should* work, and yet didn't when I initially posted this branch with the latest click version. There's surely a way to get it working under that, I'll look at that later.

rolandwalker:
> However it seems on modern clicks - since [3/27 (8.2.0?)](https://github.com/pallets/click/commit/43a7d70f1e2a4a929af99b848612e0fd478e6e50#diff-f25fea0c0c214556563712a06f29270750a9b484a56d6a4df07ad5ae181b3eabR383) - it tries to shlex.split the pager env var

Great!  Then we only have to blocklist some range of versions.

rolandwalker:
Do you need me to handle the lint fixes?

rolandwalker:
I opened an issue with `click` based on your research: https://github.com/pallets/click/issues/3039

rolandwalker:
Closing in favor of #1396, after a `click` release containing the bugfix.  Many thanks @wrmsr !

PR #1246, opened 2025-07-04
Remove `requirements-dev.tx`t; prefer uv/`pyproject.toml`

rolandwalker:
## Description
 * remove `requirements-dev.txt`
 * verify that needed dependencies are covered in `pyproject.toml`.  The following are removed: twine, autopep8, pep8radius, and setuptools. The first three should be obsolete, and setuptools should be handled by the `build-system` stanza of `pyproject.toml`.
 * update `CONTRIBUTING.md`
 * update `MANIFEST.in`

## Checklist


rolandwalker:
I'm still not sure about the interaction between tox and uv, and whether we now really need tox.

amjith (approved):

amjith:
> I'm still not sure about the interaction between tox and uv, and whether we now really need tox.

`tox` is used to run the tests in multiple python versions. Tox will still use uv to install the dependencies.

rolandwalker:
Yes, it is weird that we are installing and running tox via uv, but that tox turns around and invokes uv again.

PR #1248, opened 2025-07-05
Trailing shell-style redirect syntax

rolandwalker:
## Description
Inspired by redis-cli, add trailing shell-style redirect syntax.  This mostly duplicates `\once` and `\pipe_once` in functionality but with much greater convenience.

Before:

```sql
\T csv;
\once -o name.csv;
select * from user where username = 'name';
\T ascii;
```

After

```sql
select * from user where username = 'name' $> name.csv;
```

Like the shell, a single angle bracket overwrites, and a double angle bracket appends.  The `$` character is needed to resolve ambiguity with SQL's built-in operators.

Limitation: the filename must not contain spaces or angle brackets.

Limitation: a trailing semicolon is still required in multiline mode.

Limitation: the input line must be a single statement.

The parsing algorithm is:
 * does the input contain a dollar-operator sequence, as tokenized by sqlglot?  This guarantees that the sequence is outside of a quote.
 * is the text to the left of the rightmost dollar-operator valid SQL?
 * is the text to the right of the rightmost dollar-operator free of space and angle-bracket characters?

If so, execute the SQL to the left of the operator and store the results to the file named after the right of the operator, appending or overwriting depending on the value of the operator.  Tilde expansion is applied to the filename.

Limitation: if any of the above conditions is not met, mycli will fall back to dispatching the command as plain SQL.  This means that error messages may not be clear when the user's intent is to use redirection, but the input is somehow not parseable.

A new setting `\Tr`, alias redirectformat, is introduced, making the redirected output format independent from the main interactive format. This defaults to CSV, which can be considered a breaking change, since previously the format was whatever the user had selected for interactive use.  The default for redirectformat can also be changed in myclirc.

`\once`, `\pipe_once`, and shell-style redirects now redirect content only to the designated output target and not to the TUI.  This can also be considered a breaking change.

tee is unchanged here, as the intention with tee may be different, considering that queries are also logged.

Support for a `|` operator is also included.  In this case, the shell portion may contain spaces.  The limitation is that there may be only one `|`, not a chained pipeline.  Still, one can do

```sql
select * from user where username = 'name' $| gh gist create;
```

or

```sql
select 100 as constant $| jq '. | length';
8
3
```

Documentation in this PR is a single line in `README.md`, and must be extended separately.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,7 @@ Features
 * Log every query and its results to a file (disabled by default).
 * Pretty prints tabular data (with colors!)
 * Support for SSL connections
+* Shell-style trailing redirects with `$>`, `$>>` and `$|` operators.
 * Some features are only exposed as [key bindings](doc/key_bindings.rst)
 
 Contributions:
@@ -150,7 +151,7 @@ https://github.com/dbcli/mycli/blob/main/CONTRIBUTING.md
 
 ## Additional Install Instructions:
 
-These are some alternative ways to install mycli that are not managed by our team but provided by OS package maintainers. These packages could be slightly out of date and take time to release the latest version. 
+These are some alternative ways to install mycli that are not managed by our team but provided by OS package maintainers. These packages could be slightly out of date and take time to release the latest version.
 
 ### Arch, Manjaro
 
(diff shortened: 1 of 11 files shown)


rolandwalker:
Grr, tests are passing locally, but not in CI.

amjith (approved):

PR #1252, opened 2025-07-09
Post-save command hook for redirected output

rolandwalker:
## Description
Let a command be executed after successful save-to-file by redirect or `\once`, only if `main.post_redirect_command` is set in .myclirc.

## Checklist


amjith:
Nit: `post_redirect_command` would be a slightly better name.

amjith (approved):

PR #1266, opened 2025-07-17
Allow trailing file redirects after pipe commands

rolandwalker:
## Description

Allow the form

```
SQL $| command $> capture.txt
```

Simple example:

```
select 10 $| tail -1 $> ten.txt
```

Changes:
 * Extract four parts in `get_redirect_components()` instead of three: `sql_part`, `command_part`, `file_operator_part`, `file_part`.
 * Log the redirect parse at DEBUG level.
 * Support post_redirect_command in `flush_pipe_once_if_written()`, factoring out `_run_post_redirect_hook()`.
 * Gather pipe_once globals into a single `PIPE_ONCE` dictionary.
 * Give leading underscores to unused variables.
 * Recast redirect parsing functions to all begin with `_redirect`.
 * When assembling tokens to a string, synthesize surrounding quotation marks, fixing a bug introduced in #1263 , but not released.
 * Reorder decorators on `get_redirect_components()` so that caching actually happens.

Notes:
 * Normally at the size of four elements I would return something other than a tuple from `get_redirect_components()`.  But this form is intuitive because they are strictly ordered.
 * Some elements of the `get_redirect_components()` tuple may be `None`s. Previously the result could be only all-strings or all-Nones.  Now if only a command is present, but the output is not captured to a file, the tuple will be mixed.  As before, `sql_part` is always None unless we can successfully parse out a redirect.
 * The redirect parsing really belongs in a separate file at this point.
 * Our reportformat (default CSV) only applies to the first operator seen.  After that what happens is up to the shell commands.  So while

    ```
    select 10 $> output.csv
    ```

   produces a CSV,

    ```
    select 10 $| wc $> output.txt
    ```

   does not.  There is really no other way for it to work, but this should be explained when documenting.
 * The more interesting the possible shell commands, the less sense it makes to have a hardcoded timeout of 60 seconds on the pipeline, as we do now.

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,7 +4,8 @@ Upcoming Release (TBD)
 Features
 --------
 
-* Support chained pipe operators.
+* Support chained pipe operators such as `select first_name from users $| grep '^J' $| head -10`.
+* Support trailing file redirects after pipe operators, such as `select 10 $| tail -1 $> ten.txt`.
 
 
 1.34.4 (2025/07/15)
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -709,10 +709,10 @@ class MyCli(object):
                 return
 
             if special.is_redirect_command(text):
-                sql_part, operator_part, shell_part = special.get_redirect_components(text)
+                sql_part, command_part, file_operator_part, file_part = special.get_redirect_components(text)
                 text = sql_part
                 try:
-                    special.set_redirect(shell_part, operator_part)
+                    special.set_redirect(command_part, file_operator_part, file_part)
                 except (FileNotFoundError, OSError, RuntimeError) as e:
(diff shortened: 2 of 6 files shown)


changelog.md
```diff
@@ -5,6 +5,7 @@ Features
 --------
 
 * Support chained pipe operators.
+* Support trailing file redirects after pipe operators.
```
amjith-openai: Could you add an example in the changelog itself? I feel like the PR description does a fantastic job of describing the feature with a single command. Adding that here would help.

rolandwalker: Done!  What we need at some point is a blog post with examples.

rolandwalker:
I also took my own advice about not returning a 4-tuple from `_redirect_find_token_indices()`, which now returns a dict.

PR #1278, opened 2025-07-19
Make control-r reverse search style configurable

rolandwalker:
## Description
Allow the old control-r incremental search to be recovered via a config file setting.  In case control-r is bound to traditional incremental search, add alt-r bindings for fzf search so that it is always available.

The "fzf" setting for control_r in myclirc remains a bit of a lie, because we still fall back to reverse_isearch if fzf is not installed. But it could be nice to emit a warning in that case.

This is intended to address the discussion comment at

 * https://github.com/dbcli/mycli/discussions/1265#discussioncomment-13791547

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,11 @@
 Upcoming Release (TBD)
 ======================
 
+Features
+--------
+* Make control-r reverse search style configurable.
+
+
 Internal
 --------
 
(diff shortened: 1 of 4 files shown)


lazmond3 (approved):

dbolser:
Does this fix my 'annoyance'? How do I test it?

rolandwalker:
It only attempts to fix your ctrl-r annoyance when `fzf` is installed.  Since you don't have it installed, I might not have understood the problem.

The test _would_ be to add this section to `~/.myclirc`:

```
[keys]
# possible values: auto, fzf, reverse_isearch 
control_r = reverse_isearch
```

PR #1293, opened 2025-07-26
Typehint `completion_engine.py`

rolandwalker:
## Description

 * add type hints
 * rewrite `identifies()` to return only a `bool`, not the empty string
 * prefix some unused variable names with underscore
 * bugfix: check that `token` is a `Token` before invoking `is_keyword()` on it

This completes typehinting in the packages subdirectory.

## Checklist

### DIFF
--- a/mycli/packages/completion_engine.py
+++ b/mycli/packages/completion_engine.py
@@ -1,13 +1,15 @@
-# type: ignore
+from __future__ import annotations
+
+from typing import Any
 
 import sqlparse
-from sqlparse.sql import Comparison, Identifier, Where
+from sqlparse.sql import Comparison, Identifier, Token, Where
 
 from mycli.packages.parseutils import extract_tables, find_prev_keyword, last_word
-from mycli.packages.special import parse_special_command
+from mycli.packages.special.main import parse_special_command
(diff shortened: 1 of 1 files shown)


amjith (approved):

mycli/packages/completion_engine.py
```diff
@@ -80,9 +82,9 @@ def suggest_type(full_text, text_before_cursor):
     return suggest_based_on_last_token(last_token, text_before_cursor, full_text, identifier)
 
 
-def suggest_special(text):
+def suggest_special(text: str) -> list[dict[str, Any]]:
     text = text.lstrip()
-    cmd, _, arg = parse_special_command(text)
+    cmd, _separator, _arg = parse_special_command(text)
```
amjith, line 87: Even if we're not using a var we still need to name it? Was it flagged by mypy?

rolandwalker: We don't have to here, but I did it for clarity.  There were other cases where `_` was used for two different types in a row, and mypy did get upset.  Instead, mypy should ignore `_`.

Happy to revert if you think that change is a net loss.

PR #1295, opened 2025-07-26
Typehint `completetion_refresher.py`

rolandwalker:
## Description
 * add typehints
 * import annotations for Python 3.9 compatibility
 * remove `OrderedDict` as we no longer support old Pythons
 * correct "pgexecute" comment
 * reformat some long type signatures

## Checklist


mycli/completion_refresher.py
```diff
@@ -1,21 +1,26 @@
...
+from typing import Callable
 
 from mycli.packages.special.main import COMMANDS
 from mycli.sqlcompleter import SQLCompleter
 from mycli.sqlexecute import ServerSpecies, SQLExecute
 
 
 class CompletionRefresher:
-    refreshers = OrderedDict()
+    refreshers: dict = {}
```
amjith, line 12: Changing from OrderedDict to Dict didn't change behavior? 

I'm trying to remember why this was an OrderedDict to begin with.

rolandwalker: It shouldn't change behavior because `dict` is ordered after Python 3.7.  There are some small differences with OrderedDict per https://docs.python.org/3/library/collections.html#collections.OrderedDict , but one difference is that `dict` is supposed to be more performant for the common case.

PR #1300, opened 2025-08-02
Typehint most of main.py, with consequent fixes elsewhere

rolandwalker:
## Description

 * typehint >80% of `main.py`
 * requiring additional hints and fixes in `config.py` and `sqlexecute.py`, `packages/parseutils.py`, and others
 * typehint unhinted portions of `sqlexecute.py`
 * remove `@export` decorator and import explicitly into `special`
 * recast some uninformative variable names
 * assert that a connection is available before using it
 * take more care with: ssl config, port, and ssh_port types/defaults before creating a connection
 * `local_infile` is a boolean, not a string
 * ensure `get_password_from_file()` returns a value
 * assert that a `PromptSession` is available before invoking methods on it
 * set a fallback col/row size in case a prompt session is not available
 * declare and type `self.conn` in `SQLExecute` much earlier
 * take more care to distinguish falsey values from `None`s in `SQLExecute` `connect()`
 * improve defaults passed to `pymysql.connect(`), matching them to the pymysql docs
 * test that `self.conn` is not `None` rather than checking for a `conn` attribute, before invoking `close()`
 * remove Python 2.x-compat Unicode literals trick for click
 * better catch empty database names in `change_db()`
 * take more care for edge-case values of `text` in `one_iteration()`
 * don't return `None` for `connection_id_to_kill`
 * clarify overlapping `e` variables
 * don't use `self.sqlexecute` when `sqlexecute` is already available
 * `is_dropping_database()` takes a string, not a list
 * typehint unhinted function in `packages/parseutils.py`
 * use only `chain()` for `format_output()` output
 * test `isinstance(…, Cursor`) rather than `hasattr(…, "description")`
 * set `start = time()` before `try` block
 * don't `return output_res(…)` when no return value is expected
 * update changelog

@amjith I tried to keep the cute `@export` decorator but it is too dynamic for a static checker to understand.

## Checklist

### DIFF
--- a/mycli/packages/special/__init__.py
+++ b/mycli/packages/special/__init__.py
@@ -1,19 +1,95 @@
 from __future__ import annotations
 
-from typing import Callable
-
-__all__: list[str] = []
-
-
-def export(defn: Callable):
-    """Decorator to explicitly mark functions that are exposed in a lib."""
-    globals()[defn.__name__] = defn
-    __all__.append(defn.__name__)
-    return defn
-
-
-from mycli.packages.special import (
-    dbcommands,  # noqa: E402 F401
-    iocommands,  # noqa: E402 F401
-    llm,  # noqa: E402 F401
+from mycli.packages.special.dbcommands import (
+    list_databases,
+    list_tables,
+    status,
 )
+from mycli.packages.special.iocommands import (
+    clip_command,
+    close_tee,
+    copy_query_to_clipboard,
+    disable_pager,
+    editor_command,
+    flush_pipe_once_if_written,
+    forced_horizontal,
+    get_clip_query,
+    get_current_delimiter,
+    get_editor_query,
+    get_filename,
+    is_expanded_output,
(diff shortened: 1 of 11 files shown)


mycli/packages/special/__init__.py
```diff
@@ -1,18 +1,83 @@
...
+    'set_redirect',
+    'set_timing_enabled',
+    'special_command',
+    'split_queries',
+    'status',
+    'unset_once_if_written',
+    'write_once',
+    'write_pipe_once',
+    'write_tee',
+]
```
amjith, line 95: Non-blocking question. Do you think we need to hoist all these functions to the top level? 

BTW, I'm fine with removing the `@export` decorator.

amjith (approved):

mycli/packages/special/__init__.py
rolandwalker: I don't think we do need to hoist.  It could be nicer not to.

PR #1302, opened 2025-08-04
Remove code duplication.

amjith:
Also use nonlocal for mutating.

## Description

Fixes the output_res() return signature.
Dedupes the code from output_res()

## Checklist

- [ ] I ran `uv ruff check && uv ruff format` to lint and format the code.

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -8,9 +8,6 @@ import shutil
 import sys
 import threading
 import traceback
-
-from prompt_toolkit import output
-
 try:
     from pwd import getpwuid
 except ImportError:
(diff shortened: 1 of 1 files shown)


mycli/main.py
```diff
@@ -700,7 +705,14 @@ def output_res(res, start):
...
+                    title,
+                    cur,
+                    headers,
+                    special.is_expanded_output(),
+                    special.is_redirected(),
+                    max_width,
+                )
 
                 t = time() - start
                 try:
```
rolandwalker, line 718: In this try block, we should probably also restore from the deleted code:

```
                        if self.beep_after_seconds > 0 and t >= self.beep_after_seconds:
                            self.bell()
                        if special.is_timing_enabled():
                            self.echo("Time: %0.03fs" % t)
```

rolandwalker (approved):

mycli/main.py
amjith: Thanks! Fixed it.

PR #1339, opened 2025-08-22
Only read "my" configuration files once, rather than once per call to…

fabzzap:
… read_my_cnf_files

## Description
Only read "my" configuration files once, rather than once per call to read_my_cnf_files

`read_my_cnf_files` used to call `read_config_files` every time.  Now, `read_config_files` is called once with parameter `self.cnf_files`. This way, it is avoided to re-read the same file(s) several times from the file system

## Checklist

- [ x I've added my name to the `AUTHORS` file (or it's already there).


rolandwalker:
This looks like a good idea!

rolandwalker:
Is this ready for review?

fabzzap:
Yes, the CI passes

rolandwalker:
Solid!

rolandwalker:
Released in v1.38.4

PR #1355, opened 2025-09-27
CI: test on Python 3.14

rolandwalker:
## Description
Test on Python 3.14, which is soon to be released.

## Checklist


amjith (approved):

rolandwalker:
This is currently blocked by a transitive dependency of the latest `llm` library.  `llm` will presumably catch up when Python 3.14 is released.

rolandwalker:
I'm not sure the above diagnosis was exactly correct, but the dependencies do build now.  However, there are some legitimate test failures.  The visible one here currently is a failure in watch_query, which runs too fast:
 * https://github.com/dbcli/mycli/actions/runs/18320417332/job/52171843662?pr=1355#step:8:455

But even if we skip that test, other behave tests fail:
* https://github.com/dbcli/mycli/actions/runs/18320133163/job/52171277946#step:8:975

```
Errored scenarios:
  test/features/auto_vertical.feature:4  auto_vertical on with small query
  test/features/basic_commands.feature:15  check our application_name
  test/features/basic_commands.feature:19  insert the date
  test/features/crud_table.feature:4  create, insert, select from, update, drop table
  test/features/crud_table.feature:24  select null values
  test/features/iocommands.feature:10  tee output from query
```

This is a surprising number of problems.  I can duplicate some of them locally.  Some of them seem at first glance that they might be related to our pinning of `click`.

rolandwalker:
Some of the `behave` tests appear to be failing due to changes in the stdlib around dedenting or similar.  This says that the tests are fragile, not that Python 3.14 is buggy in some way.  Still, we have to go through the tests one-by-one.

PR #1356, opened 2025-09-28
fix: ssl check wrong

Sherlock-Holo:
## Description

when sqlexecutre check if user use ssl, it just check the ssl variable is not None, but if user use ssl actually, the ssl map will have a key 'enable' and value will be True
this will fix when using plain connection mode(without ssl), even use correct password, mycli still return error Bad Handshake

we can confirm it by using wireshark to capture the login request package, with this fix, the `Switch to SSL after handshake` will be `not set`

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,5 @@
+* Fix ssl_context always created.
+
 1.38.4 (2025/09/06)
 ==============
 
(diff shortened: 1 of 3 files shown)


rolandwalker:
Hi!  Thanks!  Can you resolve the (trivial) doc conflicts?

PR #1357, opened 2025-09-28
Switch from pyaes to pycryptodomex

meeuw:
## Description

Switch from pyaes to pycryptodomex

https://src.fedoraproject.org/rpms/mycli/blob/rawhide/f/0002-Switch-from-pyaes-to-pycryptodomex.patch

## Checklist


rolandwalker:
This was interesting; why close?

meeuw:
it was based on an old master branch instead of the latest main branch, created as #1366

PR #1360, opened 2025-09-28
Python 3.13

meeuw:
## Description

Python 3.13 [fixes]

https://src.fedoraproject.org/rpms/mycli/blob/rawhide/f/0004-Python-3.13.patch

## Checklist

### DIFF
--- a/mycli/config.py
+++ b/mycli/config.py
@@ -103,14 +103,14 @@ def read_config_files(files: list[str | TextIOWrapper], list_values: bool = True
 def create_default_config(list_values: bool = True) -> ConfigObj:
     import mycli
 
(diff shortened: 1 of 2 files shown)


rolandwalker:
Hi!  Thank you for doing this systematically.

This one might be good, but would need a little more explanation.  It's not clear how avoiding `open_text()` would relate to Python 3.13.  It's also failing mypy as written.

meeuw:
:ok_hand: I cannot find any problems building mycli without this patch so I've created a PR to remove it from the package:

https://src.fedoraproject.org/rpms/mycli/pull-request/4

Maybe terjeros remembers what exactly failed.

rolandwalker:
Sounds like we can close?

PR #1361, opened 2025-09-28
Revert to sqlglot 5.1.3

meeuw:
## Description

Revert to sqlglot 5.1.3

https://src.fedoraproject.org/rpms/mycli/blob/rawhide/f/0005-Revert-to-sqlglot-5.1.3.patch

## Checklist


rolandwalker:
I suspect that downgrading `sqlglot` has to do with some particular constraints of the Fedora package manager/package collection, and propose that we close this.

PR #1362, opened 2025-09-28
Revert to older toml format

meeuw:
## Description

Revert to older toml format

https://src.fedoraproject.org/rpms/mycli/blob/rawhide/f/0006-Revert-to-older-toml-format.patch

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ dynamic = ["version"]
 description = "CLI for MySQL Database. With auto-completion and syntax highlighting."
 readme = "README.md"
 requires-python = ">=3.9"
-license = "BSD-3-Clause"
+license = { text = "BSD" }
 authors = [{ name = "Mycli Core Team", email = "mycli-dev@googlegroups.com" }]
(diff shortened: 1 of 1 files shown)


rolandwalker:
Hi again!

From https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files :

> As per [PEP 639](https://peps.python.org/pep-0639/) licenses should be declared with two fields:
>
> * license is an [SPDX license expression](https://packaging.python.org/en/latest/glossary/#term-License-Expression) consisting of one or more [license identifiers](https://packaging.python.org/en/latest/glossary/#term-License-Identifier).
> * license-files is a list of license file glob patterns.
>
> A previous PEP had specified license to be a table with a file or a text key, this format is now deprecated.

The previous PEP was https://peps.python.org/pep-0621/#license.

As this form is formally deprecated, we shouldn't switch mycli to it, though it is no big deal if a downstream packager must do so if they are, for instance, stuck on an old version of setuptools.

PR #1363, opened 2025-09-28
Relax click reqs

meeuw:
## Description

Relax click reqs

https://src.fedoraproject.org/rpms/mycli/blob/rawhide/f/0007-Relax-click-reqs.patch

## Checklist


rolandwalker:
This is actually a subtle issue in `click` which will be fixed in click release 8.3.1.  See https://github.com/pallets/click/pull/3055.

PR #1402, opened 2025-11-23
disconnect cleanly of server using pymysql close() method on connection

tmijieux:
## Description
When mycli exits, the server may (mariadb here) emit a warning in its logs when a client socket is closed without sending a final "mysql packet" with a quit command.

Example of the warning 
2025-11-23 17:31:41 225 [Warning] Aborted connection 225 to db: 'unconnected' user: 'app' host: 'localhost' (Got an error reading communication packets)

Though this warning does no real harm in this case, i think it is still desirable to avoid getting it, as it may be also triggered for other reason and thus may be misleading if other problem shows up for other reasons.

This PR add codes to call the pymysql.Connection.close() method instead of just letting the socket be closed by the os when the process exits.

## Checklist

- [X] ran `uv run tox` two tests failed but they were also failing on the main branch on my machine (something about ssh?)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -107,6 +107,11 @@ You can check this by running:
 $ readlink -f $(which ex)
 ```
 
+# Github PR checklist
+- add the contribution to the `changelog.md`
+- add your name to the `AUTHORS` file (or it's already there).
+- run `uv run ruff check && uv run ruff format && uv run mypy --install-types .`
+
 
 ## Releasing a new version of mycli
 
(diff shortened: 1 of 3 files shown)


rolandwalker:
Hi!  Thanks!  This looks like a good idea, and those local test failures were nothing to worry about.  Can you complete the checklist?

tmijieux:
hello, is that ok ?

 remark: I put question mark for date of next release ...

rolandwalker:
Looks great, thanks!

Issue #555, opened 2017-12-28
Support for \w and \W


santec:
Please support \W to show warnings after each statement and \w to hide them, just like regular mysql cli.

anthony-geoghegan:
Currently, `mycli` does not even show that warnings were generated. e.g., it prints only `1 row in set` while the official CLI client does indicate that there was a problem:

>     1 row in set, 1 warning (0.00 sec)

One of the biggest drawbacks with MySQL is that the server only warns the user when it truncates or corrupts data when silently converting from one data type to another.

----

This would also be useful as a configuration option to show warnings permanently, similar to how the default client has a `--show-warnings` option.

peterdutton:
+1 for this feature!

federico-razzoli:
More notes:
* I agree with anthony-geoghegan that there should be an option to show warnings by default.
* If there are many identical warnings, please group them (it happens with big INSERT SELECT)
* There should also be the ability to show warnings but not notes.

dveeden:
Looks like mycli uses https://github.com/PyMySQL/PyMySQL which doesn't seem to expose warnings in a good official API at the moment.

```python
>>> import pymysql
>>> c = pymysql.connect(host='127.0.0.1',port=4000,user='root')
>>> cur = c.cursor()
>>> cur.execute('SELECT 1 + "2 foo"')
1
>>> cur._result.warning_count
1
```

Looks like https://github.com/PyMySQL/PyMySQL/pull/1056 might fix that.

dveeden:
This could be a start, once something like this is in the offical API:

```patch
diff --git a/mycli/sqlexecute.py b/mycli/sqlexecute.py
index 9461438..b7c60a1 100644
--- a/mycli/sqlexecute.py
+++ b/mycli/sqlexecute.py
@@ -248,6 +248,8 @@ class SQLExecute(object):
             except special.CommandNotFound:  # Regular SQL
                 _logger.debug('Regular sql statement. sql: %r', sql)
                 cur.execute(sql)
+                if cur._result.warning_count > 0:
+                    print(f"UHOH, THERE ARE {cur._result.warning_count} WARNINGS")
                 while True:
                     yield self.get_result(cur)
 
```

hawkingrei:
This feature is extremely important.

rolandwalker:
Agreed, this would be great!

dveeden:
Looks like PyMySQL now exposes the warning count. 

What probably needs to be done:
[...]

PR #1413, opened 2025-12-19
[Feature] Support automatic showing of warnings after SQL execution (#555)

scottnemes:
## Description

Adds support for the automatic displaying of warnings after a SQL statement is executed. Addresses feature request from issue #555.

May be set with:
* Commands \W and \w
* In the config file with show_warnings
* With --show-warnings/--no-show-warnings on the command line

Note there is currently some errors from mypy about Cursor.warning_count, however I already submitted a PR to typeshed to get that updated so the errors will go away once the change makes it into mypy for types-PyMySQL.
https://github.com/python/typeshed/pull/15150#issuecomment-3673985375

## Checklist

### DIFF
--- a/mycli/sqlexecute.py
+++ b/mycli/sqlexecute.py
@@ -351,6 +351,7 @@ class SQLExecute:
             status = f'Query OK, {cursor.rowcount} row{plural} affected'
 
         if cursor.warning_count > 0:
+            plural = '' if cursor.warning_count == 1 else 's'
             status = f'{status}, {cursor.warning_count} warning{plural}'
 
         return (title, cursor if cursor.description else None, headers, status)


rolandwalker:
Outstanding!  Looks like the type stubs will be updated tomorrow.

your-diary:
For those who read [the release note of v1.42.0](https://github.com/dbcli/mycli/releases/tag/v1.42.0) to find this functionality and tried to use `SELECT INTO @var` with `--show-warnings` but got no automatic warning display, there is [a follow-up PR](https://github.com/dbcli/mycli/pull/1417) to complete the feature ~that has NOT been released yet~ (Edit: the PR is included in [v1.43.0](https://github.com/dbcli/mycli/blob/main/changelog.md#1430-20260102).)

The PR says

>  Currently the new automatic show_warnings logic does not work for queries that return zero results, like DDL.

If you want to test the functionality in v1.42.0, try this:

```sql
SELECT 1 + '0 foo';
```

PR #1416, opened 2025-12-23
[feat] Rework reconnect logic to actually create a new connection instead of only changing the database (#746)

### LINKED ISSUE #746: Make reconnect command create a new connection to the database
when mycli command line is idle for long time,then the connection is lost.When I try to reconnect using command
\r mydb
it takes time so long to wait for displaying reconnecting ....
once I see the Reconnecting.....  displayed
It connect to server so fast.
so,I wonder why it take me so long to wait,even slower a new command line " mycli -hhost -ppass -uroot "

scottnemes:
## Description

What:
Reworks the reconnect logic (both the command `\r` and calls from other parts of the code) to actually create a new database connection instead of simply running a command to change the database.

Why:
With the existing reconnect logic, some people would get long reconnect times. Plus the logic to reconnect was more indirect as it was a result of calling another command instead of specifically creating a new connection.

Resolves #746

## Checklist

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -214,7 +214,7 @@ class MyCli:
         special.register_special_command(self.change_db, "use", "\\u", "Change to a new database.", aliases=["\\u"])
         special.register_special_command(
             self.manual_reconnect,
-            "reconnect",
+            "connect",
             "\\r",
             "Reconnect to the database. Optional database argument.",
             aliases=["\\r"],
--- a/test/features/fixture_data/help_commands.txt
+++ b/test/features/fixture_data/help_commands.txt
@@ -13,6 +13,7 @@
 | \once          | \o [-o] filename           | Append next result to an output file (overwrite using -o). |
 | \pipe_once     | \| command                 | Send next result to a subprocess.                          |
 | \timing        | \t                         | Toggle timing of commands.                                 |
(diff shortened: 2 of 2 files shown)


mycli/main.py
```diff
@@ -213,8 +213,8 @@ def close(self) -> None:
     def register_special_commands(self) -> None:
         special.register_special_command(self.change_db, "use", "\\u", "Change to a new database.", aliases=["\\u"])
         special.register_special_command(
-            self.change_db,
-            "connect",
```
rolandwalker, line 217: Changing the full command from "connect" to "reconnect" is a needless breaking change for the user.  Maybe it is more rational since both the long and short versions start with "r"?  We should be able to add "reconnect" as an alias here: 

```
 aliases=["\\r", "reconnect"],
```

if that is the motivation.

scottnemes: Good point! I switched it back to connect. Goal was consistency, but your point makes sense and it matches the official client anyway

scottnemes: And RE: adding the alias, I pondered it but decided to leave it off for now. Theoretically in some future change we could split it out to actually be "connect" and "reconnect" with different logic (may never happen, but possible) so didn't want to further muddy the waters there.

PR #1417, opened 2025-12-24
[feat] Update query handling to allow automatic show_warnings to work for DDL

scottnemes:
## Description

What:
This PR updates the query processing functions to allow queries that return zero results, like DDL, to also hit the automatic show_warnings code path.

Why:
Currently the new automatic show_warnings logic does not work for queries that return zero results, like DDL.

## Checklist


rolandwalker (approved):

rolandwalker:
Nice followup!

PR #1418, opened 2025-12-24
[feat] Update SSL option to connect securely by default

### LINKED ISSUE #760: Access denied for user
I was trying to connect a remote mysql server. and I got this message

> (1045, u"Access denied for user 'oms-dev'@'10.1.9.18' (using password: YES)"

following is mysql server connection information

> Connection id:		2996639
SSL:			Cipher in use is DHE-RSA-AES256-SHA
Current pager:		less
Using outfile:		''
Using delimiter:	;
Server version:		5.7.22-log Source distribution
Protocol version:	10
Connection:		127.0.0.1 via TCP/IP
Server characterset:	latin1
Db     characterset:	utf8
Client characterset:	utf8
Conn.  characterset:	utf8
TCP port:		3307
Uptime:			248 days 20 hours 11 min 23 sec

scottnemes:
## Description

What:
Currently mycli does not connect via SSL by default, while the official MySQL client does.

Why:
Connecting via SSL by default is more secure.

Resolves #760

## Checklist

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -156,7 +156,7 @@ class MyCli:
         self.post_redirect_command = c['main'].get('post_redirect_command')
 
         # set ssl_mode if a valid option is provided in a config file, otherwise None
-        ssl_mode = c["ssl"].get("ssl_mode", None)
+        ssl_mode = c["main"].get("ssl_mode", None)
         if ssl_mode not in ("auto", "on", "off", None):
             self.echo(f"Invalid config option provided for ssl_mode ({ssl_mode}); ignoring.", err=True, fg="red")
             self.ssl_mode = None
@@ -1545,7 +1545,7 @@ def cli(
 
     if ssl_enable is not None:
         click.secho(
-            "Warning: The --ssl/--no-ssl CLI options will be deprecated in a future release. "
+            "Warning: The --ssl/--no-ssl CLI options are deprecated and will be removed in a future release. "
             "Please use the ssl_mode config or --ssl-mode CLI options instead.",
             err=True,
             fg="yellow",
--- a/mycli/myclirc
+++ b/mycli/myclirc
@@ -5,6 +5,13 @@
 # after executing a SQL statement when applicable.
 show_warnings = False
 
+# Sets the desired behavior for handling secure connections to the database server.
+# Possible values:
+# auto = SSL is preferred. Will attempt to connect via SSL, but will fallback to cleartext as needed.
+# on = SSL is required. Will attempt to connect via SSL and will fail if a secure connection is not established.
+# off = do not use SSL. Will fail if the server requires a secure connection.
+ssl_mode = auto
+
 # Enables context sensitive auto-completion. If this is disabled the all
 # possible completions will be listed.
 smart_completion = True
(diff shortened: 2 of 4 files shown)


rolandwalker:
While I favor this, there is a lot to consider.

The vendor client does not simply use SSL/TLS by default, but tries encrypted transport first, then retries without encryption on failure (https://dev.mysql.com/doc/refman/9.0/en/encrypted-connections.html).  That's probably the minimum we need to do, to avoid a flood of incoming issues.

To help the user out, and to do better than the vendor client, we could also have a config-file option for encrypted transport with three states: on/off/auto, with the default of "auto".

`_connect()` already uses a `try`/`except` formulation to work out whether a password prompt is needed:
 * https://github.com/dbcli/mycli/blob/3683b9fa4a01360e0f5e22d7c8983a376b2aacb4/mycli/main.py#L508
 so it might get complicated to also do the same thing for encrypted transport.

scottnemes:
> While I favor this, there is a lot to consider.
> 
> The vendor client does not simply use SSL/TLS by default, but tries encrypted transport first, then retries without encryption on failure (https://dev.mysql.com/doc/refman/9.0/en/encrypted-connections.html). That's probably the minimum we need to do, to avoid a flood of incoming issues.
> 
> To help the user out, and to do better than the vendor client, we could also have a config-file option for encrypted transport with three states: on/off/auto, with the default of "auto".
> 
> `_connect()` already uses a `try`/`except` formulation to work out whether a password prompt is needed:
> 
> * https://github.com/dbcli/mycli/blob/3683b9fa4a01360e0f5e22d7c8983a376b2aacb4/mycli/main.py#L508
>   
>   so it might get complicated to also do the same thing for encrypted transport.

Good callout. I actually thought it would downgrade automatically as long as the server wasn't requiring SSL, so definitely did not expect that; I will rework that bit.

RE: the config, I did think about adding a config file option but thought it might be scope creep. But if you also think it's a good idea I will work on that bit as well!

rolandwalker:
> I actually thought it would downgrade automatically as long as the server wasn't requiring SSL

Ah, well I haven't check pymysql docs/internals before writing that — maybe the library does!

scottnemes:
> > I actually thought it would downgrade automatically as long as the server wasn't requiring SSL
> 
> Ah, well I haven't check pymysql docs/internals before writing that — maybe the library does!

It doesn't apparently; that's why I was surprised anyway, hah! So it will need a decent amount of work to add in the config and retry logic. Will take a look though.

mycli/main.py
```diff
@@ -1510,6 +1542,15 @@ def cli(
         warn=warn,
         myclirc=myclirc,
     )
+
+    if ssl_enable is not None:
+        click.secho(
+            "Warning: The --ssl/--no-ssl CLI options will be deprecated in a future release. "
```
rolandwalker: Can we say "are deprecated and will be removed in a future release"?

mycli/myclirc
```diff
@@ -190,3 +190,11 @@ output.null = "#808080"
 [alias_dsn.init-commands]
 # Define one or more SQL statements per alias (semicolon-separated).
 # example_dsn = "SET sql_select_limit=1000; SET time_zone='+00:00'"
+
+[ssl]
```
rolandwalker: Are other SSL/TLS settings expected in the `[ssl]` section?  If not, should this be part of `[main]`?  (If so, a new section makes sense.)

PR #1423, opened 2026-01-02
Configurable string for missing_value in outputs

### LINKED ISSUE #1120: How to set the null string?
pgcli gives an option:
```
null_string = 'NULL'
```

rolandwalker:
## Description
The default is `<null>`, meaning no change.

This is configurable in pgcli, and there is no reason not to make it configurable here.

The setting only works for tabular outputs which default to the `cli_helpers` default.  Thus the setting does not change the behavior of CSV or JSON formats.

Fixes #1120 .

## Checklist

### DIFF
--- a/mycli/myclirc
+++ b/mycli/myclirc
@@ -50,6 +50,11 @@ table_format = ascii
 # Recommended: csv.
 redirect_format = csv
 
+# How to display the missing value (ie NULL).  Only certain table formats
+# support configuring the missing value.  CSV for example always uses the
+# empty string, and JSON formats use native nulls.
+null_string = <null>
+
 # A command to run after a successful output redirect, with {} to be replaced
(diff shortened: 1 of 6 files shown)


mycli/myclirc
```diff
@@ -50,6 +50,11 @@ table_format = ascii
 # Recommended: csv.
 redirect_format = csv
 
+# How to display the missing value (ie NULL).  Only certain table formats
+# support configuring the missing value.  CSV for example always uses the
+# empty string, and JSON formats use native nulls.
+missing_value = <null>
```
amjith: We called it `null_string` in pgcli. If you're not strongly opinionated about the config name we should try and match the names between the projects.

rolandwalker: Good idea, fixed!

PR #1428, opened 2026-01-06
[fix] Update watch command execution time to be correct for all iterations (#763)

### LINKED ISSUE #763: Query time output when using watch includes the watch time
One thing I've noticed when using the watch command in mycli is that the query time reported from the second and subsequent queries includes the time spent waiting by watch, rather than just the time spent running the query.

For example:

```
mysql acp@web-mfdbmaster:mobileforms> watch select 1;                                                                                                                                                                                            
> select 1
+-----+
| 1   |
|-----|
| 1   |
+-----+
Time: 0.024s

> select 1
+-----+
| 1   |
|-----|
| 1   |
+-----+
Time: 5.025s

> select 1
+-----+
| 1   |
|-----|
| 1   |
+-----+
Time: 5.034s
```

Not a huge issue as I can deduct the default five seconds waiting in my head, but it has caught me out once or twice and led me to spending a moment debugging a long running query when it actually didn't run all that long! :)

Dan

scottnemes:
## Description
Currently the watch command execution time is wrong for the 2nd+ iteration as it includes the execution time plus the watch command time. This PR fixes that issue so the correct execution time is reported for all watch iterations.

Fixes #763

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,6 +5,10 @@ Internal
 --------
 * Create new data class to handle SQL/command results to make further code improvements easier
 
+Bug Fixes
+--------
+* Update watch query output to display the correct execution time on all iterations (#763).
+
 
 1.44.1 (2026/01/10)
 ==============
@@ -21,10 +25,6 @@ Features
 --------
 * Add enum value completions for WHERE/HAVING clauses. (#790)
 * Add `show_favorite_query` config option to control query printing when running favorite queries. (#1118)
-
-Bug Fixes
---------
-* Update watch query output to display the correct execution time on all iterations (#763).
 
 
 1.43.1 (2026/01/03)


mycli/main.py
```diff
@@ -817,14 +817,30 @@ def show_suggestion_tip() -> bool:
...
 
-        def output_res(res: Generator[tuple], start: float) -> None:
+        def output_res(results: Generator[tuple], start: float) -> None:
             nonlocal mutating
             result_count = 0
-            for title, cur, headers, status in res:
+            for result in results:
+                # unpack the results, including the command info if present
+                try:
+                    title, cur, headers, status, command = result
```
rolandwalker: While this works, it is tricky. It would be much much better to just bite the bullet and switch to a dataclass instead of a tuple for the result.  I can work on that in a separate PR if you like.

scottnemes: Hah I figured you'd want a dataclass first! I'll take a stab at it and let you know if I get stuck.

scottnemes:
@rolandwalker Let me know if this looks better. I was hoping to not have to include it in the iterator, but I do not see another way around that unless we get rid of the iterator bit to begin with. Could potentially just next(res) it each time and then use the dataclass without iteration (i.e. no for loop at all), but not sure if that is any less hacky or would even work. So if you know a better way to do that let me know

changelog.md
```diff
@@ -19,10 +19,13 @@ Bug Fixes
 
 Features
 --------
-
 * Add enum value completions for WHERE/HAVING clauses. (#790)
 * Add `show_favorite_query` config option to control query printing when running favorite queries. (#1118)
 
+Bug Fixes
```
rolandwalker: This seems to be in the wrong part of the document.

mycli/packages/special/iocommands.py
```diff
@@ -580,11 +580,15 @@ def watch_query(arg: str, **kwargs) -> Generator[SQLResult, None, None]:
             set_pager_enabled(False)
             for sql, title in sql_list:
                 cur.execute(sql)
+                command = {
```
rolandwalker, line 583: Why a dict instead of two new properties?

scottnemes: Didn't want to add a bunch of random properties (presuming other commands will need other args), so figured a general command property with the dict that could be added to as needed would be better. But if you prefer having standalone properties for it all I can do that.

changelog.md
scottnemes: Will fix; merged it wrong apparently.

mycli/packages/special/iocommands.py
rolandwalker: Your implementation, your call.

scottnemes: Okay I'll keep it as-is then. Don't want to make too many assumptions about future use since this is the first use case, and it can be adjusted later if we find another approach makes more sense given new use cases. Current imagined use case is other commands needing special args, and it fits that anyway.

PR #1434, opened 2026-01-10
[internal] Create new data class for handling SQL/command results to make further code improvements easier

scottnemes:
## Description

This PR creates a new dataclass to store SQL/command results instead of passing around tuples. This should make further code improvements / extensions easier to maintain.

## Checklist

### DIFF
--- a/mycli/packages/sqlresult.py
+++ b/mycli/packages/sqlresult.py
@@ -6,12 +6,12 @@ from pymysql.cursors import Cursor
 @dataclass
 class SQLResult:
     title: str | None = None
-    cursor: Cursor | list[tuple] | None = None
+    results: Cursor | list[tuple] | None = None
     headers: list[str] | str | None = None
     status: str | None = None
 
     def __iter__(self):
-        return iter((self.title, self.cursor, self.headers, self.status))
+        return iter((self.title, self.results, self.headers, self.status))
 
     def __str__(self):
-        return f"{self.title}, {self.cursor}, {self.headers}, {self.status}"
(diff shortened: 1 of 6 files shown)


scottnemes:
@rolandwalker Let me know what you think! Went better than expected but not sure what you were expecting. Once this goes in I'll update the watch query PR; should hopefully be easier with this.

mycli/packages/sqlresult.py
```diff
@@ -0,0 +1,17 @@
+from dataclasses import dataclass
+
+from pymysql.cursors import Cursor
+
+
+@dataclass
+class SQLResult:
+    title: str | None = None
+    cursor: Cursor | list[tuple] | None = None
```
rolandwalker: Feels wrong to call this property `cursor` when it can also hold a list.  What would be a better name?  `result`?  Or is that repetitive?

scottnemes: Two options here:
1) Change the name to "results" and leave everything as is, with the value being overloaded; used as a Cursor sometimes, a list[tuple] other times in the code.

2) Do another full refactor now to split it into two fields on SQLResult, cursor and results. Then update all the downstream logic that uses the old field to use the correct new field. This will be a fairly big lift as there is a lot of random logic that uses the overloaded field currently. Should be done at some point either way.

I'm fine with either depending on your vision and what you'd like to happen now.

rolandwalker: Option 1, with some name like "results" or "contents", at your choice.

rolandwalker:
Thank you!

PR #1438, opened 2026-01-11
[chore] Rework SQLResult dataclass to not require all fields when used in for loops

scottnemes:
## Description

Initial goal of the SQLResult dataclass was to be able to use it without having to add all new attributes (i.e. command) to the for loops of every section of the code that uses the basic title/cursor/headers/status fields. This PR reworks the initial implementation to meet that goal.

## Checklist

### DIFF
--- a/mycli/packages/special/iocommands.py
+++ b/mycli/packages/special/iocommands.py
@@ -580,9 +580,9 @@ def watch_query(arg: str, **kwargs) -> Generator[SQLResult, None, None]:
             set_pager_enabled(False)
             for sql, title in sql_list:
                 cur.execute(sql)
-                command = {
+                command: dict[str, str | float] = {
                     "name": "watch",
-                    "seconds": str(seconds),
+                    "seconds": seconds,
                 }
                 if cur.description:
                     headers = [x[0] for x in cur.description]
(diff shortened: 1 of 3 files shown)


scottnemes:
@rolandwalker Had a moment of (I think) insight on how to meet my initial goal for this, which was to not have to use all SQLResult fields (i.e. command) in the for loops as that would get out of hand quick on future additions. This also makes things a bit more consistent. Let me know if you see any problems with it though, however it appears to work as I would expect anyway. Didn't update the changelog, as if this gets included it can just be part of the initial dataclass change anyway.

mycli/packages/special/iocommands.py
```diff
@@ -582,7 +582,7 @@ def watch_query(arg: str, **kwargs) -> Generator[SQLResult, None, None]:
                 cur.execute(sql)
                 command = {
                     "name": "watch",
-                    "seconds": seconds,
+                    "seconds": str(seconds),
```
rolandwalker: Why `str()` here and `float()` later?

scottnemes: In the weeds on trying to make mypy happy. I reworked it a bit to  be more explicit and found a combo mypy liked, so let me know if that's any better.

rolandwalker: Better!

rolandwalker:
I would work toward not needing `get_output()` at all and using the dataclass properties directly.  But it is good for now!

PR #1439, opened 2026-01-13
Set correct database on reconnect when applicable.

### LINKED ISSUE #1437: [bug] Database is not set after a reconnect
Currently when your connection is lost and a reconnect occurs, mycli thinks you are still connected to a database, while the connection itself does not:

```
MySQL scottn@localhost:(none)> use dev;
You are now connected to database "dev" as user "scottn"
Time: 0.001s
MySQL scottn@localhost:dev> select * from t2;
+----+
| id |
+----+
| 2  |
| 2  |
+----+

2 rows in set
Time: 0.004s
MySQL scottn@localhost:dev> select * from t2;
Reconnecting...
Reconnected successfully.
Any session state was reset.
(1046, 'No database selected')
MySQL scottn@localhost:dev>
```

This should be fixable by using the `connection.select_db()` method after the `connection.ping(reconnect=True)` is sent.

scottnemes:
## Description
Currently on reconnect the database connection loses the previously selected database (if applicable), however mycli still thinks the database is selected. This PR addresses the issue by telling the new connection what database to select after a reconnect, which brings the connection and mycli back in sync.

Fixes #1437 

## Checklist


scottnemes:
@rolandwalker This one should be good now. The test reliably reproduces the issue so will help with any future changes as well.

rolandwalker (approved):
Great!

PR #1445, opened 2026-01-17
Downgrade to paramiko 3.5.1

rolandwalker:
## Description

Paramiko 4.x no longer supports DSA keys, and can cause MyCLI to exit with a cryptic error if the user has a DSA key present:

    module 'paramiko' has no attribute 'DSSKey'

See https://github.com/paramiko/paramiko/issues/2537

## Checklist


rolandwalker:
> assuming we want to keep supporting DSA

The issue is that MyCLI simply exits if the user has a DSA key under `~/.ssh/`.  There's no way to just not support those particular keys.  MyCLI won't start, which is user-hostile.  Maybe there is another way to solve this, but I didn't find it.

PR #1446, opened 2026-01-19
Make destructive-warning keywords configurable

### LINKED ISSUE #708: Add a confirmation prompt for DML.
Can we add a confirmation prompt for every INSERT,UPDATE,DELETE,MODIFY script. This helps while connected to production environment.

rolandwalker:
## Description

Make destructive-warning keywords configurable.

Fixes #708 .

## Checklist


mycli/packages/parseutils.py
```diff
@@ -264,15 +264,18 @@ def query_has_where_clause(query: str) -> bool:
...
+def is_destructive(keywords: list[str], queries: str) -> bool:
+    """Returns True if any of the queries in *queries* is destructive."""
     for query in sqlparse.split(queries):
-        if query:
-            if query_starts_with(query, list(keywords)) is True:
-                return True
-            elif query_starts_with(query, ["update"]) is True and not query_has_where_clause(query):
+        if not query:
+            continue
+        if query_starts_with(query, keywords) and query_starts_with(query, ["update"]):
```
scottnemes, line 273: Maybe add a comment here clarifying that the first part of the IF statement is to check if UPDATE is in keywords to begin with. Took me a minute to figure out why it was using query_starts_with twice instead of just saying `"update" in keywords`. Which you can't do as-is currently due to case issues (would have to either lower case or upper case all keywords from the user to match the defaults).

rolandwalker: Good idea, done.

scottnemes (approved):

PR #1447, opened 2026-01-19
Import MySQL completion candidates from `pygments`

rolandwalker:
## Description
Since the `pygments` library is already required, why not use its list of reserved words for completions?

This fixes #767 and should keep us up-to-date. It's really great to have the `JSON_*` functions.

Obsoletes #925, and addresses the two points in https://github.com/dbcli/mycli/pull/925#issuecomment-754743293 .

Compared to the previous code, this PR

 * removes `LEN` and `TOP`, which are not MySQL reserved words
 * preserves extra completion candidates containing space, such as `ORDER BY`

Downsides and bugs:

* `pygments.lexers._mysql_builtins` does contain a leading underscore, so we should be aware that the library reserves the right to break this usage. The library version has been more tightly defined to remediate issues here.
* Similarly, certain tests might become more brittle with regard to library updates.
* Certain reserved words are duplicated with special commands, so if the first word of a command-line is any of the following, duplicated completions will show, as both upper- and lower-case: `exit`, `help`, `source`, `status`, `system`, `use`. This is fixable, but should we prefer the upper- or lower-case flavor?
* There are _many_ more completion candidates now, which may inspire us to do further work soon on prioritizing which completions are seen at the top.

## Checklist


scottnemes:
> This is fixable, but should we prefer the upper- or lower-case flavor

I always prefer lowercase, but as long as it's consistent; i.e. all options, regardless of source, are uppercase or lowercase ideally.

scottnemes (approved):

mycli/sqlcompleter.py
```diff
@@ -18,141 +19,27 @@
...
-        "VALUES",
-        "VARCHAR",
-        "VIEW",
-        "WHEN",
-        "WITH",
+        x.upper()
+        for x in favorite_keywords
+        + list(MYSQL_DATATYPES)
+        + list(MYSQL_KEYWORDS)
+        + ['ALTER TABLE', 'CHANGE MASTER TO', 'CHARACTER SET', 'FOREIGN KEY']
```
scottnemes, line 40: While looking at your follow-up PR, I noticed these get duped, I.e.:

image

image

Is that desired?

scottnemes: Never mind I realized it makes sense in the context of you technically don't know if they want just the keyword `master`, or the command `change master to` in this example. Was overthinking it since there is no valid "change master" command, but that is another layer of context.

rolandwalker:
> > This is fixable, but should we prefer the upper- or lower-case flavor
> 
> I always prefer lowercase, but as long as it's consistent; i.e. all options, regardless of source, are uppercase or lowercase ideally.

I'll make lowercase duplicates preferred in a followup PR.

PR #1450, opened 2026-01-20
Stream input from stdin, rather than reading the entire script into memory.

### LINKED ISSUE #1184: How to Emulate --quick and --unbuffered options from MySQL Client
The MySQL client has the options: `--quick and --unbuffered` that are extremely useful when pulling large amounts of data. This essentially avoids putting all the results into memory which could cause OOM for the client since it writes the output immediately. Is there an equivalent options for mycli for this?

rolandwalker:
## Description
 * Stream STDIN input, running queries a line at a time.
 * Remove `MemoryError` check, and recommendation for the vendor client.
 * Use CSV/TSV formats with headers for the first line only.
 * Exit with an error code if we are unable to open `/dev/tty`.
 * Add `--noninteractive` flag to suppress the destructive-warning prompt from the CLI.
 * Add `--format=` option to control output formats, leaving the default format as "extra headers pseudo TSV".
 * Commentary on edge cases and followups.

Fixes #1184.

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,6 +5,7 @@ Features
 --------
 * Add `--unbuffered` mode which fetches rows as needed, to save memory.
 * Default to standards-compliant `utf8mb4` character set.
+* Stream input from STDIN to consume less memory, adding `--noninteractive` and `--format=` CLI arguments.
 
 
 Bug Fixes
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -1517,8 +1517,8 @@ class MyCli:
 @click.option(
     "--show-warnings/--no-show-warnings", "show_warnings", is_flag=True, help="Automatically show warnings after executing a SQL statement."
 )
-@click.option("-t", "--table", is_flag=True, help="Display batch output in table format.")
-@click.option("--csv", is_flag=True, help="Display batch output in CSV format.")
+@click.option("-t", "--table", is_flag=True, help="Shorthand for --format=table.")
+@click.option("--csv", is_flag=True, help="Shorthand for --format=csv.")
 @click.option("--warn/--no-warn", default=None, help="Warn before running a destructive query.")
 @click.option("--local-infile", type=bool, help="Enable/disable LOAD DATA LOCAL INFILE.")
 @click.option("-g", "--login-path", type=str, help="Read this path from the login file.")
@@ -1532,6 +1532,10 @@ class MyCli:
     "--password-file", type=click.Path(), help="File or FIFO path containing the password to connect to the db if not specified otherwise."
 )
 @click.argument("database", default=None, nargs=1)
+@click.option("--noninteractive", is_flag=True, help="Don't prompt during batch input.  Recommended.")
+@click.option(
+    '--format', 'batch_format', type=click.Choice(['default', 'csv', 'tsv', 'table']), help='Format for batch or --execute output.'
+)
 @click.pass_context
 def cli(
     ctx: click.Context,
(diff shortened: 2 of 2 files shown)


rolandwalker:
> Use CSV/TSV formats with headers for the first line only.

The one breaking change here is that previously, the header lines would have been printed for every query in the script, leading to every other line being a header.  It seems improbable that anyone would be depending on that behavior.

Edit: this description of the previous behavior may not be correct, and needs checking.

Edit: the description of the previous behavior was correct.

scottnemes:
What's the intended use case of only showing the header for the first query? That seems to be more confusing at first glance, as you lose the details of what is actually happening. I.e.:

test.sql queries:
```
select count(*) from dev.t1;
select * from dev.t2 limit 5;
select count(*) from dev.t3;
```

command
```
uv run mycli < test.sql
```

Current behavior (able to see the header for each, so you know what the values are for:
```
count(*)
19
id
2
2
1
2
1
count(*)
1
```

New behavior:
```
count(*)
19
2
2
1
2
1
1
```

Makes it look like all the values are count(*)s due to only the one header.

rolandwalker:
> What's the intended use case of only showing the header for the first query?

The intended use case is a uniform script which produces the same columns for each line.  I think that's the most common usecase when scripting!  But maybe we can make it more configurable.

scottnemes:
> The intended use case is a uniform script which produces the same columns for each line. I think that's the most common usecase when scripting! But maybe we can make it more configurable.

Hmm I'd have to see some examples of what you mean. Anything I can think of I'd want to preserve the headers for each query. I.e. for a migration script, there might be various selects to check different things, some updates/inserts, etc. where I'd expect each query to come back with whatever headers it needed. If you're actually doing something with the output, i.e. creating a CSV, it's going to be in one query, and only have one header then regarless.

rolandwalker:
```
SELECT data FROM table WHERE id >= 1 AND id < 1000;
SELECT data FROM table WHERE id >= 1000 AND id < 2000;
SELECT data FROM table WHERE id >= 2000 AND id < 3000;
…
```

As a data practitioner, this pattern is common.  But again, we can finesse it with the CLI args.

scottnemes:
> ```
> SELECT data FROM table WHERE id >= 1 AND id < 1000;
> SELECT data FROM table WHERE id >= 1000 AND id < 2000;
> SELECT data FROM table WHERE id >= 2000 AND id < 3000;
> …
> ```
> 
> As a data practitioner, this pattern is common. But again, we can finesse it with the CLI args.

Huh interesting. I'd expect to do that in a loop; putting that all in a SQL file (and running as a combo of queries at once) seems like an anti-pattern.

scottnemes:
> SELECT data FROM table WHERE id >= 1 AND id < 1000;
> SELECT data FROM table WHERE id >= 1000 AND id < 2000;
> SELECT data FROM table WHERE id >= 2000 AND id < 3000;

Also the vendor client works as mycli currently does, with headers for each query. I know we are expanding on the vendor client, but in this case I prefer the current behavior that matches with headers per query.

Edited to add; if you are planning to add an option to control this, I would say have the current behavior be the default at least.

rolandwalker:
My main issue is with the specification of `--csv`.  Interspersed headers are not CSV.  So the current (weird) behavior can be the default.

rolandwalker:
@scottnemes updated to leave the default output format unchanged.

scottnemes:
Is the plan to deprecate the csv/table options eventually in lieu of the batch_mode option or keep them all? If keeping, I'm wondering if instead of exposing batch_mode to the user it would make sense to instead just add a --tsv option, so you'd have --csv/--tsv/--table. Probably would still need the batch_mode variable in the code as you have it, just not user defined (if csv/tsv/table isn't set, batch_mode var would be default).

That aside, everything appears to work as expected, so if you want to keep this implementation then it's good to go!

rolandwalker:
The plan for `--format=` was not to deprecate, but to be able to support any of other many formats available without adding toplevel options.

PR #1451, opened 2026-01-20
Let `GRANT ... ON` forms offer schema names as completions

### LINKED ISSUE #578: No autocomplete while grant databases access
All autocomplete works fine, except when I try to create grant access, it doesn't autocomplete the database name.
Example (no autocomplete):
`> GRANT ALL ON database_name.* TO 'db_user'@'localhost'`

I thought it will autocomplete like when you use USE command (selecting database)
Example (with autocomplete):
`> USE database_name`

=== 

Below my environtment:
- mycli v1.5.2 (install via apt-get)
- mysql v15.1 Distrib 10.2.12-MariaDB
- Ubuntu 16.04.3 LTS

rolandwalker:
## Description

Let `GRANT ... ON` forms offer schema names as completions instead of only table names.

Fixes #578 .

As the commentary notes, the test fixtures are insufficient to test the change, but it can be verified interactively.  See image.

One thing that is confusing in the code is the use of "database" vs "schema".  Here the "type" is "database".

last image

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -13,6 +13,7 @@ Bug Fixes
 * Respect `--logfile` when using `--execute` or standard input at the shell CLI.
 * Gracefully catch Paramiko parsing errors on `--list-ssh-config`.
 * Downgrade to Paramiko 3.5.1 to avoid crashing on DSA SSH keys.
+* Offer schema name completions in `GRANT ... ON` forms.
 
 
 1.44.2 (2026/01/13)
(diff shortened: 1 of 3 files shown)


rolandwalker:
> Future potential idea

Yes, it looks like there is a lot of room for improvement, especially if we used `sqlglot` to more deeply examine the structure of the query.  Though we might find out that `sqlglot` is too strict for the case of partial queries.

PR #1452, opened 2026-01-20
[chore] Add new connection scheme check function and updated password check logic to use it

scottnemes:
## Description
This adds additional logic to the password check to try and prevent any DSN URIs from being considered as the password. Function may also be used later to actually apply or reject provided scheme's if/when we want to go that route.

## Checklist

### DIFF
--- a/mycli/packages/parseutils.py
+++ b/mycli/packages/parseutils.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import re
-from typing import Any, Generator, Tuple
+from typing import Any, Generator
 
 import sqlglot
 import sqlparse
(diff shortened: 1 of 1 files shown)


mycli/packages/parseutils.py
```diff
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import re
-from typing import Any, Generator
+from typing import Any, Generator, Tuple
```
rolandwalker: Capitalized `Tuple` isn't needed with newer Pythons.   Just `tuple` will do.

scottnemes: Fixed, thanks!

PR #1456, opened 2026-01-21
Put special commands first in the list of completion candidates

rolandwalker:
## Description

Put special commands first in the list of completion candidates and suppress any duplication between pygments keywords and special commands, preferring special commands over pygments keywords.

Example: whereas "exit" previously appeared twice, once as "EXIT" and once as "exit", it now appears only once, and at the top of the completion candidates.

Followup to #1447.

last image

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -10,6 +10,7 @@ Features
 Bug Fixes
 --------
 * Fix CamelCase fuzzy matching.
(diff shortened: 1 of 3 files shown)


scottnemes (approved):
Works for me!

PR #1457, opened 2026-01-21
Add true fuzzy-match completions with rapidfuzz

rolandwalker:
## Description
 * Don't attempt true fuzzy matches until the given text is 4+ characters.
 * Require a rapidfuzz WRatio score of ~80+~ 75+.
 * Limit rapidfuzz candidates to 20 or fewer.
 * Don't report rapidfuzz candidates which are much shorter than the given text.

Example:
last image

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -9,6 +9,7 @@ Features
 * Remove suggested quoting on completions for identifiers with uppercase.
 * Allow table names to be completed with leading schema names.
 * Soft deprecate the built-in SSH features.
+* Add true fuzzy-match completions with rapidfuzz.
 
 
 Bug Fixes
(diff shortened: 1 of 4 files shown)


scottnemes:
Could you provide some more examples showing the difference? The `JON_V` example acts the same on main, so wanting to test out some additional examples to see how it works.

rolandwalker:
Oh, right.  Try `JZON_VAL`.

rolandwalker:
`JZON_VAL` actually doesn't match.  The thresholds might need some tuning.  This one does:

last image

rolandwalker:
Lowering `score_cutoff=80` to 75 lets `JZON_VAL` match so I made that update.

scottnemes:
Nice! Noticed that the suggestion comes and goes. Posting here in case you think it's related to these changes. The suggestion disappears once it hits `jzon_va` and then returns when you get to `jzon_val`:

image

image

image

image

image

rolandwalker:
Yes, there are some janky behaviors like that.  You can change the `scorer` to one of these https://rapidfuzz.github.io/RapidFuzz/Usage/fuzz.html, or reduce the `score_cutoff`, to force `JSON_VALUE` to stay as a candidate the whole way, but generally the overall results get worse, in the form of additional poor-quality candidates.

Maybe we will ultimately make the scorer and the cutoff configurable in `~/.myclirc` because tastes are going to differ.

PR #1458, opened 2026-01-21
Remove quoting on uppercase-identifier completions

### LINKED ISSUE #620: Is it possible to disable quoting on auto-completed fields
I read the [FAQ](https://www.mycli.net/faq) and the [Docs](https://www.mycli.net/docs) but I didn't find anything about this issue. 

Is it possible to disable quoting on auto-completed fields that do not require quoting? For example if I type:

```
SELECT CustID FROM C<tab>
```

I end up with this SQL:

```
SELECT CustID FROM `CustInfo`
```

The table name `CustInfo` doesn't need quoting, but `mycli` adds quoting to **every** field regardless. Is there a setting that I can apply to `mycli` will only quote fields that require quoting?

rolandwalker:
## Description

Mandatory quoting is retained for identifiers which have any non-alphanumeric characters, but an uppercase character alone does not activate the suggested quoting.

The original logic looks intentional but doesn't seem to be needed.  Maybe this was because the list of keywords was limited, and capitalization was a proxy for keyword?  After #1447, we have a much more complete list of keywords.

Fixes #620.

## Checklist

### DIFF
--- a/mycli/sqlcompleter.py
+++ b/mycli/sqlcompleter.py
@@ -782,7 +782,7 @@ class SQLCompleter(Completer):
         self.reserved_words = set()
         for x in self.keywords:
             self.reserved_words.update(x.split())
-        self.name_pattern = re.compile(r"^[_a-z][_a-z0-9\$]*$")
+        self.name_pattern = re.compile(r"^[_a-zA-Z][_a-zA-Z0-9\$]*$")
 
         self.special_commands: list[str] = []
         self.table_formats = supported_formats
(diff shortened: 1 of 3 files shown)


mycli/sqlcompleter.py
scottnemes, line 1: One additional thing I was going to add was escaping database names:

```
    def extend_database_names(self, databases: list[str]) -> None:
        self.databases.extend([self.escape_name(db) for db in databases])
```

rolandwalker: Sounds good.  Can we do that in a separate PR as its own feature?

scottnemes: Works for me; I'll put together another PR today probably.

Issue #109, opened 2015-08-03
Autocompletion of schema names after FROM keyword


amjith:
It should be possible to list other databases as schema names and provide column completions for them. 

Caveats:
- This requires fetching all the databases, tables and their columns. Which could get expensive if the user has thousands of databases.

Possible Solutions:
- Fetch the current database information right away. Fetch the rest of the databases lazily after launch. This could also put undue strain on the database. 
- Fetch the database metadata on demand. For eg: `SELECT * FROM dbname.` will parse the `dbname` and fetch the metadata in a background thread. Cache this data for next use. This might feel laggy to the user but the users are programmers, so I'm guessing they'll understand the concept of latency and such.

adamchainz:
Another possible solution: local file cache of 'probably exists' stuff that gets checked in lazily in the background. Schemas rarely change, and if they do no one is going to think worse of mycli for it temporarily autosuggesting a database/table/column that was removed recently.

amjith:
Yeah this was suggested by someone on the gitter channel. I'm not sure when to invalidate this file cache. 

I think adding a background thread for fetching the completions is the first step.

adamchainz:
It doesn't need to be invalidated - just refreshed. The background thread re-downloads database and table names and replaces whatever's in there, perhaps one table at a time. If table X was cached and no longer exists - remove it from the cache and list.

mdsrosa:
@amjith Yes, that's what I suggested on the gitter channel and I agree with @adamchainz. It doesn't need to be invalidated - just updated. Now, when should we do the cache update? It should be the first thing to do when the program starts, right? Launch the 'update thread'. What do you think?

amjith:
Yes. You're right it I could kick off a background thread that starts populating the file and use the file as a cache that is partially refreshed every time the user switches the databases or relaunches mycli. 

I'm still concerned about the performance impact. Although for that we can add an option to disable the eager fetching.

mdsrosa:
@amjith Exactly! That's what I was thinking. And maybe we could create a command to just update the cache file. Something like `mycli --update-db-cache-file`, if the user disable the eager fetching.

tsroten:
@amjith I think I'm not quite understanding this all the way. Is this the same thing as https://github.com/dbcli/mycli/issues/111?

amjith:
Yes, this is loosely coupled with #111. I was told that it is common practice to have multiple databases in MySQL and run JOIN queries between the tables in those different databases. 

[...]

PR #1459, opened 2026-01-21
Allow table names to be completed with leading schema names

### LINKED ISSUE #1142: Support multiple databases?
```mysql
mycli> USE db1
mycli> SELECT * FROM db1.tb1 t1  JOIN db2.tb2 t2 ON t1.<cursor_1>= t2.<cursor_2>
```

cursor_1 can auto complete, but cursor_2 nothing shows.

```mysql
mycli> SELECT * FROM <cursor_3>
```

cursor_3 doesn't show anything about my databases, mycli just show the table name from my connection( mycli mysql://user:pwd@127.0.0.1:3306/db_dev)

rolandwalker:
## Description

Allow table names to be completed with a leading schema name as in

    schema.table

Note however that the table names can only be successfully completed if they are in the currently-selected schema.  Completing table names in another schema is still a todo.

This is a one-line fix that appears to be due to a confusion between the strings "database" and "schema" in the completion code.

Strictly speaking this fixes #1142, but not #109.

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -7,6 +7,7 @@ Features
 * Default to standards-compliant `utf8mb4` character set.
 * Stream input from STDIN to consume less memory, adding `--noninteractive` and `--format=` CLI arguments.
 * Remove suggested quoting on completions for identifiers with uppercase.
+* Allow table names to be completed with leading schema names.
 
 
 Bug Fixes
--- a/mycli/packages/completion_engine.py
+++ b/mycli/packages/completion_engine.py
@@ -301,7 +301,7 @@ def suggest_based_on_last_token(
 
         if not schema:
             # Suggest schemas
-            suggest.insert(0, {"type": "schema"})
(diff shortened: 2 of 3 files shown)


scottnemes:
Side effect of this appears to be that it now suggests databases along with tables, which isn't super useful since you can't currently get suggestions for tables in other databases as you mentioned. Also imagine we'd want tables to be listed first anyway, as I think you are more likely to want a table vs another database:

image

So might be worth putting this off until after completions are available for other schemas, otherwise I think it makes it less user friendly at first glance anyway.

rolandwalker:
> Also imagine we'd want tables to be listed first anyway, as I think you are more likely to want a table vs another database:

Good point!  Fixed.

> So might be worth putting this off until after completions are available for other schemas

"The perfect is the enemy of the good".  This is essentially a typo fix that inches us forward.  Let's merge and take the inspiration to work on the next step.

PR #1460, opened 2026-01-21
Add `--throttle` option for batch mode pauses

rolandwalker:
## Description

The `--throttle` option adds a pause between queries in batch mode, which can be useful for long or intensive scripts.

Technically we pause between each line of input, which is usually equivalent to one query.

~To reduce conflicts, this branch is based off of #1450 , and will need to be rebased once that is merged.  There's no need to review until that is done.~ Rebased.

## Checklist

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -19,7 +19,7 @@ from datetime import datetime
 from importlib import resources
 import itertools
 from random import choice
-from time import time
+from time import sleep, time
 from urllib.parse import parse_qs, unquote, urlparse
 
 from cli_helpers.tabular_output import TabularOutputFormatter, preprocessors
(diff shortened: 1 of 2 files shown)


mycli/main.py
```diff
@@ -1920,6 +1922,8 @@ def cli(
             try:
                 if warn_confirmed:
                     mycli.run_query(stdin_text, new_line=True)
+                    if throttle:
```
scottnemes: Currently this pauses after the query runs, so there is also a pause after the last query which isn't ideal. You can move it up a line and add a counter check so it doesn't pause before the first query:

```
if throttle and counter > 1:
   sleep(throttle)
mycli.run_query(stdin_text, new_line=True)
```

rolandwalker:
The intention was not to have it apply to `--execute`  Wouldn't that be a single line anyway?

scottnemes:
> The intention was not to have it apply to `--execute` Wouldn't that be a single line anyway?

@rolandwalker Could be multiple queries from -e or from the prompt even. However I wasn't suggesting you should do that, just clarifying your goals for the feature.

rolandwalker:
Hm.  We probably should support multiple `--execute` and run them in series.  It seems that we don't.

PR #1461, opened 2026-01-21
Add `--unbuffered` mode

### LINKED ISSUE #1184: How to Emulate --quick and --unbuffered options from MySQL Client

rolandwalker:
## Description

Per https://pymysql.readthedocs.io/en/latest/modules/cursors.html#pymysql.cursors.SSDictCursor

    Unbuffered Cursor, mainly useful for queries that return a lot of
    data, or for connections to remote servers over a slow network.

    Instead of copying every row of data into a buffer, this will fetch
    rows as needed. The upside of this is the client uses much less
    memory, and rows are returned much faster when traveling over a slow
    network or if the result set is very big.

Fixes #1184.

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,11 @@
 TBD
 ==============
 
+Features
+--------
+* Add `--unbuffered` mode which fetches rows as needed, to save memory.
+
+
 Bug Fixes
 --------
 * Fix CamelCase fuzzy matching.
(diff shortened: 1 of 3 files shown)


scottnemes (approved):
Hard to verify it's actually doing anything different, but looks good according to the docs anyway!

Issue #1464, opened 2026-01-22
Deprecate built-in SSH features


rolandwalker:
According to #714, the built-in SSH features were "experimental" in 2019 and therefore not fully documented.

Even after some recent work (#1445, #1444) I am unable to make the feature work locally.

There are some long-term open issues and PRs (#867, #806, #869 ) suggesting disrepair.

It seems probable that the feature is unused at this time, and we may begin deprecating the feature, with a removal timeline of say six months.

This issue is for tracking and discussion, including votes to keep the feature!

rolandwalker:
Update: the deprecation message started in https://github.com/dbcli/mycli/releases/tag/v1.46.0 on 22 Jan 2016.  The message links back to this issue.

rolandwalker:
Update: Two weeks after release of the deprecation message there have been no comments, suggesting that the SSH feature is not in use.

I am provisionally closing the other open issues and PRs regarding the SSH feature.  They can always be re-opened.  Listed here in case we need to find them:

Issues: #714, #867, #806 .

Pull Requests: #869 .

After a full month we will move from a "soft deprecation" to a decision.

rolandwalker:
The "soft" deprecation of the built-in SSH functionality was upgraded to a full deprecation in release https://github.com/dbcli/mycli/releases/tag/v1.56.0 on 2026-02-23.

ialexs:
so, if the feature removed, we can not do a tunneling via a 'jump server'? 
mycli --ssh-config-host $ssh_config_host -h $db_host -u $db_user -p $db_pwd

rolandwalker:
Hello!  Yes, we are proposing that the `--ssh-*` options would all go away.  Or, we reimplement them using the `ssh` executable, rather than the `paramiko` library.

My theory has been that the SSH features are not in use.  Are you using them?

ialexs:
hi Roland.. actually yes, using it actively  :)  mycli is my daily driver, to works with many db, across  many jump-server. 
Looking fwd for the new implementation of SSH tunneling, if you decide to ditch paramiko lib \o/ Thanks..

Poshi:
[...]

PR #1465, opened 2026-01-22
Soft-deprecate the built-in SSH features

rolandwalker:
## Description

Soft-deprecate the built-in SSH features, with a reference to an open issue for discussion and voting.

The theory behind this is that the SSH features are not being used.  We will find out!  See #1464 for more detail.

I didn't use the built-in `click` deprecation feature here as it wasn't flexible enough.

## Checklist


mycli/main.py
```diff
@@ -1652,6 +1654,16 @@ def cli(
             fg="yellow",
         )
 
+    # ssh_port and ssh_config_path have truthy defaults and are not included
+    if any([ssh_user, ssh_host, ssh_password, ssh_key_filename, list_ssh_config, ssh_config_host]):
+        if not ssh_warning_off:
```
scottnemes: Personal preference nitpick would be adding the `not ssh_warning_off` to the outer if to reduce nesting, but will approve as-is it works fine.

rolandwalker: Nitpick gladly handled.

PR #1466, opened 2026-01-22
Support a `--checkpoint` file for batch input

rolandwalker:
## Description

Support a `--checkpoint` file for batch input which saves only successful queries, not results.

The motivation is that a checkpoint file allows an operator to resume an interrupted script.  The `--logfile` option is inconvenient for this since it logs results.

## Checklist

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -1322,7 +1322,12 @@ class MyCli:
         string = string.replace("\\_", " ")
         return string
 
-    def run_query(self, query: str, new_line: bool = True) -> None:
+    def run_query(
+        self,
+        query: str,
+        checkpoint: TextIOWrapper | None = None,
+        new_line: bool = True,
+    ) -> None:
         """Runs *query*."""
         assert self.sqlexecute is not None
         self.log_query(query)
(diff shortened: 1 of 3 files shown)


mycli/main.py
```diff
@@ -1322,11 +1322,18 @@ def get_prompt(self, string: str) -> str:
...
+        self,
+        query: str,
+        checkpoint: TextIOWrapper | None = None,
+        new_line: bool = True,
+    ) -> None:
         """Runs *query*."""
         assert self.sqlexecute is not None
         self.log_query(query)
         results = self.sqlexecute.run(query)
+        if checkpoint:
```
scottnemes: This logs the query before it finishes running; is that the intended use case? I would expect it to log after completion under normal circumstances. If so could be moved down below the for loop and then it will write after. Also would recommend using write instead of print, and you can flush it so the log gets updated right away for tracking purposes:
```
        if checkpoint:
            checkpoint.write(query)
            checkpoint.flush()
```

rolandwalker: Great catch/es!

Updated, with a test to cover query failure.

PR #1470, opened 2026-01-23
Don't offer completions for numeric text

rolandwalker:
## Description

Identifiers which begin with numbers can still be completed by starting the text with a backquote.  That might be a little fragile but works for now. Example:

last image

Example of the undesirable old behavior which is fixed in this PR:

Screenshot 2026-01-23 at 10 48 53 AM

## Checklist


rolandwalker:
Valid points!  But I'm not clever enough about the completions to do that yet.

PR #1472, opened 2026-01-24
Configurable numeric alignment in tabular output

### LINKED ISSUE #1102: Numeric Alignment Incorrect In Table Output
Hi, I just started using mycli and it is super slick.  One thing that caught my eye early on is that numbers are not right aligned, compare what I see from mycli 1.22.2:

```
(none)> select 'left' `character`, 1 `integer`, 1.0 `double`;
+-----------+---------+--------+
| character | integer | double |
+-----------+---------+--------+
| left      | 1       | 1.0    |
+-----------+---------+--------+
```

With what the default mysql client shows (notice the numbers are right aligned):

```
mysql> select 'left' `character`, 1 `integer`, 1.0 `double`;
+-----------+---------+--------+
| character | integer | double |
+-----------+---------+--------+
| left      |       1 |    1.0 |
+-----------+---------+--------+
```

It's particularly jarring when there are multiple rows with different magnitudes, e.g.:

```
+-----------------+
| wind_speed_mean |
+-----------------+
| 5               |
| 7               |
| 9               |
| 4               |
| 4               |
| 9               |
| 9               |
| 14              |
| 12              |
| 11              |
| 15              |
| 10              |
| 8               |
| 5               |
| 8               |
| 7               |
| 6               |
| 7               |
| 7               |
| 7               |
+-----------------+
```

Based on screenshots I'm seeing elsewhere this is either a regression or something specific to my situation (although I'm using all defaults settings).  Related issue: https://github.com/dbcli/mycli/issues/635.

rolandwalker:
## Description

After https://github.com/dbcli/cli_helpers/pull/97 we can set the alignment on a per-column basis using the `colalign` parameter.

In this PR we upgrade `cli_helpers` and set `colalign` for columns with a numeric type.  This is more performant than using tabulate's numparse capability, and unlike numparse, works on nullable numeric columns.

The default alignment for numeric columns is changed from left to right, on the basis that this is the default for the vendor client. However, the user is free to change it back in `~/.myclirc` .

Fixes #1102 .

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,11 @@
+TBD
+==============
+
+Features
+--------
+* Right-align numeric columns, and make the behavior configurable.
+
+
 1.47.0 (2026/01/24)
 ==============
 
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from collections import defaultdict, namedtuple
+from decimal import Decimal
 from io import TextIOWrapper
 import logging
 import os
@@ -160,6 +161,7 @@ class MyCli:
         self.login_path_as_host = c["main"].as_bool("login_path_as_host")
         self.post_redirect_command = c['main'].get('post_redirect_command')
         self.null_string = c['main'].get('null_string')
+        self.numeric_alignment = c['main'].get('numeric_alignment', 'right')
 
         # set ssl_mode if a valid option is provided in a config file, otherwise None
         ssl_mode = c["main"].get("ssl_mode", None)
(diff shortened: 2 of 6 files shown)


scottnemes:
Nice! It can actually suport `center` as well if you want to include that in the config file. Not sure on the use case, but it works anyway! Also will accept `decimal` but with the way decimals work I don't think it will actually ever do anything.

rolandwalker:
In the PyMySQL source it appears to be able to return Decimal types: https://github.com/PyMySQL/PyMySQL/blob/d7bb777e503d82bf2496113f07dd4ab249615efc/pymysql/converters.py#L353-L354 .

We can do `center` later if anyone wants it.  It probably wouldn't make sense for just numeric data to be centered, and we'd need another configuration option.

PR #1474, opened 2026-01-24
Add completion candidates for stored procedures

### LINKED ISSUE #278: sproc support
I noticed there's no auto-complete for stored procedures. Would it be possible to get an autocomplete dropdown of available procedures after `CALL`?

rolandwalker:
## Description

Complete stored procedure names after `CALL`.

Fixes #278.

Example:

Screenshot 2026-01-24 at 11 18 40 AM

It's not clear to me how to add a test.

## Checklist

### DIFF
--- a/mycli/sqlcompleter.py
+++ b/mycli/sqlcompleter.py
@@ -924,6 +924,14 @@ class SQLCompleter(Completer):
             metadata[self.dbname][func[0]] = None
             self.all_completions.add(func[0])
 
+    def extend_procedures(self, procedure_data: Generator[tuple[str, str]]) -> None:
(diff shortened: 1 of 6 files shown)


mycli/sqlcompleter.py
```diff
@@ -1093,6 +1107,12 @@ def get_completions(
                     )
                     completions.extend(predefined_funcs)
 
+            elif suggestion["type"] == "procedure":
+                # procs = self.populate_schema_objects(suggestion["schema"], "procedures")
+                procs = self.populate_schema_objects("mycli_test_db", "procedures")
```
scottnemes: Need to remove the test code and enable the real one. Works well after that though!

PR #1476, opened 2026-01-24
Accept multiple `--execute` options and run each specified query in turn

rolandwalker:
## Description

Accept multiple `--execute` options and run each specified query in turn

The current behavior is to accept multiple `--execute` options, but to only execute the last one: a bug.

Since there are multiple values, we can also support `--throttle`, per the discussion in https://github.com/dbcli/mycli/pull/1460.

One inconsistency between STDIN and `--execute` is that `--execute` does not support warning on destructive queries.  This should probably be resolved in the direction of removing the warning from scripts on STDIN.

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,11 @@
+TBD
+==============
+
+Features
+--------
+* Accept multiple `--execute` arguments, running each query in turn.
+
+
 1.47.0 (2026/01/24)
 ==============
 
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -1534,7 +1534,7 @@ class MyCli:
 @click.option("--warn/--no-warn", default=None, help="Warn before running a destructive query.")
 @click.option("--local-infile", type=bool, help="Enable/disable LOAD DATA LOCAL INFILE.")
 @click.option("-g", "--login-path", type=str, help="Read this path from the login file.")
-@click.option("-e", "--execute", type=str, help="Execute command and quit.")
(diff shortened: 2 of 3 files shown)


scottnemes:
What is the benefit of multiple executes over running multiple queries within a single execute like you would normally?

I.e.
```
mycli -e "select 1; select 2; select 3"
```

vs
```
mycli -e "select 1" -e "select 2" -e "select 3"
```

Unless there is a specific use case that seems unnecessary. My original comment was around the fact that you could technically have the throttle affect the multiple queries passed at once, since we parse the queries one by one anyway.

rolandwalker:
The benefits are:
* unify the two batch modes with the same behavior
* fix an untenable bug: only the last `--execute` was (silently) respected

scottnemes:
> The benefits are:
> 
> * unify the two batch modes with the same behavior
> * fix an untenable bug: only the last `--execute` was (silently) respected

Unclear why you need multiple -e params to unify the batch modes? RE: the multiple -e bug, I'd lean towards checking for that scenario and erroring vs adding it in as functionality to avoid the bug. The behavior of accepting multiple is the same for all "basic" options, so we'd be selectively changing that in this case.

```
mycli -u scottn -u noob
(1045, "Access denied for user 'noob'@'localhost' (using password: NO)")
```

rolandwalker:
We never reached agreement about this design.  The bug remains that the user can set multiple `--execute` values.

PR #1477, opened 2026-01-24
Let favorite queries be special commands

### LINKED ISSUE #747: Can't source a file with a favorite command
Hi everyone,

I'm not sure if this is a bug or if I'm missing something so I would appreciate some help:

I have a script file that I need to source regularly, using this command it works fine:

    source /home/statox/projects/boilerplate/setup.sql

I tried to make it a favorite command with this command:

    \fs setupkey source /home/statox/projects/boilerplate/setup.sql

And also by adding this line directly in my `.myclirc`:

    [favorite_queries]
    ...
    setupkey = 'source /home/statox/projects/boilerplate/setup.sql'

In both cases everything is fine (no startup error or anything) until I try to execute the command:

    mysql user@localhost:newws> \f setupkey
    (1064, "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'source /home/statox/projects/boilerplate/setup.sql' at line 1")

I am sure that my sql script works fine since I can execute it directly and my other favorite commands (which don't use `source`) all work fine so I don't know how to fix my issue and I didn't found a related issue yet.

Also `mycli -v` returns `Version: 1.8.1`

Anyway thank you for your great tool :+1:

rolandwalker:
## Description

Fixes #747 .

## Checklist

### DIFF
--- a/mycli/packages/special/iocommands.py
+++ b/mycli/packages/special/iocommands.py
@@ -19,7 +19,9 @@ from mycli.compat import WIN
 from mycli.packages.prompt_utils import confirm_destructive_query
 from mycli.packages.special.delimitercommand import DelimiterCommand
 from mycli.packages.special.favoritequeries import FavoriteQueries
+from mycli.packages.special.main import COMMANDS as SPECIAL_COMMANDS
 from mycli.packages.special.main import ArgType, special_command
+from mycli.packages.special.main import execute as special_execute
 from mycli.packages.special.utils import handle_cd_command
 from mycli.packages.sqlresult import SQLResult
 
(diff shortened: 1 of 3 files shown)


mycli/packages/special/iocommands.py
```diff
@@ -281,12 +283,22 @@ def execute_favorite_query(cur: Cursor, arg: str, **_) -> Generator[SQLResult, N
...
-                    yield SQLResult(title=title, results=cur, headers=headers)
+                is_special = False
+                for special in SPECIAL_COMMANDS:
+                    if sql.lower().startswith(special.lower()):
+                        is_special = True
+                        break
+                if is_special:
+                    for result in special_execute(cur, sql):
+                        result.title = title
+                        yield result
```
scottnemes, line 295: Works great! Only suggestion would be adding a comment here to highlight that the special_execute function returns a SQLResult, so that is why it's just yielding the "result" var instead of a SQLResult object like the rest of the yields. Will approve as-is so feel free to add it or not

rolandwalker: Done!

PR #1488, opened 2026-01-27
"Eager" completions for the `source` command, limiting to `*.sql` files

rolandwalker:
## Description
 * Complete immediately on files in the current directory, even if the text to complete does not start with `.`.
 * This resolves a minor bug in which files in the cwd beginning with `.` were offered as suggestions before the user had typed `./`.
 * Complete only files that end in `.sql` (and directories).
 * Append `/` to suggested directory names.
 * Otherwise continue to complete on paths beginning with `./`, `~`, and `/` in the same way, and continue to offer the punctuation suggestions.

If the restriction to completion on files that end in `.sql` is too restrictive, we could make it configurable, as noted in a comment.

The user can still source a file that does not end in `.sql`.  It just won't be offered as a completion.

Example (first do `touch script.sql`):

last image

Old (shows history suggestion, but not popup completions):

Screenshot 2026-01-27 at 5 38 07 AM

## Checklist

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,11 @@
 TBD
 ==============
 
+Features
+--------
+* "Eager" completions for the `source` command, limited to `*.sql` files.
+
+
 Internal
 --------
 * Remove `align_decimals` preprocessor, which had no effect.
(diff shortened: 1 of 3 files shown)


scottnemes (approved):
I think the suggesting only sql files is fine, maybe add config later if anyone really wants it.

Only thing I noticed is when you do `source` without any text it appears to do a sorted combined list of directories and files; would be nice if files showed up first. But if that is beyond the scope you want here it could be a future pass.

rolandwalker:
> would be nice if files showed up first

Great point.  I just tried and it didn't work right.  Maybe dotfiles also shouldn't appear.  Let's leave those to a future pass.

Issue #1490, opened 2026-01-27
Deprecate reading configuration from `.my.cnf`


rolandwalker:
As mentioned in #581, it has been a longstanding desire to migrate away from reading `.my.cnf` in favor of taking configuration only from `~/.myclirc`.

The reasons in short are: either
 * a given config option exists only in one file or the other, which creates a burden on the user to know which one, or
 * a given config option exists in both, which creates a burden on the user to understand which overrides.

In addition, supporting two configuration files adds code complexity.

We can begin with warnings, which will take some work to implement, but are possible, then run a lengthy deprecation cycle as in #1464 .

It may be that we have to provide a migration tool, either built in to mycli, or as a standalone tool.

rolandwalker:
> It may be that we have to provide a migration tool, either built in to mycli, or as a standalone tool.

In #1493 we take the approach of a verbose warning with instructions to edit `~/.myclirc`, rather than a migration tool.

rolandwalker:
Update: the deprecation notice started in https://github.com/dbcli/mycli/releases/tag/v1.50.0 on 2026-02-07, with a proposed six-month cycle.

parona-source:
Just a heads up. Noticed that documentation on the website doesn't yet reflect the status of the deprecation. 

https://www.mycli.net/config
https://github.com/dbcli/mycli.net/blob/main/content/pages/config.md

rolandwalker:
Good point @parona-source thank you! The web documentation is long out of date.

rolandwalker:
Updated the webpage on this topic but #432 is still very much open.

PR #1493, opened 2026-01-30
Deprecate reading configuration from my.cnf files

rolandwalker:
## Description

 * Create corresponding `~/.myclirc` configuration options for every option which is currently exclusive to my.cnf, using a new `[connection]` section, and prepending every option with `default_`.
 * Move `default_character_set` to the new `[connection]` setting for consistency, but continue to silently read it if present in the `[main]` section.  `default_character_set` also does not activate any warnings, since it already existed.
 * Create a new config property which does not default to the packaged myclirc.
 * Emit a verbose warning if the user has any _controlling_ configuration option in a my.cnf file.
 * Let corresponding CLI arguments always take precedence over configuration.
 * To simplify logic, always create a `[connection]` section in the internal data structure representing `~/.myclirc`, and likewise for [client] and [mysqld] in the my.cnf data structure.
 * Finesse some empty controlling configuration: _eg_ an empty setting for `default_character_set` should default to `utf8mb4`.
 * For consistency, also handle `default_ssl_ca_path` in the `[connection]` section, though it has no my.cnf equivalent.
 * Add a configuration option `my_cnf_transition_done` to allow the user to just ignore all of this.

The very verbose warnings contain instructions on how to create controlling `~/.myclirc` configuration options, the presence of which will suppress the warnings.

It is important to note that the warnings only affect users with files at `/etc/my.cnf`, `/etc/mysql/my.cnf`,  `/usr/local/etc/my.cnf`, or `~/.my.cnf`, with certain options explicitly set, such as `ssl-ca` in the `[client]` section.  If the user does not have a my.cnf file, no warnings will be emitted.

Note also that since the default myclirc content has been updated to contain controlling configuration options, the warnings will never be emitted for fresh installs, only upgrades (which do not overwrite `~/.myclirc`).

~We should consider a CLI option or configuration setting to unconditionally suppress all such warnings, but the downside of that is continuing to accept the CLI option for compatibility after the deprecation cycle is over.  A configuration option could be ignored.~ EDIT: added a configuration option `my_cnf_transition_done`.

Addresses #1490 , which is referenced in the warning message, and should be kept open during the deprecation cycle.

I'm not sure how to write a test for the warning.

## Checklist

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -20,6 +20,7 @@ from datetime import datetime
 from importlib import resources
 import itertools
 from random import choice
+from textwrap import dedent
 from time import sleep, time
 from urllib.parse import parse_qs, unquote, urlparse
 
@@ -137,6 +138,11 @@ class MyCli:
         # Load config.
         config_files: list[str | IO[str]] = self.system_config_files + [myclirc] + [self.pwd_config_file]
         c = self.config = read_config_files(config_files)
+        # this parallel config exists only to compare with my.cnf and can be removed with my.cnf support
+        self.config_without_package_defaults = read_config_files(config_files, ignore_package_defaults=True)
+        for toplevel in ['main', 'connection']:
+            if not self.config_without_package_defaults.get(toplevel):
+                self.config_without_package_defaults[toplevel] = {}
         self.multi_line = c["main"].as_bool("multi_line")
         self.key_bindings = c["main"]["key_bindings"]
         special.set_timing_enabled(c["main"].as_bool("timing"))
@@ -219,6 +225,10 @@ class MyCli:
                 print("Error: Unable to read login path file.")
 
         self.my_cnf = read_config_files(self.cnf_files, list_values=False)
+        if not self.my_cnf.get('client'):
+            self.my_cnf['client'] = {}
+        if not self.my_cnf.get('mysqld'):
+            self.my_cnf['mysqld'] = {}
         prompt_cnf = self.read_my_cnf(self.my_cnf, ["prompt"])["prompt"]
         self.prompt_format = prompt or prompt_cnf or c["main"]["prompt"] or self.default_prompt
         self.multiline_continuation_char = c["main"]["prompt_continuation"]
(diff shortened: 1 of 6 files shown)


mycli/main.py
```diff
@@ -513,21 +523,65 @@ def connect(
...
+        if not charset:
+            charset = 'utf8mb4'
 
         # Favor whichever local_infile option is set.
         use_local_infile = False
-        for local_infile_option in (local_infile, cnf["local-infile"], cnf["loose-local-infile"], False):
+        for local_infile_option in (
+            local_infile,
+            self.config_without_package_defaults['connection'].get('default_local_infile'),
+            cnf['local-infile'],
```
scottnemes, line 559: The .my.cnf/etc options can be any combo of underscores and hyphens, so you'll want to at minimum check both the usual cases (local-infile, local_infile). Something like `default_character_set` can actually even be `default_character-set` and MySQL will still read it, but likely okay to not deal with that extreme case.

rolandwalker: Important knowledge!

rolandwalker: Addressed.

mycli/main.py
```diff
@@ -1863,6 +1921,67 @@ def get_password_from_file(password_file: str | None) -> str | None:
...
 
+    # todo: removeme after a period of transition
+    for tup in [
+        ('client', 'prompt', 'prompt', 'main', 'prompt'),
+        ('client', 'pager', 'pager', 'main', 'pager'),
+        ('client', 'skip-pager', 'skip-pager', 'main', 'enable_pager'),
+        # this is a white lie, because default_character_set can actually be read from the package config
+        ('client', 'default-character-set', 'default-character-set', 'connection', 'default_character_set'),
+        # local-infile can be read from both sections
+        ('mysqld', 'local-infile', 'local-infile', 'connection', 'default_local_infile'),
```
scottnemes, line 1966: These should be `mysql` (for the client) instead of `mysqld` (for the server)

mycli/main.py
```diff
@@ -1863,6 +1921,67 @@ def get_password_from_file(password_file: str | None) -> str | None:
...
+            myclirc_section_name,
+            myclirc_item_name,
+        ) = tup
+        if (
+            mycli.my_cnf[mycnf_section_name].get(mycnf_item_name) is None
+            and mycli.my_cnf[mycnf_section_name].get(mycnf_item_name.replace('-', '_')) is None
+        ):
+            continue
+        if mycli.config_without_package_defaults[myclirc_section_name].get(myclirc_item_name) is None:
+            cnf_value = mycli.my_cnf[mycnf_section_name].get(mycnf_item_name)
```
scottnemes, line 1994: This is always checking for the hyphenated version, i.e. `local-infile`, so if you have `local_infile` or  `ssl_verify_server_cert` or any of those in the .my.cnf it will show None in the warning instead of the value.

mycli/main.py
```diff
@@ -1863,6 +1921,67 @@ def get_password_from_file(password_file: str | None) -> str | None:
...
+            click.secho(
+                dedent(
+                    f"""
+                    Reading configuration from my.cnf files is deprecated.
+                    See https://github.com/dbcli/mycli/issues/1490 .
+                    The cause of this message is the following in a my.cnf file without a corresponding
+                    ~/.myclirc entry:
+
+                    [{mycnf_section_name}]
```
scottnemes: This whole message gets repeated for each value found; not the end of the world, but might be nice if you can put together all the offending sections/values and then display one message with everything in it. However if that is too complicated or more confusing then works as-is.

mycli/main.py
rolandwalker: We are only reading the `client` and `mysqld` sections:
* https://github.com/dbcli/mycli/blob/5f2253e840b1f51deb5336e20a3fccbc91dbc426/mycli/main.py#L417
so those are the only sections we need to transition.

mycli/main.py
rolandwalker: Right! Fixed.

mycli/main.py
scottnemes: Link doesn't go anywhere. However `[mysqld]` is server options; `[mysql]` is client (specifically the mysql client) options. So I don't think you'd ever want to read the server settings, just the client settings.

rolandwalker: Fixed the link!

The transition code is bound by what the current mycli behavior is.  There is no reason to start reading the `[mysql]` section now if the purpose is to transition away from my.cnf.  Likewise, even if it isn't good practice to read `[mysqld]`, we do!

scottnemes: Gotcha. Yeah that means that it has never done anything useful, as those options don't belong in the `[mysqld]` section at all (vendor client wouldn't read them there, so if someone was putting them there it would be strange). But I see your point that it doesn't matter now.

scottnemes: Only caveat there is for something like local_infile, which can be a server option, might get picked up with your warning and confuse the user (that's how I noticed it; I had that set for the server). But don't see a way around that as you have no idea if it really is a server option or a misplaced client option.

mycli/main.py
rolandwalker: You are right that it would be nicer, but it seems like a lot of work.

rolandwalker:
Added a `~/.myclirc` configuration option `my_cnf_transition_done` which shortcuts the entire check if True.  In the future we can just ignore it.

mycli/main.py
rolandwalker: The fact that we have been reading and setting a security-adjacent option without you realizing is just one more reason to stop reading my.cnf at all.

mycli/main.py
```diff
@@ -1863,6 +1921,78 @@ def get_password_from_file(password_file: str | None) -> str | None:
...
+        ('client', 'ssl-verify-server-cert', 'ssl-verify-server-cert', 'connection', 'default_ssl_verify_server_cert'),
+    ]:
+        (
+            mycnf_section_name,
+            mycnf_item_name,
+            printable_mycnf_item_name,
+            myclirc_section_name,
+            myclirc_item_name,
+        ) = tup
+        if mycli.config['main'].get('my_cnf_transition_done'):
```
scottnemes: This needs as_bool otherwise it always comes back as True (string of "False" is True, hah)

```
if mycli.config['main'].as_bool('my_cnf_transition_done'):
```

mycli/main.py
scottnemes: This works now, only other thing related I noticed is in the warning output it always uses the hyphenated version. So if you have `local_infile` in the config, the warning shows `local-infile` as what it found.

mycli/main.py
```diff
@@ -1863,6 +1921,78 @@ def get_password_from_file(password_file: str | None) -> str | None:
...
+
+                    Values are documented at https://github.com/dbcli/mycli/blob/main/mycli/myclirc .  An
+                    empty <value> is generally accepted.
+
+                    To ignore all of this, set
+
+                        [main]
+                        my_cnf_transition_done = True
+
+                    in ~/.myclirc.
```
scottnemes, line 2023: Might add a line break at the end, as when multiple warnings show up it all jams together.

rolandwalker: Good idea, done

mycli/main.py
rolandwalker: Done, with str_to_bool().

mycli/main.py
rolandwalker: Yep, I thought that was reasonable.

PR #1495, opened 2026-01-30
Show user in password prompt (fixing regression)

rolandwalker:
## Description
It would seem that there was a regression in #1436, and we need to redo #1291.

In order to do this, we have to lose some special DSN logic.  There is just no way under the current setup, since for instance the SSL params get parsed out of the DSN before calling `connect()`.

Once solution could be to do all configuration parsing early, a long-term goal.

Here we replace the DSN fallthrough with a warning message.

The `password_from_file` logic is incidentally improved.

## Checklist

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -501,22 +501,10 @@ class MyCli:
         }
 
         cnf = self.read_my_cnf(self.my_cnf, list(cnf.keys()))
-        # get user from config so we can use it in password checking logic and prompt
-        user = user or cnf["user"] or os.getenv("USER")
-
-        if database is None and passwd is not None and "://" in passwd:
-            # check if the scheme is valid. We do not actually have any logic for these, but
-            # it will most usefully catch the case where we erroneously catch someone's
-            # password, and give them an easy error message to follow / report
-            is_valid_scheme, scheme = is_valid_connection_scheme(passwd)
-            if not is_valid_scheme:
-                click.secho(f"Error: Unknown connection scheme provided for DSN URI ({scheme}://)", err=True, fg="red")
-                sys.exit(1)
-            database = passwd
-            passwd = click.prompt(f"Enter password for {user}", hide_input=True, show_default=False, default='', type=str, err=True)
 
         # Fall back to config values only if user did not specify a value.
         database = database or cnf["database"]
+        user = user or cnf["user"] or os.getenv("USER")
         host = host or cnf["host"]
         port = port or cnf["port"]
         ssl_config: dict[str, Any] = ssl or {}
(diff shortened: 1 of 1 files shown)


scottnemes:
I removed it on purpose because when prompting the user for a password at the start, the username is not yet known unless the `--user/-u` option is passed. So you'd have to move the config checks up higher, or it just shows None. But if that's what you want, then works as-is then anyway.

```
uv run mycli --password
Enter password for None:

uv run mycli -u scottn --password
Enter password for scottn:
```

https://github.com/dbcli/mycli/blob/a237ee9a45057fa346b0657c403e4fd3ce6cfd9a/mycli/main.py#L507

rolandwalker:
I see.  We don't have `cnf` yet.

mycli/main.py
```diff
@@ -538,9 +539,50 @@ def connect(
...
+                click.secho(f"Path '{password_file}' is a directory, not a file", err=True, fg="red")
+                sys.exit(1)
+            except Exception as e:
+                click.secho(f"Error reading password file '{password_file}': {str(e)}", err=True, fg="red")
+                sys.exit(1)
+
+        if passwd and '://' in passwd:
+            is_valid_scheme, _scheme = is_valid_connection_scheme(passwd or '')
+            if is_valid_scheme:
+                click.secho('Warning: password looks like a DSN. Check the command line.', err=True, fg='yellow')
```
scottnemes: Why are you removing the logic to catch the case where they use the password flag and also provide a DSN? We keep going around in circles on this, but seems you just want it this way and did not previously say so.

scottnemes: Never mind I realized you mentioned this in your original PR description. You're removing it because it changes the way things work when all of the related logic is moved to the new location. I'll see if I can figure anything else out. Though I'd question if having the user in the prompt is more important, but that part is up to you.

scottnemes: I pushed a commit that I believe gives you what you wanted plus keeping the original functionality (was easier than trying to explain it, so feel free to revert if you don't want it).

scottnemes: Made one more small change; realized that when I moved the line up to check the vendor configs for a password it changed the password hierarchy, so I moved it back to where it was now that it worked as expected.

scottnemes:
@rolandwalker Apologies for the back and forth on this. After working for a while to try and re-add the original password related functionality back into your change (while keeping the user in the prompt), I realized it could be done in a much simpler way (as far as I can tell anyway). I created a new PR with what I believe works, and included your updated password file logic. I believe that was the only other thing that you added (other than the import re-organization, I added that as well). But if I missed something let me know:

https://github.com/dbcli/mycli/pull/1502

rolandwalker:
Closing in favor of https://github.com/dbcli/mycli/pull/1502 .

PR #1497, opened 2026-01-31
[feat] Suggest column names from all tables in the current database after SELECT (#212)

### LINKED ISSUE #212: Suggest columns names when typing SELECT query
Thanks for a great tool.  I noticed that when smart completion is enabled, it doesn't suggest column names after a SELECT unless the FROM part of the query is already entered, which means that you have to enter the FROM table first and then go back to the SELECT to type the column names.  I noticed when smart completion is off, the SELECT will suggest column names from all tables.

Would it be possible to have this behavior in smart completion mode, or an option to enable it, where if you type SELECT and haven't yet entered the FROM, it will suggest column names from all tables?  I would find that more efficient so that queries can be written from left to right without having to go back to the beginning to complete it, while still taking advantage of smart completion once the table names in the FROM are entered.

scottnemes:
## Description
After typing SELECT, columns from all tables in the current database will be returned as auto complete suggestions. This allows the user to get completions for columns without having to write the FROM clause first.

image

Resolves #212 

## Checklist


mycli/sqlcompleter.py
```diff
@@ -1087,6 +1087,10 @@ def get_completions(
                     # which should suggest only columns that appear in more than
                     # one table
                     scoped_cols = [col for (col, count) in Counter(scoped_cols).items() if count > 1 and col != "*"]
+                elif not tables:
+                    # if tables was empty, this is a naked SELECT and we are
+                    # showing all columns. So make them unique and sort them.
+                    scoped_cols = sorted(set(scoped_cols), key=lambda s: s.strip('`'))
```
rolandwalker, line 1093: Why remove the backticks at this point?

rolandwalker (approved):
Outstanding!

rolandwalker:
Will keywords still complete here?

Edit: yes, but this needs some further work for `SELECT COUNT(…)'.  `COUNT` is buried.

mycli/sqlcompleter.py
scottnemes: It doesn't actually remove them (see my example screenshot); it's just for sorting purposes, otherwise all the backtick ones come first.

scottnemes:
> Will keywords still complete here?
> 
> Edit: yes, but this needs some further work for `SELECT COUNT(…)'. `COUNT` is buried.

@rolandwalker Not sure what you mean here, it works as I would expect. Columns show up first with no further input, so COUNT and other functions are farther down if that's what you mean. But that implies having an arbitrary weight (which we could do) to a function like COUNT. But once you start typing, it works as I would expect.

image

image

image

image

<br><br><br>
And comparing to current functionality, which works the same:

image

image

image

image

rolandwalker:
Right.  Working on finesse for builtins such as `COUNT()` in #1500.

PR #1499, opened 2026-01-31
Store/retrieve passwords with the system keyring

### LINKED ISSUE #1496: Support reading passwords from the system keychain
Using https://pypi.org/project/keyring, as mentioned in #581.

The keys would need to be stored per-server.

rolandwalker:
## Description
 * Off by default.
 * Can be enabled by setting `use_keyring = True` in `~/.myclirc`.
 * Can be enabled on a per-connection basis with `--use-keyring=true` at invocation time.
 * Note that the hostname is considered to be different if short or qualified.
 * A password can be reset with `--use-keyring=reset` at invocation time, or (not documented) using the `keyring` CLI tool.

Fixes #1496.

This was super easy to implement!  But realistic tests look quite difficult.

## Checklist

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -28,6 +28,7 @@ from cli_helpers.tabular_output.output_formatter import MISSING_VALUE as DEFAULT
 from cli_helpers.utils import strip_ansi
 import click
 from configobj import ConfigObj
+import keyring
 from prompt_toolkit.auto_suggest import AutoSuggestFromHistory
 from prompt_toolkit.completion import Completion, DynamicCompleter
 from prompt_toolkit.document import Document
@@ -480,6 +481,8 @@ class MyCli:
         ssh_key_filename: str | None = "",
         init_command: str | None = "",
         unbuffered: bool | None = None,
+        use_keyring: bool | None = None,
+        reset_keyring: bool | None = None,
     ) -> None:
         cnf = {
             "database": None,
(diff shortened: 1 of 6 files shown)


mycli/main.py
```diff
@@ -1538,6 +1558,13 @@ def get_last_query(self) -> str | None:
     '--format', 'batch_format', type=click.Choice(['default', 'csv', 'tsv', 'table']), help='Format for batch or --execute output.'
 )
 @click.option('--throttle', type=float, default=0.0, help='Pause in seconds between queries in batch mode.')
+@click.option(
```
scottnemes, line 1560: Would use a choice type here since you have three set options you want to pick from.

mycli/main.py
```diff
@@ -1863,6 +1891,16 @@ def get_password_from_file(password_file: str | None) -> str | None:
     if show_warnings:
         mycli.show_warnings = show_warnings
 
+    if use_keyring_cli_opt and use_keyring_cli_opt.lower() == 'reset':
```
scottnemes: You're pulling in True/False values as strings, so all of the logic needs to account for that. Easiest probably would be to just check the option variable and then set a new real bool var to use for your logic.

mycli/main.py
```diff
@@ -1863,6 +1891,16 @@ def get_password_from_file(password_file: str | None) -> str | None:
...
+        use_keyring = True
+        reset_keyring = True
+    elif use_keyring_cli_opt is None:
+        use_keyring = mycli.config['main'].get('use_keyring')
```
scottnemes: Similar to above, will need as_bool to treat it as such.

mycli/main.py
```diff
@@ -1863,6 +1891,16 @@ def get_password_from_file(password_file: str | None) -> str | None:
...
+        reset_keyring = False
+    else:
+        use_keyring = str_to_bool(use_keyring_cli_opt)
```
scottnemes, line 1900: Just noticed this function, could use it for the above as well.

mycli/main.py
rolandwalker: Good idea!  Though str_to_bool() was going to handle variations such as "on" or "off" -- that's not important.

mycli/main.py
rolandwalker: Isn't that exactly what we are doing, where the "new real bool var" is `use_keyring`?

mycli/main.py
rolandwalker: Right!  Great catch.  Our config library should really do that for us.

mycli/main.py
scottnemes: It is; I was mixing it up with the other suggestions because `use_keyring_cli_opt` is still a string and in this IF it is being treated like a bool. It works because all that really matters is that `use_keyring_cli_opt` is not none so you can .lower it. So I would recommend switching it to be more clear

```
if use_keyring_cli_opt is not None and use_keyring_cli_opt.lower() == 'reset':
```

That (to me at least) makes it clear what you're doing and that you don't actually care that it's a bool (or not really a bool in this case)

rolandwalker: Done!  (This may be against a style PEP, however.  But I agree that it is clear.)

scottnemes: If ruff and mypy are happy with it then I'm good with it since that's the current "standard" here anyway!

PR #1502, opened 2026-02-02
[fix] Show user in password prompt

scottnemes:
## Description
Re-add the user to the password prompt. Was previously removed due to the way the new password prompting worked, and lack of user info early on in the application runtime.

This delays the user requested password prompt `(--password/-p)` until after the my.cnf related config is loaded. This technically delays password prompt time, however almost the entirety of the delay is from startup, so in reality the difference is moot currently.

## Checklist


scottnemes:
> Can you rewrite the subject line and description from scratch, removing the references to the defunct PR?

Updated!

PR #1509, opened 2026-02-04
Add `--checkup` mode to check user configuration

rolandwalker:
## Description
This helps the user who has upgraded see which new features are silently being set to the package defaults.

last image

## Checklist

    ```bash
    ```

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -138,11 +138,10 @@ class MyCli:
         # Load config.
         config_files: list[str | IO[str]] = self.system_config_files + [myclirc] + [self.pwd_config_file]
         c = self.config = read_config_files(config_files)
-        # this parallel config exists only to compare with my.cnf and can be removed with my.cnf support
+        # this parallel config exists to
+        #  * compare with my.cnf
+        #  * support the --checkup feature
         self.config_without_package_defaults = read_config_files(config_files, ignore_package_defaults=True)
-        for toplevel in ['main', 'connection']:
-            if not self.config_without_package_defaults.get(toplevel):
-                self.config_without_package_defaults[toplevel] = {}
         self.multi_line = c["main"].as_bool("multi_line")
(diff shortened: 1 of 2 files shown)


mycli/main.py
```diff
@@ -2217,5 +2225,28 @@ def read_ssh_config(ssh_config_path: str):
         return ssh_config
 
 
+def do_config_checkup(mycli: MyCli) -> None:
+    did_output = False
+    for section_name in mycli.config.keys():
+        if section_name not in mycli.config_without_package_defaults:
+            if not did_output:
+                print('\nMissing in user ~/.myclirc:\n')
+            print(f'The entire section:\n\n    [{section_name}]\n')
```
scottnemes: Due to:
https://github.com/dbcli/mycli/blob/3fd95b0d2300d40452170661c5d2c9871961c566/mycli/main.py#L145

This won't ever fire for the `[main]` or `[connection]` sections (will just output every option in the section). Likely wouldn't happen for `[main]` when upgrading, but possible for `[connection]` since that is newer. Not a big deal, but wanted to point it out in case it matters.

mycli/main.py
```diff
@@ -2217,5 +2225,28 @@ def read_ssh_config(ssh_config_path: str):
         return ssh_config
 
 
+def do_config_checkup(mycli: MyCli) -> None:
+    did_output = False
+    for section_name in mycli.config.keys():
+        if section_name not in mycli.config_without_package_defaults:
```
scottnemes: If no user .myclirc exists, it will output every option as missing. Similar vein to the below comment that this may not matter for an upgrade situation, but might be worth adding a check to do nothing if no file exists.

mycli/main.py
rolandwalker: Great catch, fixed.

mycli/main.py
rolandwalker: Great idea!  Done.

PR #1510, opened 2026-02-04
Don't emit keyring-saved message unless needed

rolandwalker:
## Description
Don't emit keyring-saved message unless needed

Incidentally update message wordings to match CLI option: "keyring" rather than "keychain".

## Checklist

    ```bash
    ```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -9,6 +9,7 @@ Features
 Bug Fixes
 --------
 * Link to `--ssl`/`--no-ssl` GitHub issue in deprecation warning.
+* Don't emit keyring-updated message unless needed.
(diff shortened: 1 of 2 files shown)


scottnemes (approved):
Works as-is.

Question for this or a future PR though:

Should we be waiting until after a successful connection to save the password to the keyring? As-is it will save and then go on to fail the connection if the password is wrong. I don't normally use a keyring so not sure what behavior the user would expect, but seems like it'd be preferable to only save if the password works.

rolandwalker:
Great point for future work: save the password only on a successful connection.

PR #1517, opened 2026-02-07
Move `ssl_mode` to `[connection]` section

rolandwalker:
## Description
 * move to new section
 * change the name to `default_ssl_mode`
 * place with other SSL options
 * continue to silently accept the old spelling in `[main]`

## Checklist

    ```bash
    ```

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -180,7 +180,7 @@ class MyCli:
             self.llm_prompt_section_truncate = 0
 
         # set ssl_mode if a valid option is provided in a config file, otherwise None
-        ssl_mode = c["main"].get("ssl_mode", None)
+        ssl_mode = c["main"].get("ssl_mode", None) or c["connection"].get("default_ssl_mode", None)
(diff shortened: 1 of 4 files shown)


mycli/main.py
```diff
@@ -171,7 +171,7 @@ def __init__(
         self.binary_display = c['main'].get('binary_display')
 
         # set ssl_mode if a valid option is provided in a config file, otherwise None
-        ssl_mode = c["main"].get("ssl_mode", None)
+        ssl_mode = c["connection"].get("default_ssl_mode", None) or c["main"].get("ssl_mode", None)
```
scottnemes: This results in it basically ignoring the old `[main]` option since it will always either be the new `default_ssl_mode` if it exists or `None`. So will need a little extra logic if you want to still accept the old value (either a try/except without using `.get` or split it up so you get the new var, check if it's `None`, and then check the old var if so).

rolandwalker: Actually I think just reversing the two does the right thing.

PR #1518, opened 2026-02-07
Add unsupported and deprecated `--checkup` sections

rolandwalker:
## Description
Let `--checkup` show config options which are read and ignored, as well as config options which are read, but deprecated.

## Checklist

    ```bash
    ```


scottnemes (approved):
These config related changes are getting a bit hard to follow, so will have to trust it's doing what you want it to anyway.

I.e. if I have a value like `local_infile` in `.my.cnf`, and then comment out the `[connection]` section of the `~/.myclirc`, it only complains about the `local_infile` one and not the missing `[connection]` section.

rolandwalker:
Let's not worry about any interaction with `my.cnf` which is deprecated.

PR #1519, opened 2026-02-07
Give destructive warning on multi-table `UPDATE`s

### LINKED ISSUE #791: Mark dangerous multiple-table updates as destructive queries
MySQL allows to update multiple tables at once, like this:

    update foo, bar
    set foo.a = 1, bar.b = 2
    where bar.b = 1

The table for which there is no `where`-clause is specified will be updated completely. We now warn about complete table updates (issue #728, resolved in #785), but the current realization only works with the one-table form.

rolandwalker:
## Description
Only suppress the destructive-command warning for the simple case of a single-table `UPDATE` with a `WHERE` clause.

Fixes #791.

## Checklist

    ```bash
    ```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -9,6 +9,7 @@ Features
 Bug Fixes
 --------
 * Correct mangled schema info sent in LLM prompts.
+* Give destructive warning on multi-table `UPDATE`s.
 
 
 1.50.0 (2026/02/07)
--- a/mycli/packages/parseutils.py
+++ b/mycli/packages/parseutils.py
@@ -275,6 +275,25 @@ def query_has_where_clause(query: str) -> bool:
     return any(isinstance(token, sqlparse.sql.Where) for token_list in sqlparse.parse(query) for token in token_list)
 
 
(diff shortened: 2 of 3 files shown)


rolandwalker:
> So long and short, this will start forcing a warn on all multi table updates, even if they have a full where clause. 

That was the intention of the PR!  We'd have to go deep into the SQL to do otherwise.  Open to that, too!

scottnemes:
> > So long and short, this will start forcing a warn on all multi table updates, even if they have a full where clause.
> 
> That was the intention of the PR! We'd have to go deep into the SQL to do otherwise. Open to that, too!

Yeah I'm pointing out this makes divergent behavior, which in the past you've always been against. But if it's working as you want, then that's okay; it works as described.

rolandwalker:
Agreed, @scottnemes, this is a super tricky feature.  Even with the prior and current implementations, nothing helps you if you type `WHERE 1 = 1`.  It probably would have been better all along to treat `UPDATE` the same as other keywords.

PR #1545, opened 2026-02-13
Accept special commands without trailing semicolons even in multi-line mode

rolandwalker:
## Description
"exit" was always an exception to the multi-line trailing semicolon rule.  At minimum, "help" should likewise be an exception.

This makes _all_ special commands exceptions to the multi-line trailing semicolon requirement, on the theory of consistency, because special commands starting with backslash have also always been exceptions.

Tests seem difficult.  We'd want to do a "behave" test using an alternate configuration.

## Checklist

    ```bash
    ```

### DIFF
--- a/mycli/clibuffer.py
+++ b/mycli/clibuffer.py
@@ -3,6 +3,7 @@ from prompt_toolkit.enums import DEFAULT_BUFFER
 from prompt_toolkit.filters import Condition, Filter
 
 from mycli.packages.special import iocommands
+from mycli.packages.special.main import COMMANDS as SPECIAL_COMMANDS
 
 
 def cli_is_multiline(mycli) -> Filter:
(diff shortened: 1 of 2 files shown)


mycli/clibuffer.py
```diff
@@ -21,38 +22,36 @@ def cond():
...
             "\\g",
             "\\G",
             r"\e",
             r"\clip",
         ))
         or
-        # Exit doesn't need semi-column`
-        (text == "exit")
+        # non-backslashed special commands such as "exit" or "help" don't need semicolon
+        first_word in SPECIAL_COMMANDS
```
scottnemes, line 47: Since first_word seems to come in as-is (could be mixed case), I'd either change this first one to .upper() or just combine the two cases and have it be something like `first_word.lower()` in (key.lower() for key in SPECIAL_COMMANDS.keys())` to cover both cases maybe.

rolandwalker: I'm failing to understand the suggestion.  By the time we get here, the words are only the textual forms, because the backslash forms were handled above.  So we should only need to account for accepting _eg_ `EXIT` like `exit`.  If this is wrong we can follow up.

scottnemes: It's not wrong, I just think you could get rid of the first condition. You're currently checking if the as-is value is in special commands, and then you check if the lowercase version is in special commands. So you really only need to check the lowercase value as that will cover both situations (user gives EXIT or exit), or ExiT, or whatever).

PR #1549, opened 2026-02-14
Ability to suppress control-d exit behavior

rolandwalker:
## Description
Arguably the default should be changed since it can be confusing to new users.  But exiting on control-d/EOF is also the default for _eg_ bash.

Ringing the bell when no action can be taken is consistent with other control keys on the empty line.

Tests would be nice!  But would require an alternate myclirc and simulated keypresses, which I don't know how to do.

## Checklist

    ```bash
    ```


scottnemes (approved):
Verified new option controls the ctrl-d behavior as expected.

PR #1550, opened 2026-02-14
Support Truecolor in prompt_toolkit sessions

rolandwalker:
## Description
Support Truecolor in prompt_toolkit sessions if the environment variable `$COLORTERM` contains "truecolor"; otherwise fall back to 8-bit color depth, the prompt_tookit default.

I must admit that between this and https://github.com/dbcli/cli_helpers/pull/102 I cannot yet see a difference.  But there was a request for it, which is reasonable.

It would be great if someone attached some screenshots showing the difference.

## Checklist

    ```bash
    ```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -9,6 +9,7 @@ Features
 * Optionally defer auto-completions until a minimum number of characters is typed.
 * Make the completion interface more responsive using a background thread.
 * Option to suppress control-d exit behavior.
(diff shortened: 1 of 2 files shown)


scottnemes (approved):
Think you'd have to being using gradients or similar to be able to notice this, so it's a pretty niche use case from what I can gather anyway.

PR #1551, opened 2026-02-14
Don't offer completions within strings

rolandwalker:
## Description
 * Don't offer completions when the cursor is within a string.
 * Move the similar implementation for numbers to a better location, next to the string check.

It would be great if `suggest_based_on_last_token()` was refactored to use named parameters.

Example of the behavior solved by this PR:

Screenshot 2026-02-14 at 6 35 42 AM

## Checklist

    ```bash
    ```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,11 @@
+Upcoming (TBD)
+==============
+
+Bug Fixes
+--------
+* Don't offer autocomplete suggestions when the cursor is within a string.
+
+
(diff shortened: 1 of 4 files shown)


rolandwalker:
Codex is right about `_is_inside_quotes()` and the feature being incomplete.  That way might not be too performant.  It's a shame, because somewhere there is the state of the Pygments lexer, which we presumably can't access.

PR #1553, opened 2026-02-14
Add the ability to send app-layer keepalive pings

rolandwalker:
## Description
The config option is named with `default_` under the `[connection]` section on the theory that it might one day be configurable by individual connections.

It may be hard to prove when this works/does something useful.  But I have the issue of frequent reconnections with some servers, and this feature seems to be provided by MySQL Workbench:

* https://stackoverflow.com/questions/31811517/mysql-workbench-drops-connection-when-idle/37890150#37890150

image

## Checklist

    ```bash
    ```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -10,6 +10,7 @@ Features
 * Make the completion interface more responsive using a background thread.
 * Option to suppress control-d exit behavior.
 * Better support Truecolor terminals.
+* Ability to send app-layer keepalive pings to the server.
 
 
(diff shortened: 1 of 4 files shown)


scottnemes (approved):
Works with the basic use case of running queries in the CLI. Can test by setting `wait_timeout ` to something low, waiting that long, then running a query and it will reconnect. But if you enable the keepalive to less than the `wait_timeout`, then you should not see a reconnect after that time.

PR #1572, opened 2026-02-16
Catch `getpwuid()` `KeyError` on an unknown user id

### LINKED ISSUE #1567: Crashing with unknown uid
Running `mycli` with a socket owned by an unknown user causes `getpwuid()` to return an error.

This can happen when the socket is owned by a docker container. That call is only used for a log message; it should be wrapped in a `try/except`.

The offending line is here: https://github.com/dbcli/mycli/blob/d38ff4dfcf5064462755c7e0391a4d44c10e07ac/mycli/main.py#L686

### Steps to Reproduce

Bind a socket from a mysql docker container to the host, and try to run `mycli` on the host.

### System

 * mycli version: 1.44.2
 * OS/version: NixOs unstable.

rolandwalker:
## Description
Fixes #1567, if `KeyError` is the correct exception.

This can happen in the Docker context.

## Checklist

    ```bash
    ```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,7 @@ Upcoming (TBD)
 Bug Fixes
 --------
 * Don't offer autocomplete suggestions when the cursor is within a string.
+* Catch `getpwuid` error on unknown socket owner.
 
 
 1.54.0 (2026/02/16)
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -722,7 +722,10 @@ class MyCli:
 
         try:
             if not WIN and socket:
-                socket_owner = getpwuid(os.stat(socket).st_uid).pw_name
+                try:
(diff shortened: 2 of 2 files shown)


scottnemes (approved):
If you want to test the bots suggestion, you can change the owner of the socket and try. 
```
sudo chown 323:323 /var/run/mysqld/mysqld.sock
```

Error should look something like this anyway:
```
'getpwuid(): uid not found: 323'
```

rolandwalker:
It seems KeyError is right.  The bot suggestions don't seem overly interesting.  This is solely for logging.

scottnemes:
> It seems KeyError is right. The bot suggestions don't seem overly interesting. This is solely for logging.

Yes, point was you could get the uid from the exception to display as the owner if you wanted. But sounds like you in fact do not want, so doesn't matter hah.

Issue #1564, opened 2026-02-14
Difference in completion behavior after a backtick


scottnemes:
### Suggested troubleshooting steps for bug reports

 * [X] Upgraded to the latest mycli if possible.
 * [X] Ran `mycli --checkup`, if supported.

### Expected Behavior
Suggestions to always start showing up after the first character is typed.

<img width="385" height="167" alt="Image" src="https://github.com/user-attachments/assets/a4efd2c1-711d-43bd-8824-ecce55c19936" />

### Actual Behavior
After a backtick, suggestions only start showing up after the third character is typed.

<img width="385" height="29" alt="Image" src="https://github.com/user-attachments/assets/94fea67d-5d12-455f-9a78-545b26b0ed70" />
<br>
<img width="504" height="88" alt="Image" src="https://github.com/user-attachments/assets/c2db3634-ac0f-4b41-b157-722b5745c58b" />

### Steps to Reproduce
Type something like ``select `jso`` and see when completions start.

### System

 * mycli version: 1.53.1 (main branch)
 * OS/version: WSL2 Ubuntu 24.04 under Windows 11

### Discussion
This may be expected behavior, but the reason was not immediately obvious.

Note this behavior might change or be suppressed by #1547 depending on the setting.

rolandwalker:
It's also suggesting keywords when it could be suggesting only identifiers.

rolandwalker:
The current behavior seems to be that when introducing say a column with a backtick, only columns which _require_ a backtick are suggested.

PR #1574, opened 2026-02-16
Refactor `is_inside_quotes()`

rolandwalker:
## Description
 * Rename without underscore, as it might be used in another file.
 * Recognize backtick quoting, including doubled backticks as escapes.
 * Accept negative numbers for `pos`.
 * `escaped` should not be toggled to `True` unless inside a double- or single-quoted string.
 * Return a string or `False`, typed as Literal.
 * Optimize: `is not in` is much faster than looping.

Motivation: to improve completions which start with backtick.

xref #1564 .

## Checklist

    ```bash
    ```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -10,6 +10,7 @@ Bug Fixes
 Internal
 --------
 * Tune Codex reviews.
+* Refactor `is_inside_quotes()` detection.
 
 
 1.54.0 (2026/02/16)
--- a/mycli/packages/completion_engine.py
+++ b/mycli/packages/completion_engine.py
@@ -1,6 +1,6 @@
 import functools
 import re
-from typing import Any
+from typing import Any, Literal
 
 import sqlparse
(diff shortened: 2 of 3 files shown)


rolandwalker:
Addressed both of the Codex findings.  The doubling of backticks as escapes was a good point.

scottnemes:
Could you give some examples of what you're addressing here? It's unclear to me what the goal is. I.e. I see:

```
select "json = no suggestions
select 'json = no suggestions
select `jso = suggestions after three characters
select ``js = suggestions after two characters
```

rolandwalker:
Sorry if I was unclear @scottnemes .  This is just an enabling refactor, which does not itself change any functionality other than fixing some bugs in which "inside quotes" was wrongly determined.

But, we would need this to attack https://github.com/dbcli/mycli/issues/1564 .

scottnemes (approved):

PR #1581, opened 2026-02-18
Improve completion suggestions inside backticks

### LINKED ISSUE #1564: Difference in completion behavior after a backtick

rolandwalker:
## Description
Previously the behavior was a little janky: on the first character after a backtick, only identifiers which might _require_ a backtick were offered as suggestions (because, for instance, they matched reserved words).  Sometimes, that list would be empty, and no suggestions were offered.  Then, after typing a few more characters, rapidfuzz matching kicked in, and more suggestions were offered, with backticks off for the additional rapidfuzz suggestions.

Now the behavior is more consistent: if a backtick is typed, _all_ suggestions which could work in that place are offered, with uniform backticks on the suggestions, even if backticks are not required for the given identifier.

Of course, how early the suggestions are offered is still dependent on the `min_completion_trigger` option in `~/.myclirc`, so precise behavior in the above paragraph is conditional on having the default settings.

This PR still struggles on the case of spaces within the backtick, which I'd like to leave for future work, and an `xfail` test is included to represent the todo.

Fixes #1564 .

Changes

 * Tuck optimization check inside `_find_doubled_backticks()`, and make the optimization test for double instead of single backticks.
 * Remove function `unescape_name()`, which is unused, and seems wrongly named since it oriented towards strings rather than identifiers.
 * Let `find_matches(`) add backticks to all suggestions if backtick quoting is detected at the cursor.
 * Recast a variable name in `_find_doubled_backticks()`.

Per some comments in the tests, it would be nicer if column names sorted more strongly to the top in the `SELECT` context, but that is not new to these changes.

Another idea could be sorting to the top only those suggestions which require a backtick, when in the backtick context -- something for the future.

We also seem to be needlessly suggesting some generic keywords in the `SELECT` context, but that is also not new to these changes.  And if they are going to appear, it seems to make more sense to have them in backticks like the other suggestions, for the sake of uniformity.

Example of the new behavior:
new_backtick

## Checklist

    ```bash
    ```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,7 @@ Upcoming (TBD)
 Features
 ---------
 * `--checkup` now checks for external executables.
+* Improve completion suggestions within backticks.
 
 
 Bug Fixes
@@ -11,6 +12,7 @@ Bug Fixes
 * Watch command now returns correct time when ran as part of a multi-part query (#1565)
 * Don't diagnose free-entry sections such as `[favorite_queries]` in `--checkup`.
 * When accepting a filename completion, fill in leading `./` if given.
+
 
 Internal
 --------
--- a/mycli/packages/completion_engine.py
+++ b/mycli/packages/completion_engine.py
@@ -45,8 +45,12 @@ def _is_where_or_having(token: Token | None) -> bool:
 
 def _find_doubled_backticks(text: str) -> list[int]:
     length = len(text)
-    doubled_backticks: list[int] = []
+    doubled_backtick_positions: list[int] = []
     backtick = '`'
+    two_backticks = backtick + backtick
+
+    if two_backticks not in text:
+        return doubled_backtick_positions
 
     for index in range(0, length):
         ch = text[index]
@@ -54,13 +58,13 @@ def _find_doubled_backticks(text: str) -> list[int]:
             index += 1
             continue
         if index + 1 < length and text[index + 1] == backtick:
-            doubled_backticks.append(index)
-            doubled_backticks.append(index + 1)
+            doubled_backtick_positions.append(index)
+            doubled_backtick_positions.append(index + 1)
             index += 2
             continue
         index += 1
 
-    return doubled_backticks
+    return doubled_backtick_positions
 
 
 @functools.lru_cache(maxsize=128)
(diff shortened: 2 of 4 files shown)


rolandwalker:
> 1 Update backtick context detection to use document.text_before_cursor (or otherwise pass the full buffer) when calling is_inside_quotes() in find_matches().

Mostly correct, addressed.

> Add the whitespace-in-backticks completion test above.

Added as an `xfail`, because the issue with spaces runs deeper.

scottnemes:
This one is a bit complex, and my gut says to scale back the suggestions. You can't run most things surrounded by backticks, so suggesting most keywords and what not often does not make sense. I.e.

```
`SELECT` * FROM blah;
```

From a user perspective, it seems like it would make more sense to only ever suggest object type items when suggesting inside backticks; databases, tables, columns, procedures, functions. If a user is wanting a reserved word/keyword to show up inside backticks, it means they're using it as a database/table/column name most likely anyway, and would be covered by only suggesting those objects to begin with.

rolandwalker:
select_select_inappropriate_keywords

I think this is explained in the description pretty well.  We _are_ completing too many keywords, but it isn't new to this bugfix, nor particular to backticks at all.

Future work could include:
 * better handling spaces within backticks (we currently start a new completion)
 * recognizing a special backtick context within which only identifiers can be completed
 * sorting column names closer to the top in the SELECT + backtick context
 * sorting identifiers which actually _need_ backticks closer to the top in the backtick context

but it is highly preferable to do those things in separate PRs.

The cause of exuberantly offering inappropriate keywords as completions could be #1447 .  We might need to do better (manual) separation of functions vs other keywords.  In addition, the `favorite_keywords` from `sqlcompleter.py` seem to always be offered, and we expanded that list:

last image

scottnemes:
Understood. This just brings all the issues to the forefront, but with a path forward it will be better in the long run. If you put all of these things that need to be done in either a todo list or issues, I can help work on whatever. Have noticed a few PRs where you list things that would be nice to have but not sure those are getting tracked anywhere outside your brain anyhow. 🧠

rolandwalker:
I guess I became too obsessed with trying to get the number of open Issues down to 10!

PR #1582, opened 2026-02-18
Fix regression accepting pathname completion

### LINKED ISSUE #1557: Regression in completions on filenames with `source`
After #1500 there seems to be a regression on filename completions (it needs to be confirmed that #1500 is the cause).

* When choosing a completion the displayed text is not precisely what is placed into the buffer, enough so that the filename may no longer be valid.
* In addition, the completion popup does not go away when expected.
* In addition, when successfully completing a directory name, the popup doesn't show the files inside the completed directory.

In the images below the leading `./` gets turned into a `.` on completion:

Image

<kbd>TAB</kbd>

Image

Presumably this has to do with a length calculation being off, which suggests that ignoring `position` here
 * https://github.com/dbcli/mycli/pull/1500/changes#diff-878f97580507e93ffbe6d032f603bdf63c3d105854afbf416662ddbd18b7b3fbR1240

is the cause.

There's not a oneline fix, because the refactor in #1500 changed the `find_` functions to return plain strings instead of `Completion`s.

rolandwalker:
## Description
Ensure that when accepting a pathname completion, the path as typed is accepted and not modified into an invalid path.  See #1557 for an image.  The regression is most evident when typing a path with a leading `./`.

Changes
 * Use a `Literal` for the `include` argument to `last_word()`
 * Create a separate length-measurement string for pathname completions, based on a different last-word cleanup regex.
 * Remove some todo comments from tests which are now resolved.

The logic of `length_based_on_path` assumes that if _any_ of the suggestions are pathnames, then _all_ of the suggestions are pathnames, which is currently true, and likely to remain true.

It would be nice to have a test which simulated accepting the first suggestion, to check that the filled-in completion does not get mangled, as happened before.

Let's say this fixes #1557, because it fixes the actual regression.  There are some other behaviors listed there which may not be desirable but were not caused by #1500 .  Let's open separate issues if we want to do further work on pathname completion.  For example, there is no pathname completion for `\edit <pathname>` at all.  And completion on `~/` does not seem to provide candidates.

I'm also not 100% clear on the subtleties of `last_word()`, and why the `alphanum_underscore` cleanup regex is the best one for this case, but it does work exactly right per @scottnemes 's diagnosis, and the relevant change was indeed introduced in #1500.  Maybe there should be a cleanup regex named `pathnames`, for clarity.

## Checklist

    ```bash
    ```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,6 +5,7 @@ Bug Fixes
 ---------
 * Watch command now returns correct time when ran as part of a multi-part query (#1565)
 * Don't diagnose free-entry sections such as `[favorite_queries]` in `--checkup`.
+* When accepting a filename completion, fill in leading `./` if given.
 
 
 1.54.1 (2026/02/17)
--- a/mycli/packages/parseutils.py
+++ b/mycli/packages/parseutils.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import re
-from typing import Any, Generator
+from typing import Any, Generator, Literal
 
 import sqlglot
 import sqlparse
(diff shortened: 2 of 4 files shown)


scottnemes (approved):
Nice! Verified path completion behavior is working as expected.

PR #1590, opened 2026-02-20
Make `--character-set` an alias for `--charset` at the CLI

rolandwalker:
## Description
Make `--character-set` an alias for `--charset` at the CLI, for alignment with the spelling of the equivalent option in `~/.myclirc`.

Internally, refactor to use `character_set` instead of `charset` wherever possible.  The form `charset` is limited to the invocation of `pymysql's connect()`.

## Checklist

    ```bash
    ```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,7 @@ Upcoming (TBD)
 Features
 ---------
 * Let the `--dsn` argument accept literal DSNs as well as aliases.
+* Accept `--character-set` as an alias for `--charset` at the CLI.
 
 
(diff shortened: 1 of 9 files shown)


rolandwalker:
@scottnemes I'm fine with accepting both indefinitely.  We have enough pending deprecations!

But, I pushed an amended version of the PR and am re-requesting review.  It is functionally identical, but it goes farther and changes all the internal uses of `charset` to `character_set`.  Sort of making `--character-set` the default, and `--charset` the alias.

Issue #1584, opened 2026-02-18
Cannot connect to cloud-sql-proxy with passwordless login


ptwales:
I'm running `cloud-sql-proxy` with `--auto-iam-authn` to connect to a google hosted sql database.  I cannot connect to that database proxy with `mycli>=1.43.0`.  Older versions of mycli work (latest 1.42.0), so does usql and the stock mariadb client.

### Expected Behavior

While running
```
cloud-sql-proxy --auto-iam-authn <PROJECT>:<ZONE>:<DB> --port 3308
```

I should be able to immediately connect to that database without providing a password with 
```
mycli -h 127.0.0.1 -P 3308 -u <USER> -D <DATABASE>
```

### Actual Behavior
 
I get the output

```
# 2026-02-18 12:12:47,988 WARNING mycli.config config.py:23 -- Unable to parse line 10 of config file '/etc/my.cnf'.

# 2026-02-18 12:12:47,988 WARNING mycli.config config.py:23 -- Using successfully parsed config values.

# 2026-02-18 12:12:47,988 WARNING mycli.config config.py:23 -- Unable to parse line 52 of config file '/etc/my.cnf.d/server.cnf'.

# 2026-02-18 12:12:47,988 WARNING mycli.config config.py:23 -- Using successfully parsed config values.

Enter password for <USER>: <CR>
[...]

PR #1591, opened 2026-02-20
[fix] Allow users to use empty passwords without being prompted or having to provide an empty password option (#1584)

scottnemes:
## Description
A previous password logic rework made it so users could not use an empty password without either specifying it in the config / CLI option or being prompted. This PR fixes that issue by attempting to connect with an empty password if no password is found through any of the available password sources. If the connection fails with an empty password, it will then prompt the user to enter a password.

## Checklist

    ```bash
    ```

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -532,7 +532,7 @@ class MyCli:
         host: str | None = "",
         port: str | int | None = "",
         socket: str | None = "",
-        charset: str | None = "",
+        character_set: str | None = "",
         local_infile: bool = False,
         ssl: dict[str, Any] | None = None,
         ssh_user: str | None = "",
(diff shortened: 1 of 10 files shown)


scottnemes:
This PR addresses the password portion of #1584; will create a second PR to add some additional error output for SSL errors.

scottnemes:
@rolandwalker Could use whatever password related testing you want to throw at this, especially for the keyring as I'm not as familiar with that. But as best I can tell everything still works as expected

mycli/main.py
```diff
@@ -660,63 +660,63 @@ def connect(
...
+                    connection_info["ssl"] = None
+                    _connect(retry_ssl=True)
+                elif e1.args[0] == ACCESS_DENIED_ERROR and connection_info["password"] is None:
+                    # if we already tried and failed to connect with a new password, raise the error
+                    if retry_password:
+                        raise e1
+                    # ask the user for a new password and try to connect again
+                    new_password = click.prompt(
+                        f"Enter password for {user}", hide_input=True, show_default=False, default='', type=str, err=True
+                    )
```
rolandwalker, line 717: Shouldn't we update the keyring here?

scottnemes: It should get updated on the retry (line 700)
https://github.com/dbcli/mycli/pull/1591/changes#diff-8e9c55cacc296c33014d775f5deffcb0aafeac0e0026ffebb9acb60911becc25R700

PR #1592, opened 2026-02-20
Add SSL/TLS version to `status` output

rolandwalker:
## Description

last image

## Checklist

    ```bash
    ```


scottnemes (approved):
Verified SSL version is shown when connected via SSL, and that `<null>` is shown when connected without SSL.

PR #1600, opened 2026-02-21
Let `--ssl_mode=auto` prefer SSL only for TCP/IP

rolandwalker:
## Description
While it is _possible_ to negotiate an SSL connection to MySQL over a local socket, it shouldn't be needed (should it?).

Nothing is transmitted over a network in that scenario.

If the user still does need SSL/TLS, `--ssl-mode=on` is available to force it.

@scottnemes does this PR make sense?  I'm open to being convinced otherwise.  It just made sense to me.   (There could even be a fourth ssl_mode but that seemed like needless complexity.)

## Checklist

    ```bash
    ```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -11,6 +11,7 @@ Features
 * Fully deprecate the built-in SSH functionality.
 * Let `--keepalive-ticks` be set per-connection, as a CLI option or DSN parameter.
 * Accept `character_set` as a DSN query parameter.
+* Don't attempt SSL for local socket connections when in "auto" SSL mode.
 
 
 Bug Fixes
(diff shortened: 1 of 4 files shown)


rolandwalker:
I _think_ bot's suggestion 1 is wrong, and I'm not concerned about the interaction with the deprecated `--ssl` option.

scottnemes (approved):
Yeah according to "experts" SSL isn't needed for a socket connection, so makes sense for that. Likely little impact in reality for our use case, but since it's a fairly straightforward fix then might as well.

Verified SSL is no longer used when connecting via socket and `ssl-mode=auto` and SSL is used with socket and `ssl-mode=on`.

PR #1601, opened 2026-02-21
Add prompt format string for SSL/TLS version

rolandwalker:
## Description
`\T` will show the TLS version for the connection, or `(none)` when appropriate.

The negotiated version seems to require a trip to the server, but `get_ssl_version()` has been cached per `thread_id` so that we don't need to make that trip for every prompt refresh.

xref #1592 

## Checklist

    ```bash
    ```


scottnemes (approved):
Verified new \T prompt value works with and without SSL enabled.

PR #1605, opened 2026-02-21
Allow multiline input in batch mode on the standard input

### LINKED ISSUE #1593: multi_line from stdin no longer works
I recently upgraded from 1.44.2 to 1.55.0 and I can no longer run multi-line statements from stdin. I have a collection of .sql files that I use to run a variety of queries from the cli via stdin but this no longer works if a query is spread over multiple lines.

This works as expected:
```
$ echo "SELECT id FROM table LIMIT 1;" | mycli db
id
16757
$ echo "SELECT id FROM table LIMIT 1;
SELECT id FROM table LIMIT 1;" | mycli db
id
16757
id
16757
```

This does not:
```
$ echo "SELECT id
FROM table LIMIT 1;" | mycli db
(1054, "Unknown column 'id' in 'SELECT'")
```

I have `multi_line = True`. Within mycli, multi-line support works as expected — this seems to affect only when input is from stdin.

rolandwalker:
## Description
Tokenize each line of input with sqlglot, and dispatch the (possibly multi-statement) query if the last token is a semicolon.  If the last token is _not_ a semicolon, accumulate the line towards the next dispatch.

We don't handle the case where the input script itself changes the delimiter.

A limit of 5000 lines is set, after which, if we can't find a line ending in semicolon, we assume that something is wrong with the input and exit.

Caused (intentionally) by #1450, but we still can improve on it.

Fixes #1593 .

## Checklist

    ```bash
    ```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -21,6 +21,7 @@ Bug Fixes
 * Make `--ssl-capath` argument a directory.
 * Allow users to use empty passwords without prompting or any configuration (#1584).
 * Check the existence of a socket more directly in `status`.
+* Allow multi-line SQL statements in batch mode on the standard input.
 
 
 1.55.0 (2026/02/20)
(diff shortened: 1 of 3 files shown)


rolandwalker:
Let's punt on the configuration option until someone asks for it, or until we have a well thought out `[batch]` section.

PR #1606, opened 2026-02-21
Accept true/false for `--ssl_mode`, keeping on/off

rolandwalker:
## Description
More liberally accept true/false/on/off (and capitalization variants) for `--ssl_mode`.

Some inconsistencies in the interface have crept in as features were added.

Some settings from `~/.myclirc` especially, but also in the CLI arguments, use `str_to_bool()` for liberal boolean parsing.  Other settings, such as `--ssl_mode`, are using on/off only.  Other settings, such as `--use_keyring`, are using true/false only.

This change is a step towards converging all boolean settings on liberal `str_to_bool()`.  It should ultimately include DSN query parameters, too.

The "auto" value for `ssl_mode` is still special-cased, and not affected by these changes.

The only loss is that click doesn't automatically display the choices in the helpdoc, but they are already listed in the description.  We keep the description targeted to on/off, and only silently accept the other variants.

## Checklist

    ```bash
    ```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -10,6 +10,7 @@ Features
 * Let `help <keyword>` list similar keywords when not found.
 * Optionally highlight fuzzy search previews.
 * Make `\edit` synonymous with the `\e` command.
+* More liberally accept `on`/`off` values for `true`/`false`, and vice versa.
 
 
 Bug Fixes
--- a/mycli/config.py
+++ b/mycli/config.py
@@ -305,6 +305,13 @@ def str_to_bool(s: str | bool) -> bool:
         raise ValueError(f'not a recognized boolean value: {s}')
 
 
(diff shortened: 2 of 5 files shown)


scottnemes (commented):
Agree consistency is good, but `--ssl_mode = true/false` doesn't make sense from a naming perspective (it's not like `--ssl = true/false`, which makes more sense). A mode isn't true or false, it's on / off / auto. Converting that to true/false on the backend could make sense if desired for type checking, but I wouldn't expose it to the user.

Also the point of choice was to restrict it to those values, so isn't any issue with people putting in unexpected values.

rolandwalker:
We never reached an agreement about accepting liberal boolean values at the CLI.

PR #1607, opened 2026-02-21
Accept on/off values for `--use-keyring`

rolandwalker:
## Description
Accept on/off values for `--use-keyring`, keeping true/false, as well as treating "reset" as special.

The motivation is uniformity in the interface.

The only loss is that click does not display the value choices in the helpdoc, but they are already listed in the description.

Here we also change the description to suggest on/off, while continuing to silently accept true/false.  The reason for the change is that `--ssl-mode` came first, and `--ssl-mode` takes on/off.

See #1606  for more on the motivation.

## Checklist

    ```bash
    ```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -10,6 +10,7 @@ Features
 * Let `help <keyword>` list similar keywords when not found.
 * Optionally highlight fuzzy search previews.
 * Make `\edit` synonymous with the `\e` command.
+* More liberally accept `on`/`off` values for `true`/`false`, and vice versa.
 
 
 Bug Fixes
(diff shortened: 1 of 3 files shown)


scottnemes:
Similar comment to the ssl_mode one, the naming of `--use-keyring` actually makes sense for true/false, but doesn't really make sense for on/off. If anything this is more about option naming vs values.

I think setting clear expectations in the command/option is the most important anyway, so if it's true/false, then the option should be bool, and that is clear to the user. If it's not true/false, and has a set of values like ssl_mode (on/off/auto), then choice is preferred to enforce that.

rolandwalker:
We never reached an agreement about accepting liberal boolean values at the CLI.

PR #1608, opened 2026-02-21
Reduce extraneous prompt refreshing

rolandwalker:
## Description
Only refresh prompt escapes when the entered text is empty.  The current behavior is to refresh on every keystroke, which for dynamic format strings, such as the current time, is needless and distracting.

With this change, we usually refresh the prompt only after starting a new query.

Changes

 * Move the `message` parameter to `prompt()` from `PromptSession()`, so that the value can be loaded with the application itself.
 * Recast `get_message()` as `get_prompt_message()` for clarity.
 * In the `get_prompt_message()` callback, peek at the content which the user has typed, and return a cached value if there is any current text.

There still can be an extraneous refresh if the user deletes all content from the line, creating an empty line.

## Checklist

    ```bash
    ```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -23,6 +23,7 @@ Bug Fixes
 * Allow users to use empty passwords without prompting or any configuration (#1584).
 * Check the existence of a socket more directly in `status`.
 * Allow multi-line SQL statements in batch mode on the standard input.
+* Fix extraneous prompt refresh on every keystroke.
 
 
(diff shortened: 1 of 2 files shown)


rolandwalker:
~I've realized that the call in `handle_editor_command()` isn't actually interactive,~ 
Edit: this is wrong.  It is interactive.

but want to still check the LLM usage.
Edit: checked.

PR #1627, opened 2026-02-25
Add help on F1 keystroke

rolandwalker:
## Description
 * open documentation index in web browser on F1
 * emit output using `prompt_toolkit` methods such that the prompt is restored including pending input
 * include the docs URL in the output in case the browser did not open
 * include alternate F1 key sequence that `prompt_toolkit` doesn't handle
 * update `TIPS` and `key_bindings.rst`

## Checklist

    ```bash
    ```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,7 @@ Upcoming (TBD)
 Features
 ---------
 * Add `\bug` command.
+* Let the `F1` key open a browser to mycli.net/docs and emit help text.
 
 
 Bug Fixes
--- a/doc/key_bindings.rst
+++ b/doc/key_bindings.rst
@@ -7,6 +7,12 @@ Most key bindings are simply inherited from `prompt-toolkit <https://python-prom
 The following key bindings are special to mycli:
 
 ###
+F1
+###
+
+Open documentation index in a browser tab.
+
+###
 F2
 ###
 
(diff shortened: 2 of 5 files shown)


rolandwalker:
Interesting!   Does this command not work on your setup?

```bash
python -m webbrowser -t "https://www.python.org"
```

A try/except wouldn't help.  The error output is coming from an external program.  But I think the issue can be fixed by the same invalidation trick as in #1632 .

rolandwalker:
See if the pushed changes fix the spurious output for you by invalidating the display.

scottnemes:
> Interesting! Does this command not work on your setup?
> 
> ```shell
> python -m webbrowser -t "https://www.python.org"
> ```
> 
> A try/except wouldn't help. The error output is coming from an external program. But I think the issue can be fixed by the same invalidation trick as in #1632 .

I use WSL2 on Windows11 for most of my testing, so always a bit different. I am able to install packages to make it work to load the browser though, so verified that bit works.

However the error output is the same after your most recent change. But it looks fine in general, and I couldn't find any good way to suppress the error without getting hacky. So should be good as-is.

rolandwalker:
That's good to know.  Maybe we could improve our Windows support, both in and out of WSL2.  Maybe we could at least improve the documentation.

PR #1629, opened 2026-02-25
Rewrite bottom toolbar, but keep it compact

rolandwalker:
## Description
* Recast `show_fish_help()` as `show_initial_toolbar_help()` since it is be used more generally.
 * Recast `show_suggestion_tip()` as `show_initial_toolbar_help()` so that the caller matches the callee.
 * Make a toolbar section divider with a vertical bar.
 * Add Tab to permanent list of suggested keys.
 * Add F1 to permanent list of suggested keys (assumes #1627).
 * Add F2 to permanent list of suggested keys and show smart-complete status.
 * Only highlight the "ON" part of multiline when on, and remove space. Constraining the total highlighted characters to a smaller number makes the line in general more readable, though the styling of the highlighted letters depends on the configuration..
 * Only highlight the vi modes when vi edit mode is on, and remove space.
 * Only show delimiter text if non-standard or initial, and make the text shorter.
 * Make right-arrow help text explain _which_ suggestions are referred to, avoiding the confusing word "complete".
 * Make "refreshing" message shorter with a Unicode ellipsis.

A key feature of the reorganization is that transient text only can appear on the right, so the keystroke help stays more aligned.

Toolbar in "initial" mode with extra notes to the right (specific colors/styles depend on user settings):

toolbar_initial

Toolbar after "initial" mode:

toolbar

## Checklist

    ```bash
    ```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,7 @@ Features
 * Add `\bug` command.
 * Let the `F1` key open a browser to mycli.net/docs and emit help text.
 * Add documentation index URL to inline help.
+* Rewrite bottom toolbar, showing more statuses, but staying compact.
 
 
 Bug Fixes
--- a/mycli/clitoolbar.py
+++ b/mycli/clitoolbar.py
@@ -7,35 +7,60 @@ from prompt_toolkit.key_binding.vi_state import InputMode
 from mycli.packages import special
 
 
-def create_toolbar_tokens_func(mycli, show_fish_help: Callable) -> Callable:
+def create_toolbar_tokens_func(mycli, show_initial_toolbar_help: Callable) -> Callable:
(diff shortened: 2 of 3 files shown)


scottnemes:
Not sure if expected, but it takes two enters to get from the initial toolbar to the final shorter one:
<br>
Iniitial:
image
<br>
First enter:
image
<br>
Second enter:
image
<br>
And if you manage to launch without the "refreshing completions" bit showing up, it still takes two enters to get from initial to the final shorter version.

Also F2 for smart completions doesn't update the toolbar, but does seem to update the variable as functionality does appear to change.

rolandwalker:
@scottnemes the "two enters" thing has always been that way and was untouched here.

The smart-complete indicator is a great catch; fixed.

PR #1646, opened 2026-02-27
Make prettify/unprettify logic more robust

rolandwalker:
## Description
  * explicitly handle more exceptional cases such as empty input
  * clarify in error message that only a single statement is expected
  * don't return an empty string on failure; instead return the original text

Works together with #1645.

## Checklist

    ```bash
    ```


rolandwalker:
Merging this without review, since it is pretty much required for https://github.com/dbcli/mycli/pull/1645, which I merged without noticing the state of this one.

PR #1655, opened 2026-02-28
Bottom toolbar format string for customization

rolandwalker:
## Description
Like the prompt, the bottom toolbar can be customized, using the same format strings, with a special format string `\B` to represent the standard toolbar (in the first position only).

When `\B` is included the user's customizations appear on the second line.

When `\B` is not included, the user may override the toolbar completely. Transient notices will still appear to the right of the user's format.

Like the prompt, both a CLI option `--toolbar` and a `~/.myclirc` option are provided, with the CLI option taking precedence.

Motivation: useful information takes up repeated space in the prompt, but needs to be displayed only once in the toolbar.

Example:

status_bar_styled

## Checklist

    ```bash
    ```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,7 @@ Features
 * Offer filename completions on more special commands, such as `\edit`.
 * Allow styling of status, timing, and warnings text.
 * Set up customization of prompt/continuation colors in `~/.myclirc`.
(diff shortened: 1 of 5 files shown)


scottnemes (approved):

scottnemes:
Verified toolbar customization works as expected

PR #1657, opened 2026-02-28
Avoid calling `get_prompt()` unless needed

rolandwalker:
## Description
Since some prompt escapes are expensive, and can even require a trip to the server, avoid calling `get_prompt()` unless needed, preferring to use the cached value in the `last_prompt_message` property, or a new saved value for the number of lines in the prompt.

Incidentally, explicitly strip ANSI formatting from prompts before writing them to a file, when `tee` is in effect.

## Checklist

    ```bash
    ```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -15,6 +15,7 @@ Features
 Bug Fixes
 ---------
 * Make toolbar widths consistent on toggle actions.
+* Don't write ANSI prompt escapes to `tee` output.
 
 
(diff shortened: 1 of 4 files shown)


rolandwalker:
Ah, so something is still very wrong.

rolandwalker:
With caching, it's down now to 2 or 3 calls per prompt from just pressing return.

PR #1660, opened 2026-02-28
Use prompt_toolkit's clipboard for the `\clip` command

rolandwalker:
## Description
This currently has the same functionality as depending on `pyperclip` directly; however it simplifies things to depend only on `prompt_toolkit`, reducing the chance that there is ever a conflict over the `pyperclip` version declarations.

Incidentally remove a needless f-string.

Part of a series: trying to depend on `prompt_toolkit` more, handroll less, and depend on `click` less for output.

## Checklist

    ```bash
    ```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,11 @@
+Upcoming (TBD)
+==============
+
+Internal
+---------
+* Use the clipboard provided by prompt_toolkit for `\clip`.
+
+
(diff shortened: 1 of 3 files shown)


rolandwalker:
Closing, because prompt_toolkit does not actually install pyperclip transitively.  So there is no benefit.

PR #1663, opened 2026-02-28
Handle/document more attributes such as "dim"

rolandwalker:
## Description
Handle and document more attributes for the `[colors]` section of the `~/.myclirc` configuration file.

This is a bit tricky as we try to support both Pygments and prompt_toolkit attribute names, and there is some mismatch between them.

Before adding any style item we check it in a `try` block to see if it is supported (for the software, not the terminal).  This works fine, with the only downside being that we silently swallow some spelling errors, applying no style in that case.  We cannot reasonably warn in all such cases, since there exist styles supported by one library and not the other.

In the end, we gain the support of new attributes such as `dim` and `strike`.  Some, such as `hidden`, may not be supported most terminals.

Other attributes exist, such as `roman`, but it isn't clear that they are supported anywhere, so they are not documented.

The attribute that is most likely to be useful is `dim`.  (I would like the border colors of tables to be dimmed).

last image

## Checklist

    ```bash
    ```

### DIFF
--- a/mycli/clistyle.py
+++ b/mycli/clistyle.py
@@ -104,6 +104,28 @@ def parse_pygments_style(
         return token_type, style_dict[token_name]
 
 
+def is_valid_pygments(name: str) -> bool:
+    try:
+
+        class TestStyle(PygmentsStyle):
+            default_style = ''
+            styles = {Token.Default: name}
+
+        return True
+    except AssertionError:
+        # can't emit error because some styles are valid pygments and not valid ptoolkit
+        return False
+
+
+def is_valid_ptoolkit(name: str) -> bool:
+    try:
+        _s = Style([("default", name)])
+        return True
+    except ValueError:
+        # can't emit error because some styles are valid pygments and not valid ptoolkit
+        return False
+
+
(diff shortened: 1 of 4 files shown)


mycli/clistyle.py
```diff
@@ -96,6 +96,28 @@ def parse_pygments_style(
         return token_type, style_dict[token_name]
 
 
+def is_valid_pygments(name: str):
```
scottnemes: For both these new functions, shouldn't they have a return type of bool? Surprised mypy doesn't complain about this.

For the error case of both functions, would it make sense to just return false for any exception? Otherwise it's not returning anything in the case a different error occurs (unless you want to actually do something on different errors). And seems weird one is `AssertionError` and one is `ValueError`, unless they actually do return those different error types (but would be moot if you just return false regardless).

rolandwalker: Good catch on the return type!  We don't have mypy settings as strict as requiring them.

Yes, those are the actual expected exceptions for the two different cases, and we want to throw if some other error happens.

PR #1829, opened 2026-04-08
Feat/440/add sandbox mode

### LINKED ISSUE #440: Add support for expired accounts
MySQL 5.7 and newer can set password for account expired:

https://dev.mysql.com/doc/refman/5.7/en/password-expiration-policy.html

When this is done, account can login, however all what can be done is to
reset password.  It seems mycli don't understand this feature:

mysql root@localhost:(none)> CREATE USER 'tst'@'localhost' IDENTIFIED BY 'MyNewPass4.';
Query OK, 0 rows affected
Time: 0.002s
mysql root@localhost:(none)> ALTER USER 'tst'@'localhost' PASSWORD EXPIRE;
Query OK, 0 rows affected
Time: 0.001s
mysql root@localhost:(none)> exit
Goodbye!
$ mycli -utst
Password:
(1862, 'Your password has expired. To log in you must change it using a client that supports expired passwords.')

Would be nice if mycli could add such support.

scottnemes:
## Description

This PR adds the ability for mycli to go into "sandbox mode" when a user with an expired passwords is used to connect. Prior to this change, when a user with an expired password connects it will simply tell them the password is expired. With this change, the user goes into a restricted "sandbox" mode, which only allows the user to change their password, or quit. Once the password is changed, the client will then automatically reconnect using the updated password and go into the normal "full" mode and allow normal operation.

The behavior may be controlled by either the `connect_expired_password` config option (True by default), or the `--connect-expired-password` flag.

Expire user's password:
```
alter user 'noob'@'localhost' password expire;
```

Existing behavior when logging in with the expired password user:
```
uv run mycli -u noob -p
Enter password for noob:
(1862, 'Your password has expired. To log in you must change it using a client that supports expired passwords.')
```

New behavior when logging in with the expired password user:
```
uv run mycli -u noob -p
Enter password for noob:
Your password has expired. Use ALTER USER to set a new password, or quit.
mycli 1.44.3.dev684+ga5308b2c0.d20260408
Home: https://mycli.net
Bug tracker: https://github.com/dbcli/mycli/issues
Thanks to the contributor — Scrappy Soft
MySQL noob@127.0.0.1:(none)>
```

Trying to run a command:
```
MySQL noob@127.0.0.1:(none)> use dev;
ERROR 1820: You must reset your password using ALTER USER or SET PASSWORD before executing this statement.
MySQL noob@127.0.0.1:(none)>
```

Change password:
```
MySQL noob@127.0.0.1:(none)> alter user 'noob'@'localhost' identified by 'yolo';
You're about to run a destructive command.
Do you want to proceed? (y/n): y
Your call!
Query OK, 0 rows affected
Time: 0.009s
Password changed successfully. Reconnected.
MySQL noob@127.0.0.1:(none)> use dev;
You are now connected to database "dev" as user "noob"
Time: 0.001s
MySQL noob@127.0.0.1:dev> select * from t2;
+----+-----+
| id | def |
+----+-----+
|  1 | d   |
|  2 | e   |
|  3 | f   |
+----+-----+

3 rows in set
Time: 0.005s
MySQL noob@127.0.0.1:dev>
```

Note it then reconnects using the updated password and you are free to continue on as normal.

Resolves #440 

## Checklist

    ```bash
    ```


scottnemes:
@rolandwalker This is a pretty big one, so feel free to suggest any changes you might want for the general flow or messaging or whatever. Tried to keep it as minimal as possible to get the functionality, but it touches a lot of areas since a normal login has a lot going on behind the scenes as you know.

mycli/main_modes/repl.py
```diff
@@ -518,6 +519,52 @@ def _build_prompt_session(
...
+    return bool(_PASSWORD_CHANGE_RE.match(text.strip()))
+
+
+_IDENTIFIED_BY_RE = re.compile(
+    r"IDENTIFIED\s+BY\s+'([^']*)'",
+    re.IGNORECASE,
+)
+
+_SET_PASSWORD_RE = re.compile(
+    r"SET\s+PASSWORD\s*=\s*'([^']*)'",
```
rolandwalker: The expression could use double quotes instead of single quotes, and/or there can be an internal quote which is escaped.  Instead of a regex, we could use sqlglot to tokenize the statement:

```python
sqlglot.tokenize('SET PASSWORD = "auth_string"', dialect='mysql')
```

In addition there are some variations of the statement which are not handled, which at least deserves a comment.

mycli/main_modes/repl.py
```diff
@@ -518,6 +519,52 @@ def _build_prompt_session(
...
+
+def _is_sandbox_allowed(text: str) -> bool:
+    """Return True if the command is allowed in expired-password sandbox mode."""
+    stripped = text.strip()
+    if not stripped:
+        return True
+    return bool(_SANDBOX_ALLOWED_RE.match(stripped))
+
+
+def _is_password_change(text: str) -> bool:
```
rolandwalker: Perhaps these functions belong somewhere else such as `sql_utils.py`.  We can imagine this being used elsewhere: password-change lines could be excepted from history storage.

scottnemes:
@rolandwalker I switched the query parsing over to sqlglot and expanded what it checks for, so this should cover the normal use cases of `ALTER USER` and `SET PASSWORD`. Testing using a password with quotes and all of that and seems to work fine (i.e. `"pass'word"`)

rolandwalker (approved):
Sweet!

rolandwalker:
I'm going to merge this because it looks like we would hit merge conflicts today otherwise.

PR #1835, opened 2026-04-11
Don't persist password-change SQL statements to the history file

rolandwalker:
## Description
Leveraging the new `is_password_change()` detection from #1829, avoid persisting password-change statements to the history file, but leave them available for navigation in the current session.

Wrap the `sqlglot.tokenize()` call for `is_password_change()` in a `try` block, since it could throw if given invalid SQL.

Add and improve tests for `sql_utils.py`.

## Checklist

    ```bash
    ```

### DIFF
--- a/test/pytests/test_ptoolkit_history.py
+++ b/test/pytests/test_ptoolkit_history.py
@@ -2,6 +2,7 @@
 
 from pathlib import Path
 
+from mycli.packages.ptoolkit import history as history_module
 from mycli.packages.ptoolkit.history import FileHistoryWithTimestamp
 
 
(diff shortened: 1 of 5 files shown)


test/pytests/test_ptoolkit_history.py
```diff
@@ -13,6 +14,29 @@ def test_file_history_with_timestamp_sets_filename(tmp_path: Path) -> None:
...
 
 
+def test_append_string_caches_and_stores_non_password_statement(tmp_path: Path, monkeypatch) -> None:
+    history = FileHistoryWithTimestamp(tmp_path / 'history.txt')
+    stored: list[str] = []
+    monkeypatch.setattr(history, 'store_string', stored.append)
+
+    history.append_string('SELECT 1')
+
+    assert history._loaded_strings[0] == 'SELECT 1'
```
scottnemes: Could this (and the one below) use the public method `get_strings()` instead of this private method?

scottnemes (approved):
Looks good overall; see the one comment if you want to address it.

test/pytests/test_ptoolkit_history.py
rolandwalker: Good point.  Done!

PR #1853, opened 2026-04-18
Skip `--checkpoint` file statements with `--resume`

rolandwalker:
## Description

In `--batch` mode, when the batch input script is not STDIN, and when `--checkpoint` is also given, `--resume` causes mycli to replay the checkpoint file, looking for leading matching statements, and skip execution of batch statements already present in the checkpoint file.

Motivation: resumption of interrupted batch scripts.

The number of statements in the checkpoint file must be fewer than the number of statements in the batch script, and form a leading match, or mycli will exit without executing anything.

Once execution is picked up again from the midpoint of the `--batch` script, we continue to append _new_ statements to the checkpoint file, after each statement is successfully executed.  That behavior is unchanged.  This allows the checkpoint file to be used again if the batch script is interrupted multiple times.

The `--progress` bar and included ETA calculation account for the statements replayed from the checkpoint file, and show corrected views.

Further work could include creating a `[batch]` section in myclirc and adding a default value for resumption, with a `--no-resume` option at the CLI.

Note: some SQL statements change server/session state or start transactions.  But _any_ successful statement will be checkpointed and then not executed upon resumption in `--resume` mode.  It is incumbent on the user to account for such state when resuming from a checkpoint.

This PR represents a major goal in a series of PRs to strengthen batch execution, some of which are:

 * https://github.com/dbcli/mycli/pull/1450
 * https://github.com/dbcli/mycli/pull/1460
 * https://github.com/dbcli/mycli/pull/1466
 * https://github.com/dbcli/mycli/pull/1723
 * https://github.com/dbcli/mycli/pull/1731
 * https://github.com/dbcli/mycli/pull/1753

## Checklist

    ```bash
    ```

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -1354,7 +1354,12 @@ class CliArgs:
     )
     checkpoint: TextIOWrapper | None = clickdc.option(
         type=click.File(mode='a', encoding='utf-8'),
-        help='In batch or --execute mode, log successful queries to a file.',
+        help='In batch or --execute mode, log successful queries to a file, and skipped with --resume.',
+    )
+    resume: bool = clickdc.option(
+        '--resume',
+        is_flag=True,
+        help='In batch mode, resume after replaying statements in the --checkpoint file.',
     )
     defaults_group_suffix: str | None = clickdc.option(
         type=str,
@@ -1521,6 +1526,14 @@ def click_entrypoint(
     # option cannot be an empty string, but a password can be
     if cli_args.password is None and os.environ.get("MYSQL_PWD") is not None:
         cli_args.password = os.environ.get("MYSQL_PWD")
+
+    if cli_args.resume and not cli_args.checkpoint:
+        click.secho('Error: --resume requires a --checkpoint file.', err=True, fg='red')
+        sys.exit(1)
+
+    if cli_args.resume and not cli_args.batch:
+        click.secho('Error: --resume requires a --batch file.', err=True, fg='red')
+        sys.exit(1)
 
     cli_verbosity = 0
     if cli_args.verbose and cli_args.quiet:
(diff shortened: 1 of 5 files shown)


mycli/main_modes/batch.py
```diff
@@ -70,6 +118,7 @@ def main_batch_with_progress_bar(mycli: 'MyCli', cli_args: 'CliArgs') -> int:
         click.secho('--progress is only compatible with a plain file.', err=True, fg='red')
         return 1
     try:
+        completed_statement_count = replay_checkpoint_file(cli_args.batch, cli_args.checkpoint, cli_args.resume)
```
scottnemes, line 121: Looks like currently the batch file is getting opened three times; one in this new replay_checkpoint_file function, again on line 122, and again on line 172. Would it be possible to combine those somehow? Maybe returning something from replay_checkpoint_file to re-use.

mycli/main.py
```diff
@@ -1514,6 +1519,10 @@ def get_password_from_file(password_file: str | None) -> str | None:
     if cli_args.password is None and os.environ.get("MYSQL_PWD") is not None:
         cli_args.password = os.environ.get("MYSQL_PWD")
 
+    if cli_args.resume and not cli_args.checkpoint:
```
scottnemes, line 1522: Should you also check if `--batch` mode is given with `--resume`? I.e. could use `--execute` and `--resume` currently.

mycli/main_modes/batch.py
rolandwalker: That's a good point, but we would have to read the entire file into memory to provide the same guarantee, something we got away from in #1450 .

mycli/main.py
rolandwalker: Good idea!

scottnemes (approved):

PR #1855, opened 2026-04-18
More control over verbosity levels

rolandwalker:
## Description
 * let `--verbose` be given multiple times, incrementing a counter
 * add `--quiet` option to reduce the verbosity level
 * let CLI arguments always override config-file defaults, in this case the `main.less_chatty` option, which is made equivalent to `--quiet`

This initial implementation respects three verbosity levels:

 * -1 (quiet)
 * 0 (default)
 * 1 (verbose)

and doesn't yet adopt new behaviors for verbosity levels 2 or 3.

Special-command verbosity is recast to avoid confusion with application-setting verbosity.

Motivation: enable debugging logs to the console with `-vvv`.

## Checklist

    ```bash
    ```

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -143,6 +143,7 @@ class MyCli:
         warn: bool | None = None,
         myclirc: str = "~/.myclirc",
         show_warnings: bool | None = None,
+        cli_verbosity: int = 0,
     ) -> None:
         self.sqlexecute = sqlexecute
         self.logfile = logfile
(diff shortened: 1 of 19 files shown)


mycli/main.py
```diff
@@ -1308,8 +1311,13 @@ class CliArgs:
     )
     verbose: bool = clickdc.option(
```
scottnemes, line 1312: Should this be `int` now that it's being used as a count? Looks like it's set to int in the tests already.

rolandwalker: Yes!  Fixed.

scottnemes (approved):

PR #1857, opened 2026-04-21
[feat] Add option to prefetch completion metadata and to persist completion metadata when switching schemas (#109)

scottnemes:
## Description

Currently completion suggestions only populate / apply for the selected schema, and when you switch schemas, all completions are cleared. This PR does a few important things:

1. Adds an option to prefetch completions for schemas automatically; this can be a list of schemas, or all schemas (or none, the default).
2. Prefetching of completions is done in the background when the client is launched.
3. When the user manually changes schemas (`use` command), completions are fetched for that schemas as usual. However if the user switches schemas again, the completions for the previously selected schema are now saved (this is independent of prefetching). So if you "use db1", then "use db2", completions for both db1 and db2 will now be available in addition to any prefetched schemas.

## Checklist

    ```bash
    ```

### DIFF
--- a/mycli/schema_prefetcher.py
+++ b/mycli/schema_prefetcher.py
@@ -89,7 +89,8 @@ class SchemaPrefetcher:
     def _start(self, schemas: Iterable[str]) -> None:
         self.stop()
         current = self._current_schema()
-        queue = [s for s in schemas if s and s != current and s not in self._loaded]
+        existing = set(self.mycli.completer.dbmetadata.get('tables', {}).keys())
+        queue = [s for s in schemas if s and s != current and s not in self._loaded and s not in existing]
         if not queue:
             self._invalidate_app()
             return
--- a/mycli/schema_prefetcher.py
+++ b/mycli/schema_prefetcher.py
@@ -1,0 +1,231 @@
+"""Background prefetcher for multi-schema auto-completion.
+
+The default completion refresher only populates metadata for the
+currently-selected schema.  ``SchemaPrefetcher`` loads metadata for
+additional schemas on a background thread so that users can get
+qualified auto-completion suggestions (``OtherSchema.table``) without
+switching databases first.
+"""
+
+from __future__ import annotations
+
+import logging
+import threading
+from typing import TYPE_CHECKING, Any, Iterable
+
+from mycli.sqlexecute import SQLExecute
+
+if TYPE_CHECKING:  # pragma: no cover - typing only
+    from mycli.main import MyCli
+    from mycli.sqlcompleter import SQLCompleter
+
+_logger = logging.getLogger(__name__)
+
+ALL_SCHEMAS_SENTINEL = 'all'
+
+
+def parse_prefetch_setting(raw: str | None) -> list[str] | None:
+    """Parse the ``prefetch_schemas`` option value.
+
+    Returns ``None`` when the user wants every accessible schema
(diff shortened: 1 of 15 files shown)


test/myclirc
```diff
@@ -13,6 +13,14 @@ smart_completion = True
 # Suggestion: 3.
 min_completion_trigger = 1
 
+# Comma-separated list of schemas to prefetch for auto-completion.
+# Prefetching starts in the background after launch.
+# Examples:
+# (empty) = default, disables prefetching
+# schema1,schema2 = enables prefetch for given schemas
+# all = enables prefetch for all schemas
+prefetch_schemas =
```
rolandwalker, line 22: Why off by default?  Also, `all` is a valid schema name.

mycli/schema_prefetcher.py
```diff
@@ -0,0 +1,241 @@
...
+
+def parse_prefetch_config(mode: str, schema_list: str | list[str] | None) -> list[str] | None:
+    """Parse the ``prefetch_schemas_mode`` / ``prefetch_schemas_list`` options.
+
+    Returns ``None`` when every accessible schema should be prefetched
+    (``always``), an empty list when prefetching is disabled
+    (``never``), or the explicit list parsed from ``schema_list`` when
+    the mode is ``listed``.  Unknown modes fall back to ``always``.
+
+    ``schema_list`` may be a CSV string (single-value configobj case) or
```
rolandwalker: The config object gives us an `.as_list()` method which I think if applied in the right spot would always give a list for `schema_list`.

mycli/schema_prefetcher.py
```diff
@@ -0,0 +1,241 @@
...
+
+from mycli.sqlexecute import SQLExecute
+
+if TYPE_CHECKING:  # pragma: no cover - typing only
+    from mycli.main import MyCli
+    from mycli.sqlcompleter import SQLCompleter
+
+_logger = logging.getLogger(__name__)
+
+PREFETCH_MODE_ALWAYS = 'always'
```
rolandwalker: This might be more natural as an enum.

mycli/schema_prefetcher.py
```diff
@@ -0,0 +1,241 @@
...
+            sqlexecute.ssl,
+            sqlexecute.ssh_user,
+            sqlexecute.ssh_host,
+            sqlexecute.ssh_port,
+            sqlexecute.ssh_password,
+            sqlexecute.ssh_key_filename,
+        )
+
+    def _invalidate_app(self) -> None:
+        prompt_session = getattr(self.mycli, 'prompt_session', None)
```
rolandwalker, line 235: I think we could use prompt_toolkit's get_app() here, but this works.

rolandwalker (approved):
So nice!

mycli/schema_prefetcher.py
scottnemes: Updated!

mycli/schema_prefetcher.py
scottnemes: Updated!

mycli/schema_prefetcher.py
scottnemes: From what I found that won't work because the prefetcher is working on a background thread, and get_app() won't work across threads like that.

scottnemes:
#109

PR #1859, opened 2026-04-25
Respond to `help <term>` on builtin special commands

rolandwalker:
## Description
In this implementation, the detail returned is the same as visible in the table returned from a plain `help`, but the change opens up the ability to add a longer help text for each special command.

Motivation: there is not enough room in the tabular summary, and details such as the `-c` argument to `watch` are undocumented.  Some commands, such as `watch`, return longer help text when given with no argument. Others, such as `\dt`, perform a default action.  `help cmd` gives a uniform interface.

Bugs and limitations:

 * `help \G` is ambiguous, so if help is desired on `\G` itself, the `\G` must be quoted.
 * `help use` now gives help for the mycli special command `use`.  This should be a better choice than querying the server, since the special command is what is executed.

Example:

last image

## Checklist

    ```bash
    ```

### DIFF
--- a/mycli/packages/special/iocommands.py
+++ b/mycli/packages/special/iocommands.py
@@ -190,6 +190,9 @@ def editor_command(command: str) -> bool:
     Is this an external editor command?
     :param command: string
     """
+    # special case: allow help on the \edit command
+    if re.match(r'^([Hh][Ee][Ll][Pp])\s+(\\e|\\edit)\s*(;|\\G|\\g)?\s*$', command):
+        return False
     # It is possible to have `\e filename` or `SELECT * FROM \e`. So we check
     # for both conditions.
     return (
--- a/mycli/packages/special/main.py
+++ b/mycli/packages/special/main.py
@@ -22,6 +22,8 @@ from pymysql.cursors import Cursor
 logger = logging.getLogger(__name__)
 
 COMMANDS = {}
+CASE_SENSITIVE_COMMANDS = set()
+CASE_INSENSITIVE_COMMANDS = set()
 
 SpecialCommand = namedtuple(
     "SpecialCommand",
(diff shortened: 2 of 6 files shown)


mycli/packages/special/main.py
```diff
@@ -22,6 +22,8 @@
 logger = logging.getLogger(__name__)
 
 COMMANDS = {}
+CASE_SENSITIVE_COMMANDS = set()
```
scottnemes, line 25: Seems like this new case logic might be duplicating what already exists in `register_special_command`:

https://github.com/dbcli/mycli/blob/4782d2363b84f54675d31eaf0b7b8cc121e74024/mycli/packages/special/main.py#L102

mycli/packages/special/main.py
```diff
@@ -170,14 +178,20 @@ def show_help(*_args) -> list[SQLResult]:
     return [SQLResult(header=header, rows=result, postamble=f'Docs index — {DOCS_URL}')]
 
 
-def show_keyword_help(cur: Cursor, arg: str) -> list[SQLResult]:
+def _show_special_help(keyword: str) -> list[SQLResult]:
+    header = ['name', 'description', 'example']
+    description = '\n'.join(COMMANDS[keyword][2:4])
+    rows = [(keyword, description, '')]
```
scottnemes, line 184: Looks like the `example` column is always empty; is that for future use or should it be removed or populated?

mycli/packages/special/iocommands.py
```diff
@@ -190,6 +190,9 @@ def editor_command(command: str) -> bool:
     Is this an external editor command?
     :param command: string
     """
+    # special case: allow help on the \edit command
+    if re.match(r'^([Hh][Ee][Ll][Pp])\s+(\\e|\\edit)\s*(;|\\G|\\g)?\s*$', command):
```
scottnemes, line 194: Could use `re.IGNORECASE` for this if desired

mycli/packages/special/main.py
rolandwalker: Yes, this is leveraging the same information, to make it possible to immediately and accurately look up a command, respecting case-sensitivity, in one step.

mycli/packages/special/main.py
rolandwalker: The columns are made to match the output of `help` when returned from the server.  See `help SELECT;`.  Yes, I hope to populate the examples where appropriate.

mycli/packages/special/iocommands.py
rolandwalker: Unfortunately `\edit` is case-sensitive, though I don't know why it must be.

scottnemes (approved):

PR #1868, opened 2026-04-27
Extend `status` command output

rolandwalker:
## Description
 * recast `variables` as `global_variables`
 * add `session_variables` dict
 * tighten legacy-server decoding logic
 * derive charset rows from `session_variables` instead of a new query
 * add "Using delimiter" row, matching the vendor client
 * add "Using outfile" row (tee target), matching the vendor client
 * add SSL cipher row, matching the vendor client
 * add "Result characterset" row, an extension beyond the vendor client
 * add "Server timezone" row, an extension
 * add "Local timezone" row, an extension
 * align charset key names
 * make more None/null values the empty string, for consistency and to match the vendor client
 * add header so that `status\G` doesn't crash within `cli_helpers`
 * tweak some commentary

The vendor client also shows the setting for how binary data is rendered, which seems more difficult for mycli to include.

## Checklist

    ```bash
    ```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,11 @@
 Upcoming (TBD)
 ==============
 
+Features
+---------
+* Add more output to the `status` command.
+
+
 Documentation
 ---------
 * Give example for ANSI prompt colors in `~/.myclirc`.
--- a/mycli/packages/special/dbcommands.py
+++ b/mycli/packages/special/dbcommands.py
@@ -8,7 +8,13 @@ from pymysql.cursors import Cursor
 from mycli import __version__
 from mycli.packages.special import iocommands
 from mycli.packages.special.main import ArgType, special_command
-from mycli.packages.special.utils import format_uptime, get_ssl_version
+from mycli.packages.special.utils import (
+    format_uptime,
+    get_local_timezone,
+    get_server_timezone,
+    get_ssl_cipher,
+    get_ssl_version,
+)
 from mycli.packages.sqlresult import SQLResult
 
 logger = logging.getLogger(__name__)
@@ -69,7 +75,7 @@ def status(cur: Cursor, **_) -> list[SQLResult]:
     try:
         cur.execute(query)
     except ProgrammingError:
-        # Fallback in case query fail, as it does with Mysql 4
+        # Fallback in case query fails, as it does with Mysql 4
         query = "SHOW STATUS;"
         logger.debug(query)
         cur.execute(query)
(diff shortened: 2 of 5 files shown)


mycli/packages/special/dbcommands.py
```diff
@@ -135,23 +156,28 @@ def status(cur: Cursor, **_) -> list[SQLResult]:
...
-    cur.execute(query)
-    if one := cur.fetchone():
-        charset = one
-    else:
-        charset = ("", "", "", "")
-    output.append(("Server characterset:", charset[0]))
-    output.append(("Db characterset:", charset[1]))
-    output.append(("Client characterset:", charset[2]))
-    output.append(("Conn. characterset:", charset[3]))
+    charset_spec = [
```
scottnemes, line 159: Should we let the cli_helpers render format these instead of forcing a format?

mycli/packages/special/dbcommands.py
```diff
@@ -124,9 +138,16 @@ def status(cur: Cursor, **_) -> list[SQLResult]:
         pager = "stdout"
     output.append(("Current pager:", pager))
 
-    output.append(("Server version:", f'{variables["version"]} {variables["version_comment"]}'))
-    output.append(("Protocol version:", variables["protocol_version"]))
-    output.append(('SSL/TLS version:', get_ssl_version(cur)))
+    output.append(("Using delimiter:", iocommands.delimiter_command.current))
```
scottnemes: Could this use `get_current_delimiter()` from iocommands instead?

changelog.md
```diff
@@ -1,6 +1,11 @@
 Upcoming (TBD)
 ==============
 
+Features
+---------
+* Add more output to the `status` commnad.
```
scottnemes: commnad = command

mycli/packages/special/dbcommands.py
rolandwalker: Better!

changelog.md
rolandwalker: oof, thanks

mycli/packages/special/dbcommands.py
rolandwalker: I don't quite understand.  It _does_ get formatted into a table by cli_helpers.

mycli/packages/special/dbcommands.py
scottnemes: Original suggestion was to use `iocommands.get_current_delimiter()` possibly instead of `iocommands.delimiter_command.current` since that function exists to get the value instead of directly accessing the property

mycli/packages/special/dbcommands.py
scottnemes: I mean the spacing like `Db     characterset:`, adding the spaces in between. But if that is the only way to get the alignment you want then that works.

scottnemes (approved):

mycli/packages/special/dbcommands.py
rolandwalker: Yes!  I made the change.

scottnemes: @rolandwalker Hah sorry, was looking at my own comment thinking it would be updated for some reason. Still too early I guess :p

PR #1876, opened 2026-05-01
Append dot after database name when completing on a table

rolandwalker:
## Description
Previously, database names were available as candidates when completing a table name, but if completing a database name, the user had to manually enter a trailing dot before typing a table name.

It might be even nicer if the dot was appended after the completion, but did not appear in the candidate menu.  It isn't clear how to do that. On the other hand, the dot visually distinguishes which candidates are of which type.

The trailing dot does not appear when a database is completed in a database-only position such as `USE`.

There is an xfailed test included, since `USE <database>` is not working correctly under the test suite, but there is related coverage.

Instead of checking the entire `Document`, we look at only the relevant statement using `sqlparse.split()`, and that fix is applied in a few other places.  Otherwise a separate statement can influence completion candidates.

The above split logic should also apply to `document.text_before_cursor`, which is including multiple statements.  However, that is beyond the scope of this change.

Also, we converge on using `"type":"database"` instead of `"type":"schema"` in the completion engine.  This somewhat clears up the confusion between "schema" and "database", and more importantly allows database names with a trailing dot to be suggested as completions in `DROP TABLE`.

## Checklist

    ```bash
    ```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,11 @@
+Upcoming (TBD)
+==============
+
+Features
+---------
+* Append dot to database name when completing on a table name.
+
+
 1.71.0 (2026/05/01)
 ==============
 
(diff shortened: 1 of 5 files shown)


rolandwalker:
This didn't work smoothly unless `min_completion_trigger` was set, and we didn't reach an agreement about how it should work.

PR #1889, opened 2026-05-14
Adapt test suite to pygments 2.20

terjeros:
## Checklist

    ```bash
    ```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,7 @@
 Bug Fixes
 ---------
 * Update `sqlglot` to v30.7.0 to fix has_bit_strings error.
+* Adapt test suite to pygments 2.20.0
 
 
 1.72.0 (2026/05/08)
--- a/test/pytests/test_naive_completion.py
+++ b/test/pytests/test_naive_completion.py
@@ -4,6 +4,8 @@ from prompt_toolkit.completion import Completion
 from prompt_toolkit.document import Document
 import pytest
 
+from test.utils import pygments_at_least
+
 
 @pytest.fixture
 def completer():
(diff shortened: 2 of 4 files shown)


rolandwalker:
Hi!  Thanks!  What would be the disadvantage to upgrading Pygments in `pyproject.toml`, and making the test changes unconditional?

terjeros:
That's related to an other issue I want to discuss, lately mycli has to much stricter pinning of module versions.  This makes testing simpler and might increase quality of package for some users.

However, it creates real problems for Linux distributions and others (like Homebrew for macOS)  that want to  ship mycli together other software, strict pinning of large set of modules often causes conflicts. 

For example will Fedora 45 (coming in Oct 2026) be using pygments 2.20.0, while Fedora 43 and 44 will stay on pygments 2.19.x. Each Fedora release have 18 months of support so it's not that of a problem, however consider Ubuntu 26.04 LTS which  has 10 years of support.

In short, being less strict in pinning will mean newer releases of mycli is more likely to used and attract more users in general.

rolandwalker:
Thank you for surfacing that larger issue.  Could you start a separate Issue or Discussion on pinning in general?

terjeros:
Yeah, I will create a dedicated issue for this

rolandwalker:
Since two of the tests modified in this PR are testing the order of the completions as well as the content, checking `sorted()` vs `sorted()` isn't ideal — now the order check is gone.  Instead, we should lay out the full expected order for Pygments 2.20+, including the new keywords, and have the alternative path _delete_ the new keywords from the expected list when the Pygments version is 2.19 or below.

Would you like to do that, or should I handle?

terjerosten:
Please handle this

rolandwalker (approved):

rolandwalker:
@terjeros the version check is tidied up in #1891, with the same core functionality.  I look forward to a separate discussion on pinning, since it seems like you have some knowledge we lack regarding packaging.

PR #1893, opened 2026-05-16
Migrate state, cli_args, and output out of main.py

rolandwalker:
## Description
Create a `mycli/app_state.py`, `mycli/cli_args.py`, and `mycli/output.py`, factoring logic out of `mycli/main.py`, with no functional change.

Like other refactors, this could be seen as incomplete in various ways, but it moves us forward on reducing LOC in `main.py` (which is now a third of the size from before refactoring started).

## Checklist

    ```bash
    ```

### DIFF
--- a/mycli/output.py
+++ b/mycli/output.py
@@ -1,0 +1,291 @@
+from __future__ import annotations
+
+from datetime import datetime
+from decimal import Decimal
+from io import TextIOWrapper
+import itertools
+import os
+import shutil
+from typing import Any, Generator, Literal, Protocol
+
+from cli_helpers.tabular_output import TabularOutputFormatter, preprocessors
+from cli_helpers.tabular_output.output_formatter import MISSING_VALUE as DEFAULT_MISSING_VALUE
+from cli_helpers.utils import strip_ansi
+import click
+from configobj import ConfigObj
+import prompt_toolkit
+from prompt_toolkit.formatted_text import (
+    ANSI,
+    HTML,
+    AnyFormattedText,
+    FormattedText,
+    to_formatted_text,
+    to_plain_text,
+)
+from prompt_toolkit.shortcuts import PromptSession
+from prompt_toolkit.styles.style import _MergedStyle
+from pygments.style import Style as PygmentsStyle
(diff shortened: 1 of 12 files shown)


mycli/output.py
```diff
@@ -0,0 +1,264 @@
...
+from mycli.packages.tabular_output import sql_format
+from mycli.sqlexecute import FIELD_TYPES
+
+
+class OutputMixin:
+    prompt_lines: int
+
+    if TYPE_CHECKING:
+
+        def __getattr__(self, name: str) -> Any: ...
```
scottnemes: This seems like it short circuits the mypy checks; was it done to get past errors that were happening? Ran this through Claude and it recommends something like the below instead to prevent future errors not being caught by mypy:

```
if TYPE_CHECKING:
    from cli_helpers.tabular_output import TabularOutputFormatter
    from configobj import ConfigObj

    class _MyCliState(Protocol):
        # External attributes supplied by the concrete class (MyCli).
        config: ConfigObj
        my_cnf: ConfigObj
        logfile: TextIOWrapper | None
        prompt_session: Any
        prompt_format: str
        prompt_lines: int
        explicit_pager: bool
        ptoolkit_style: Any
        helpers_style: Any
        helpers_warnings_style: Any
        main_formatter: TabularOutputFormatter
        redirect_formatter: TabularOutputFormatter

        # Provided by AppStateMixin.
        def read_my_cnf(self, cnf: ConfigObj, keys: list[str]) -> dict[str, Any]: ...

        # Provided by OutputMixin itself; declared so cross-method calls type-check.
        def log_output(self, output: str | AnyFormattedText) -> None: ...
        def get_output_margin(self, status: str | None = None) -> int: ...
        def get_reserved_space(self) -> int: ...


class OutputMixin:
    def output_timing(self: _MyCliState, timing: str, is_warnings_style: bool = False) -> None:
...
```

Will approve still in case you do not think that is an issue however.

scottnemes (approved):

mycli/output.py
rolandwalker: Thanks!  Will try it.

PR #1894, opened 2026-05-18
Refactor `main.py` to minimal content

rolandwalker:
## Description

Creates the following:

 * `mycli/cli_runner.py`
 * `mycli/client.py`
 * `mycli/client_commands.py`
 * `mycli/client_connection.py`
 * `mycli/client_query.py`

and reduces `main.py` to only:

 * `click_entrypoint()`
 * `main()`

As first PR'd, this maintains the current testing surface.  Best would be to reach the goal of individual tests over each of the above files, and outright deletion of `test_main_regression.py`.  Edit: keeping the structure as-is for this PR, and looking to followups to resolve.

Also, as first PR'd, `main.py` contains almost nothing.  It would be logical to move the content of `cli_args.py` back in.  Edit: keeping the structure as-is for this PR, and looking to followups to resolve.

While the files are smaller and the code is more approachable, like previous refactors, this is not perfect.  For example, `cli_runner.py` has extensive logic regarding environment variables and command-line arguments.  Whereas, ideally, all configuration would be collected together in a distinct layer.

## Checklist

    ```bash
    ```


test/pytests/test_main.py
```diff
@@ -2437,7 +2438,7 @@ def test_connect_reports_expired_password_login_error(monkeypatch: pytest.Monkey
     cli.logger = cast(Any, DummyLogger())
     echo_calls: list[str] = []
     cli.echo = lambda message, **kwargs: echo_calls.append(str(message))  # type: ignore[assignment]
-    monkeypatch.setattr(main, 'WIN', False)
+    monkeypatch.setattr(mycli.compat, 'WIN', False)
```
scottnemes: This should be: `monkeypatch.setattr(mycli.client_connection, 'WIN', True)` I believe, like in test_main_regression.py

test/pytests/test_main_regression.py
```diff
@@ -476,7 +477,7 @@ def test_connect_covers_defaults_keyring_prompt_retries_and_errors(monkeypatch:
     cli.echo = lambda *args, **kwargs: echo_calls.append((args, kwargs))  # type: ignore[assignment]
     logger = DummyLogger()
     cli.logger = cast(Any, logger)
-    monkeypatch.setattr(main, 'WIN', True)
+    monkeypatch.setattr(mycli.client_connection, 'WIN', True)
```
scottnemes, line 480: This is the correct one I believe; all the rest should match (this file, and test_main.py)

scottnemes (approved):

mycli/client_commands.py
```diff
@@ -0,0 +1,213 @@
...
+    def change_db(self, arg: str, **_) -> Generator[SQLResult, None, None]:
+        if arg.startswith("`") and arg.endswith("`"):
+            arg = re.sub(r"^`(.*)`$", r"\1", arg)
+            arg = re.sub(r"``", r"`", arg)
+
+        if not arg:
+            click.secho("No database selected", err=True, fg="red")
+            return
+
+        from mycli import main as main_module
```
scottnemes, line 129: Dupe import of below with comment

mycli/client_commands.py
```diff
@@ -0,0 +1,213 @@
...
+        if not arg:
+            message = "Missing required argument: filename."
+            return [SQLResult(status=message)]
+        try:
+            with open(os.path.expanduser(arg)) as f:
+                query = f.read()
+        except IOError as e:
+            return [SQLResult(status=str(e))]
+
+        from mycli import main as main_module
```
scottnemes, line 153: Dupe import of below

mycli/cli_runner.py
```diff
@@ -0,0 +1,422 @@
...
+        auto_vertical_output=cli_args.auto_vertical_output,
+        warn=cli_args.warn,
+        myclirc=cli_args.myclirc,
+        show_warnings=cli_args.show_warnings,
+        cli_verbosity=cli_verbosity,
+    )
+
+    if cli_args.checkup:
+        main_module.main_checkup(mycli)
+        main_module.sys.exit(0)
```
scottnemes: For all of these sys commands, is there a reason you're accessing that through the module instead of just importing sys at the top and using that as normal?

rolandwalker: No good reason!  I (and the agent) kept the tests working against main, which was awkward.  This and other weird things got through due to that constraint/pattern.

The goal is to not pull anything from main; the question is whether to keep this PR open while all of that incremental work is done.

rolandwalker:
Even though there are some obvious followups wanted, I've convinced myself to merge this as-is (after accounting for review comments).  It may be valuable to have the steps of refactoring in the commit history.

PR #1913, opened 2026-06-04
Fix highlighting for set-prefixed identifiers

### LINKED ISSUE #1872: When the table name starts with "set" (e.g. "settings"), the "set" prefix is highlighted as if it was a reserved keyword.
Image

### Steps to Reproduce

1. Type `SELECT * FROM settings_123 WHERE id = 123;`.

### System

 * mycli version: 1.70.0
 * OS/version: macOS

### Discussion

AI says

```
Cause: mycli uses Pygments MySqlLexer, and that lexer has a special rule:

(r'\b(set)(?!\s*\()', Keyword),

It matches set at a word start, but does not require a word end, so in settings_123 it tokenizes as:

- set -> Keyword (green)
- tings_123 -> Name

I verified this directly by tokenizing SELECT * FROM settings_123 with MySqlLexer, and it produces exactly that split.
```

(I'm sorry if I should open this issue in the Pygments repo instead.)

----

By the way, I'm not affected by this issue so much; I almost don't care. So it won't hurt me if this issue is closed as not planned :)

puneetdixit200:
## Description

Fixes #1872.

`MyCliLexer` now recognizes identifiers that start with `set` before falling through to Pygments' MySQL lexer, so table names such as `settings_123` stay highlighted as a single name instead of splitting `set` as a keyword. The standalone `SET` keyword still falls through to the inherited lexer.

I added a regression test, updated the changelog, and added my name to the authors list.

Tests run:

- `PYTHONPATH=. UV_CACHE_DIR=/tmp/mycli-1872-uv-cache uv run --extra dev pytest test/pytests/test_lexer.py -q` (failed before the lexer fix, passed after)
- `PYTHONPATH=. UV_CACHE_DIR=/tmp/mycli-1872-uv-cache uv run --extra dev pytest test/pytests/test_lexer.py test/pytests/test_naive_completion.py test/pytests/test_clistyle.py -q`
- `PYTHONPATH=. UV_CACHE_DIR=/tmp/mycli-1872-uv-cache uv run --extra dev ruff check`
- `PYTHONPATH=. UV_CACHE_DIR=/tmp/mycli-1872-uv-cache uv run --extra dev ruff format --check`
- `PYTHONPATH=. UV_CACHE_DIR=/tmp/mycli-1872-uv-cache uv run --extra dev mypy mycli/lexer.py test/pytests/test_lexer.py`
- `git diff --check`

I used AI assistance while preparing this change and reviewed/tested the result locally.

## Checklist

    ```bash
    uv run ruff check && uv run ruff format
    ```


mycli/lexer.py
```diff
@@ -1,11 +1,16 @@
...
+from pygments.token import Keyword, Name
 
 
 class MyCliLexer(MySqlLexer):
     """Extends MySQL lexer to add keywords."""
 
     tokens = {
-        "root": [(r"\brepair\b", Keyword), (r"\boffset\b", Keyword), inherit],
+        "root": [
+            (r"\bset[\w$]+\b", Name),
```
rolandwalker, line 12: Since this was fixed upstream, can we add a TODO comment on the `set` line explaining that it won't be needed after Pygments is upgraded above v2.20.0 ?

changelog.md
```diff
@@ -14,6 +9,7 @@ Internal
 * Add CI on macOS.
 * Add limited CI on Windows.
 * Add limited CI on Windows WSL.
+* Keep identifiers that start with `set` highlighted as names.
```
rolandwalker: This seems like it should be in a Bug Fixes section rather than Internal, since it is visible to the user.

rolandwalker:
Thanks for the updates!  There seem to be some unrelated changes added.

rolandwalker:
Thank you!

PR #1916, opened 2026-06-06
fix: display version for Homebrew MySQL

Dragonliu2018:
## Description

  MySQL installed via Homebrew (and some other distributions) reports a plain `X.Y.Z` version string during the handshake with no trailing suffix — for example `9.6.0` instead of the more common `9.6.0-log` or `8.0.32-debian`.

  `ServerInfo.from_version_string()` uses a series of regex patterns to detect the server species and extract the version number. All existing patterns  require a `-` separator, so a plain `X.Y.Z` string falls through to the  `else` branch, which previously hardcoded `parsed_version = ""`.

 As a result, the welcome banner displayed `MySQL ` with a trailing space and  no version number.

  **Fix**: in the `else` branch, attempt to extract the version with a simple`X.Y.Z` regex before falling back to an empty string. The species remains `MySQL` — only the displayed version number changes.

  **Before:**
  MySQL
  mycli 1.73.0

  **After:**
  MySQL 9.6.0
  mycli 1.73.0

## Checklist

    ```bash
    ```

### DIFF
--- a/mycli/sqlexecute.py
+++ b/mycli/sqlexecute.py
@@ -68,7 +68,8 @@ class ServerInfo:
             (r"(?P<version>[0-9\.]+)-MariaDB", ServerSpecies.MariaDB),
             (r"[0-9\.]*-TiDB-v(?P<version>[0-9\.]+)-?(?P<comment>[a-z0-9\-]*)", ServerSpecies.TiDB),
             (r"(?P<version>[0-9\.]+)[a-z0-9]*-(?P<comment>[0-9]+$)", ServerSpecies.Percona),
-            (r"(?P<version>[0-9\.]+)[a-z0-9]*-(?P<comment>[A-Za-z0-9_]+)", ServerSpecies.MySQL),
+            # Also matches plain "X.Y.Z" with no suffix (e.g. Homebrew MySQL).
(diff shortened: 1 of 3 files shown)


mycli/sqlexecute.py
```diff
@@ -77,8 +77,11 @@ def from_version_string(cls, version_string: str) -> ServerInfo:
                 detected_species = species
                 break
         else:
+            # Some distributions (e.g. Homebrew MySQL) return a plain "X.Y.Z"
+            # version string with no suffix, so none of the patterns above match.
             detected_species = ServerSpecies.MySQL
-            parsed_version = ""
+            plain_match = re.match(r"^([0-9]+\.[0-9]+\.[0-9]+)", version_string)
```
rolandwalker: Hi!  Can't this match be done above in `re_species` with `^` and `$` anchors?

Dragonliu2018: Good idea! Instead of keeping a separate match in the else branch, I merged it into re_species by making the suffix optional:
```python
(r"(?P<version>[0-9]+\.[0-9]+\.[0-9]+)[a-z0-9]*(-(?P<comment>[A-Za-z0-9_]+))?", ServerSpecies.MySQL),
```
This handles both X.Y.Z-suffix and plain X.Y.Z in a single rule. Thanks!

rolandwalker (approved):

rolandwalker:
Super appreciated!

PR #1917, opened 2026-06-06
feat: support doris server specie

Dragonliu2018:
## Description

[Apache Doris](https://doris.apache.org/) implements the MySQL wire protocol and hard-codes `5.7.99` as its handshake version string, making it indistinguishable from MySQL during connection. mycli previously identified Doris as `MySQL` with a blank version.

**Fix**: after every connection classified as `MySQL`, run a lightweight probe:

```sql
SELECT @@version_comment, @@version
```

If `"doris"` appears in either field (e.g. `'Doris version doris-2.1.7-rc01'`), extract the real version number and update `server_info` to`ServerInfo(ServerSpecies.Doris, "2.1.7")`.

**Before / After:**

```
# Before
MySQL 
MySQL root@doris-host:(none)>

# After
Doris 2.1.7
Doris root@doris-host:(none)>
```

## Checklist

    ```bash
    ```

### DIFF
--- a/mycli/sqlexecute.py
+++ b/mycli/sqlexecute.py
@@ -39,6 +39,7 @@ class ServerSpecies(enum.Enum):
     MariaDB = "MariaDB"
     Percona = "Percona"
     TiDB = "TiDB"
+    Doris = "Doris"
     Unknown = "Unknown"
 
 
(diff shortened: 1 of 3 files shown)


mycli/sqlexecute.py
```diff
@@ -366,6 +367,38 @@ def connect(
         if not self.sandbox_mode:
             self.reset_connection_id()
             self.server_info = ServerInfo.from_version_string(conn.server_version)  # type: ignore[attr-defined]
+            self._detect_doris()
+
+    def _detect_doris(self) -> None:
```
rolandwalker: It would be preferable if this was cast as a utility method which returned a value rather than setting `self.server_info` as a side effect.  What do you think?

rolandwalker:
Hi!  Thanks for the contributions.  You can ignore the Codex failures.  I need to update an API key.

In this PR, we could also update the list of supported servers in the README.

mycli/sqlexecute.py
Dragonliu2018: Good point! Refactored it into `_probe_doris_version()` which returns str | None (the Doris version string, or None if not Doris). The caller now handles constructing the ServerInfo. Thanks!

rolandwalker (approved):

Dragonliu2018:
> Hi! Thanks for the contributions. You can ignore the Codex failures. I need to update an API key.
> 
> In this PR, we could also update the list of supported servers in the README.

Done! Updated README.md to include Apache Doris in the list of supported servers.

rolandwalker:
Perfect!

PR #1918, opened 2026-06-06
perf: skip LLM imports when MYCLI_LLM_OFF is set

Dragonliu2018:
## Description
`special/__init__.py` unconditionally imported LLM symbols on every startup, pulling in the entire `openai` package even when LLM features are unused. `llm.py` and `main.py` already guarded their imports with `MYCLI_LLM_OFF`, but `__init__.py` bypassed that protection.

This PR brings `__init__.py` in line with the rest: when `MYCLI_LLM_OFF=1`, LLM imports are skipped and lightweight stubs are provided instead, reducing startup time.

## Checklist

    ```bash
    ```

### DIFF
--- a/mycli/packages/special/__init__.py
+++ b/mycli/packages/special/__init__.py
@@ -1,3 +1,5 @@
+import os
+
 from mycli.packages.special.dbcommands import (
     list_databases,
     list_tables,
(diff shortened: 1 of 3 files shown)


mycli/packages/special/__init__.py
```diff
@@ -1,3 +1,5 @@
+import os as _os
```
rolandwalker: Why is `as _os` needed?

Dragonliu2018: It's unnecessary. Simplified to import os. Thanks!

rolandwalker (approved):

rolandwalker:
Great catch, thank you!

PR #1943, opened 2026-06-19
Spinners for setup steps with `--progress` `--batch`

rolandwalker:
## Description

For large `--batch` scripts (the very kind of SQL scripts for which the progress bar is useful) the initial steps of pre-parsing the batch and optionally replaying the checkpoint log can take substantial time. Therefore, add text and spinners showing the stage and progress of the setup steps.

A lightweight new dependency is added due to limitations in the spinners from `prompt-toolkit`: the `prompt-toolkit` spinners must have a known upper bound to which to iterate, and they leave behind random "mid-spin" characters on completion like

    pre-parsing batch \

which creates visual confusion about whether the step completed successfully.

`yaspin` spinners can integrate with `prompt-toolkit` more deeply than is done here, and could also for instance be used during long interactive queries (though that particular usecase may have little value).

Another option would be spinners from the `rich` library, which we may try in the future, if adopting `rich` for tabular output.

Motivation: the pause before showing the progress bar was recently opaque to me when running a large `--batch` script of ~70M statements.  It took several seconds to parse the script.

Examples:

last image

Screenshot 2026-06-19 at 3 06 53 PM

## Checklist

    ```bash
    ```

### DIFF
--- a/mycli/main_modes/batch.py
+++ b/mycli/main_modes/batch.py
@@ -10,6 +10,7 @@ import prompt_toolkit
 from prompt_toolkit.shortcuts import ProgressBar
 from prompt_toolkit.shortcuts.progress_bar import formatters as progress_bar_formatters
 import pymysql
+from yaspin import yaspin
 
 from mycli.packages.batch_utils import statements_from_filehandle
 from mycli.packages.interactive_utils import confirm_destructive_query
@@ -28,6 +29,7 @@ def replay_checkpoint_file(
     batch_path: str,
     checkpoint_path: str | None,
     resume: bool,
+    progress: bool = False,
 ) -> int:
     if not resume:
         return 0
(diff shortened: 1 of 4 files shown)


mycli/main_modes/batch.py
```diff
@@ -48,16 +50,27 @@ def replay_checkpoint_file(
                 batch_gen = statements_from_filehandle(batch_h)
             except ValueError as e:
                 raise CheckpointReplayError(f'Error reading --batch file: {batch_path}: {e}') from None
+            if progress:
+                spinner = yaspin(text='replaying checkpoint', side='right', stream=sys.stderr)
+                spinner.start()
```
scottnemes: This instance of spinner doesn't use `with` like the other, so if the outer try/except errors out after the spinner starts it won't be stopped

rolandwalker: Thanks, good catch!

PR #1951, opened 2026-06-20
Make batch execution noninteractive by default

rolandwalker:
## Description
* replace `--noninteractive` flag with `--warn-batch`, with an inverted meaning
 * when `--warn-batch` is set, it overrides myclirc's `destructive_warning` setting
 * let `--execute` work exactly like `--batch`
 * send destructive-query confirmation prompts to the standard error, since users running batch scripts will often have the standard output redirected

This is a breaking change, not least because the `--noninteractive` flag no longer has an effect, and disappears from the docs.  But passing it causes a Click warning referring the user to `--warn-batch`.

Preparation for release 2.0.

## Checklist

    ```bash
    ```

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -201,6 +201,10 @@ class CliArgs:
         clickdc=None,
         help='Warn before running a destructive query.',
     )
+    warn_batch: bool = clickdc.option(
+        is_flag=True,
+        help='Warn before running a destructive query when executing a script.',
+    )
     local_infile: bool | None = clickdc.option(
         type=bool,
         is_flag=False,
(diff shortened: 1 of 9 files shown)


mycli/main.py
```diff
@@ -209,6 +209,10 @@ class CliArgs:
         clickdc=None,
         help='Warn before running a destructive query.',
     )
+    batch_warn: bool = clickdc.option(
```
scottnemes: Looks like `destructive_warning` is set to True by default, but the new `batch_warn` is False by default (default Click value for bool). Might be worth calling that out in the changelog, or setting batch_warn to True by default. Could see that one getting some people unexpectedly

rolandwalker: The intent of the PR is for batch mode to be non-interactive by default.  Added to the changelog, updating the name of the flag as well.

scottnemes (approved):

PR #1962, opened 2026-06-23
Ensure `--batch` and `--logfile` are distinct files

rolandwalker:
## Description
Motivation: to avoid some kind of infinite loop, in which mycli reads from a file which is constantly being written to.

This is much more succinct than #1931 , which changed the argument type from a file to a string.  In retrospect that change was a mistake.

## Checklist

    ```bash
    ```

### DIFF
--- a/mycli/main.py
+++ b/mycli/main.py
@@ -333,6 +333,17 @@ def preprocess_cli_args(
             click.secho('Error: --batch and --checkpoint must be different files.', err=True, fg='red')
             sys.exit(1)
 
+    if (
+        cli_args.logfile
+        and os.path.exists(cli_args.logfile.name)
+        and cli_args.batch
+        and cli_args.batch != '-'
+        and os.path.exists(cli_args.batch)
+    ):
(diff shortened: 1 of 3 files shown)


mycli/main.py
```diff
@@ -327,6 +327,17 @@ def preprocess_cli_args(
...
             sys.exit(1)
 
+    if (
+        cli_args.logfile
+        and os.path.exists(cli_args.logfile.name)
+        and cli_args.batch
+        and cli_args.batch != '-'
+        and os.path.exists(cli_args.batch)
+    ):
+        if os.stat(cli_args.batch) == os.stat(cli_args.logfile.name):
```
scottnemes: Recommended way of checking if two files are the same file is `os.path.samefile`. Looks like we use os.stat directly elsewhere so it matches, but would be worth at least using it for new stuff

scottnemes (approved):
Left one nit, but fine to merge either way

mycli/main.py
rolandwalker: Thanks!

PR #1963, opened 2026-06-23
Migrate connection-related settings from myclirc `[main]`

rolandwalker:
## Description
 * migrate `main.default_character_set` to the `[connection]` section if present
 * migrate `main.ssl_mode` to `connection.default_ssl_mode` if present

These migrations will affect a modest number of users who installed mycli for the first time during the periods in which the given connection-level settings were in `[main]` in the default myclirc file.

For `default_ssl_mode`, that is fresh installs between 2026-01-02 and 2026-02-09.  For `default_character_set`, that is fresh installs between 2026-01-22 and 2026-02-02.

Users who ran `mycli --checkup` since then were already advised to migrate the settings to the `[connection]` section.

The `--checkup` instructions for migrating the settings are not removed, though they are somewhat duplicative, to keep the checkup instructions more coherent.

Writes to the config file should probably be used sparingly, since some comments and formatting in the user's file can be lost, but this already happens if a favorite query is saved.

Motivation: general simplification; bundling all breaking changes into the 2.0 release.  The logic for determining whether the user or package configuration controls is too subtle when there are multiple possible locations for the setting.

Preparation for release 2.0

## Checklist

    ```bash
    ```

### DIFF
--- a/mycli/app_state.py
+++ b/mycli/app_state.py
@@ -11,11 +11,39 @@ if TYPE_CHECKING:
     from mycli.client import MyCli
 
 
-def normalize_ssl_mode(config: ConfigObj) -> tuple[str | None, str | None]:
-    ssl_mode = config['main'].get('ssl_mode', None) or config['connection'].get('default_ssl_mode', None)
+def normalize_ssl_mode(
+    config: ConfigObj,
+    config_without_package_defaults: ConfigObj,
+) -> tuple[str | None, str | None]:
+    error_notice: str | None = None
+    ssl_mode: str | None = None
+
+    if 'main' in config_without_package_defaults and 'ssl_mode' in config_without_package_defaults['main']:
+        # migration with notice added with mycli 2.0.0 in 2026-07
+        # todo: entirely remove support for ssl_mode in [main]
+        error_notice = (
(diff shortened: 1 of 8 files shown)


mycli/app_state.py
```diff
@@ -11,11 +11,39 @@
...
+            config_without_package_defaults['connection'] = {}
+        if config_without_package_defaults['connection'].get('default_ssl_mode', None) in (None, ''):
+            config_without_package_defaults['connection']['default_ssl_mode'] = ssl_mode
+        else:
+            ssl_mode = config_without_package_defaults['connection'].get('default_ssl_mode')
+            error_notice += f'\nBut connection.default_ssl_mode already existed, with the value: "{ssl_mode}".'
+        del config_without_package_defaults['main']['ssl_mode']
+        config_without_package_defaults.write()
+
+    if ssl_mode is None and 'default_ssl_mode' in config['connection']:
```
scottnemes: Think this should be `if not ssl_mode`, or more checks needed. Previously it used `.get` for the ssl_mode so it would be None if not set/empty, but with the new code it could end up being an empty string which would pass the `ssl_mode is None` check.

scottnemes (approved):
Left one comment. Good to go after that or if you decide not to change it

mycli/app_state.py
rolandwalker: Changed!

PR #2031, opened 2026-07-21
Dataframe transforms and Parquet save operators

rolandwalker:
## Description

Add a trailing operator `.|` which allows applying arbitrary dataframe transforms with Polars.  This is similar to the `$|` operator, except that normal tabular output is maintained, so long as the Polars expression also returns a `DataFrame` or `Series`.

Add a trailing operator `.>` which saves output to a Parquet file.  This is similar to the `$>` operator for shell redirection.

Together these features are intended to make mycli more friendly to Data Science practitioners, lowering the boundaries between the CLI/REPL and notebook modes of working.

In many cases, dataframe transformations may be achievable in plain SQL, but for many practitioners, dataframes are the preferred approach.  In other cases, there are computations which are achievable in dataframes but not in SQL.  Medians are a simple example.

There are performance implications of using the dataframe transform: the entire SQL result must be fetched and loaded into memory.

Counter to the familiarity argument above, the `.|` operator supplies a Polars dataframe rather than a Pandas dataframe.

While Pandas is the industry standard, Polars is growing, and should be familiar enough to be a workable alternative.  Polars is sometimes a bit faster, and usually far more memory-efficient.  But the reason it is chosen here over Pandas is that the dependency chain for mycli is simpler.

Polars is also configured as an optional dependency in `pyproject.toml`.

There is more extensive documentation of the functionality in the PR.

Simple example showing a transform which returns a `DataFrame`, which is then rendered as usual:

Screenshot 2026-07-21 at 6 00 22 AM

## Checklist

    ```bash
    ```


mycli/packages/polars_transform.py
```diff
@@ -0,0 +1,220 @@
...
+        polars=_load_polars(),
+        altair=_load_altair() if expression is not None else None,
+    )
+
+
+def _load_polars() -> Any:
+    try:
+        import polars as pl
+    except ImportError as exc:
+        raise PolarsTransformError("Polars transforms require Polars to be installed.'") from exc
```
scottnemes: Extra single quote at the end here and same thing just below

scottnemes (approved):
Looks good! Left one nit comment for string fixes

PR #2085, opened 2026-08-03
Let favorite queries use Jinja2 templates

### LINKED ISSUE #1871: How about Favorite queries `Positional Parameters` add Default values support
> Favorite queries support shell-style parameter substitution. Save your favorite query with parameters as placeholders (e.g. $1, $2, $3, etc.):

Maybe looks like
```
\fs user_by_name select * from users where name = '${1:-john}'
```

<br>

`\f user_by_name`
call without args use the default value john

rolandwalker:
## Description
Let favorite queries use Jinja2 templates, allowing optional arguments and control blocks in favorite queries.

Include completions for Jinja2 key names.  Supporting the completions is quite a bit more code than the core functionality!  Completions also can only be offered if the key name is referred to explicitly in the template (not merely iterated over, which is technically possible).

Extensively update the config file and internal helpdoc with examples and notes.

It seems arbitrary to choose whether Jinja2 templates are processed before or after the legacy positional arguments are substituted.  I chose Jinja2 before positional.  The agent hacked together a UUID-based solution for this, which may not be pretty.

There is one regression: whereas a positional argument could previously freely begin with the dash character, that can now be ambiguous with the `--key=value` syntax.  The solution is to disambiguate by putting such positional values after a `--`:

    /f query --key=value -- --ambiguous-positional-value--

This should be rare enough that the regression is worth the new feature.

This adds a Jinja2 direct dependency to the project.  It was already a transitive dependency via altair if mycli was installed with dataframe support.

Fixes #1871 

~Edit: it looks like there is a bug here in that we determine that a key is unused unless it is referred to explicitly.  That logic should apply to completions only.  Will fix, but it is a real edge case, and this PR is still reviewable.~ Fixed.

## Checklist

    ```bash
    ```


mycli/packages/special/iocommands.py
```diff
@@ -351,17 +361,30 @@ def execute_favorite_query(cur: Cursor, arg: str, **_) -> Generator[SQLResult, N
...
+        query, positional_values, arg_error = prepare_favorite_query_args(query, args)
         if query is None:
             yield SQLResult(status=arg_error)
         else:
+            try:
+                query = render_favorite_query(query, template_values)
+            except FavoriteQueryArgumentError as exc:
+                yield SQLResult(status=f'Invalid favorite query arguments: {exc}')
+                return
+            except Exception as exc:
```
scottnemes, line 388: This could be a more specific jinja TemplateError, then a following more generic one to catch everything else

scottnemes (approved):
Added one comment, looks good otherwise

mycli/packages/special/iocommands.py
rolandwalker: Good idea.

Issue #2102, opened 2026-08-06
killing query kills ssh connection instead


mentalstring:
When I use a `ssh_jump` (great feature!), stopping a running query seems to kill the ssh tunnel instead.

I connect to a remote server using a DSN along these lines:

```
mysql://user@localhost/database?socket=%2Fvar%2Frun%2Fmysqld%2Fmysqld.sock&ssh_jump=user@db-server
```

This connects and works fine. However, if I happen to abort a long running query with ^C, at that point it _immediately_ says:

```
Encountered error while cancelling query: (2003, "Can't connect to MySQL server on 'localhost' ([Errno 61] Connection refused)")
```

If I try to issue other commands, it just keeps saying:
```
Reconnecting...
Creating new connection...
(2003, "Can't connect to MySQL server on 'localhost' ([Errno 61] Connection refused)")
```

Now for the odd part: if I restart mycli to the same server, the original long running query is still running. So the ^C never killed it and, somehow, it seems to have killed the ssh tunnel instead.

### System

 * mycli version: 2.9.0
 * OS/version: macOS 26.6 (+homebrew)

rolandwalker:
Great report, thanks!

We recently removed the original SSH tunnel features and then reimplemented as `--ssh-jump` in https://github.com/dbcli/mycli/pull/1986 .  It would be great for the new `--ssh-jump` to be more battle-tested.

I can't duplicate the issue here, but do have an immediate idea for an improvement.

Ctrl-C/KeyboardInterrupt is tricky in Python because it gets special status.

rolandwalker:
A fix was released in https://github.com/dbcli/mycli/releases/tag/v2.11.0, which should make it into Homebrew in a few days.

If you are giving a password to SSH interactively in the terminal, the fix will unfortunately not have an effect.  So it is better to set up the remote with a public key.

rolandwalker:
Homebrew is stuck on v2.9.0 for unrelated reasons: https://github.com/Homebrew/homebrew-core/pull/297441 .

If you are able to test on a `pip`-based install of mycli, let us know.

mentalstring:
I don't really have experience with python, but managed to install 2.11.0 through `pip` and I'm happy to confirm that the issue has been fixed on 2.11.0. 👍 

Thank you for looking into this so promptly, and also for all the work done on `mycli`.

rolandwalker:
Thanks for the confirmation!  Homebrew also got past their blocker and the issue should be fixed there as well now.

PR #2103, opened 2026-08-06
start_new_session when creating SSH tunnels

rolandwalker:
## Description

According to

 * https://docs.python.org/3/library/subprocess.html#popen-constructor

> If start_new_session is true the `setsid()` system call will be made in the child process prior to the execution of the subprocess.

The hope is that `setsid()` keeps the tunnel process from receiving ctrl-c interrupts from the REPL.

Intended to address #2102.

## Checklist

    ```bash
    ```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,11 @@ Features
 * Always clean favorite queries on save and fetch.
 
 
+Bugfixes
+---------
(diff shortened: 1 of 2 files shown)


rolandwalker:
Hm.  This introduces a regression: the user cannot enter a password interactively if prompted by the SSH server.

PR #2140, opened 2026-08-15
Add alternative favorite query interface `/favorite`

rolandwalker:
## Description

`/favorite` duplicates and extends functionality of the existing interface:

 * `/favorite save` -- duplicates `/fs`, including the funky behavior requiring pressing Return twice to save.
 * `/favorite delete` -- duplicates `/fd`.
 * `/favorite list` -- duplicates `/f` alone when favorites are present.
 * `/favorite help` -- duplicates `/f` alone when favorites are not present, but shows a different helpdoc.
 * `/favorite` alone -- same as `/favorite help`.
 * `/favorite run` -- duplicates `/f` when given an argument.
 * `/favorite edit` -- new: edits a favorite in an external editor.
 * `/favorite eval` -- new: evaluates a favorite with parameters, but does not run it.  Instead, the expanded query is placed into the REPL buffer for further editing or execution.
 * `/favorite reload` -- new: re-reads favorite queries from the user's configuration files, without needing to restart mycli.  Edit: renamed from "refresh".

Motivation

 * Have an extensible interface, so we don't have to keep adding `/fe`, `/fr,` and so on for new functionality.  And those could be hard to remember.
 * `/favorite edit` is very helpful with the new Jinja templates, since they can quickly become quite verbose.
 * The many behaviors of `/f` are a bit _too_ clever.  It is nice to have a `help` subcommand which predictably returns the helpdoc.

Drawbacks

 * `/favorite` is longer to type than `/f`, but completions help with that. Completions are also available on each subcommand.

## Checklist

    ```bash
    ```

### DIFF
--- a/mycli/packages/special/favoritequeries.py
+++ b/mycli/packages/special/favoritequeries.py
@@ -1,21 +1,83 @@
 from __future__ import annotations
 
+from collections.abc import Mapping
 import logging
 import os
 import re
-from typing import Any
+from typing import IO, Any
 
+from configobj import ConfigObjError
 from jinja2 import meta, nodes
 from jinja2.sandbox import SandboxedEnvironment
 
-from mycli.config import log, read_config_file
+from mycli.config import log, read_config_file, read_config_files
 
 logger = logging.getLogger(__name__)
 
 MISSING = object()
+FAVORITE_SUBCOMMANDS = ('help', 'list', 'reload', 'run', 'eval', 'save', 'edit', 'delete')
+FAVORITE_COMMAND_HELP = '''
+Favorite Queries are a way to save frequently used queries
+with a short name.
+Examples:
+
+    # Save a new favorite query.
+    > /favorite save simple SELECT * FROM abc WHERE a IS NOT NULL;
+
+    # When multi-line mode is on, pressing Return twice is needed to save.
+    # This supports multi-statement favorites.
+
+    # List all favorite queries.
+    > /favorite list
+    ╒═══════════╤══════════════════════════════════════════════════╕
+    │ Name      │ Query                                            │
+    ╞═══════════╪══════════════════════════════════════════════════╡
+    │ simple    │ SELECT * FROM abc WHERE a IS NOT NULL            │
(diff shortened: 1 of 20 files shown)


mycli/packages/special/favoritequeries.py
```diff
@@ -164,6 +238,45 @@ def from_config(
...
+        user_queries = self._refresh_queries(self.config_file, 'user configuration')
+        system_config = read_config_files(self.system_config_files, ignore_package_defaults=True)
+        system_queries = system_config.get(self.section_name, {})
+        if not isinstance(system_queries, Mapping) or any(
+            not isinstance(name, str) or not isinstance(query, str) for name, query in system_queries.items()
+        ):
+            raise FavoriteQueryRefreshError(f'invalid [{self.section_name}] section in system configuration files')
+        queries: dict[str, str] = {}
+        if self.shared_favorites_file is not None:
+            queries.update(self._refresh_queries(self.shared_favorites_file, 'shared favorites'))
```
scottnemes: Think this needs a try/except because if the file path given does not exist / cannot be read, it will error out and prevent refreshing all the queries:

```python
if self.shared_favorites_file is not None:
    try:
        queries.update(self._refresh_queries(self.shared_favorites_file, 'shared favorites'))
    except FavoriteQueryRefreshError as exc:
        log(logger, logging.WARNING, str(exc))
```

scottnemes (approved):
Left one review comment; good after that!

mycli/packages/special/favoritequeries.py
rolandwalker: Thanks for the review as always!

I am merging without this suggestion because it looks like `_reload_queries()`, which we call here, already handles all the cases with `try` blocks.  Also I tried deleting the shared file out from under mycli and got a graceful failure like:

```
mysql> /favorite reload
Error: Unable to reload favorite queries: unable to read shared favorites file '/full/path/to/favorite.ini'.
```

(Note I renamed the "refresh" subcommand to "reload".)

Let me know if this merge is a mistake and we need a followup.

PR #2142, opened 2026-08-16
fix favorite query commands documented with a forward slash

### LINKED ISSUE #2141: Docs List Favorite Query Commands as `/fx` but They're Really `\fx`
Hi, new to mycli and was trying to create a favorite query based on [the docs](https://www.mycli.net/favorites). From within the session I was doing `/fs simple select * from abc where a is not null` but this kept given me an SQL syntax error even though there was none. Turns out the way to make an alias from within a session is not `/fs`, it's `\fs`. 

Am I missing something? Why are the docs using forward slash and not backslash?

chjnett:
Closes #2141.

The favorite query commands are registered as `\fs` / `\f` / `\fd` (backslash), but the
README and the `\fs`/`\f`/`\fd` help text showed them with a forward slash (`/fs`, `/f`,
`/fd`), so a user following the docs gets an SQL syntax error.

Changes:

- `README.md`: `/fs` → `\fs`, `/f` → `\f`.
- `mycli/packages/special/favoritequeries.py`: correct the `usage` examples
  (`/fs`, `/f`, `/fd` → `\fs`, `\f`, `\fd`).
- `changelog.md`: add a Documentation entry.

Verified against the actual registrations in `mycli/packages/special/iocommands.py`
(`"\\f"`, `"\\fs"`, `"\\fd"`). I left the `@special_command` usage strings there
untouched on purpose — they render the `/` prefix project-wide for every special
command, so changing just the favorite ones would be inconsistent. Say the word if
you'd rather normalize those too.

Docs-only, so no tests changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Features
     - `SELECT * FROM users WHERE <tab>` will only show column names.
 * Support for multiline queries.
 * Favorite queries with positional or named parameters using [Jinja](https://jinja.palletsprojects.com/en/stable/). Save a query using
-  `/fs <alias> <query>` and execute it with `/f <alias>` or `/f <alias> --key=value`.
+  `\fs <alias> <query>` and execute it with `\f <alias>` or `\f <alias> --key=value`.
 * Timing of sql statements and table rendering.
 * Log every query and its results to a file (disabled by default).
 * Pretty print tabular data (with colors!).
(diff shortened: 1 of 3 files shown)


chjnett:
You're right, my mistake — I saw `\fs`/`\f`/`\fd` registered in `iocommands.py` and assumed backslash was canonical, without checking the 2.0.0 changelog that deliberately switched to advertising the forward-slash forms. The docs were correct as-is, so this PR is wrong. Closing it — sorry for the noise.

PR #2158, opened 2026-08-22
Sort completion candidates by frecency from history

### LINKED ISSUE #351: Feature Request: Sort tab completion by frecency
Not sure how easy this would be, but if sorting by frecency is possible then that'd really kick ass.

There are tab completion terms that always show up on top that I never use, and others that are in the middle that I always use. Having the tool remember what terms were most frecently used would improve user experience for databases with may schemas, tables, and columns. Off the top of my head, using an sqlite db in the users home dir could be a great way to store these terms. I haven't looked into any python frecency libraries, but a quick google search shows a few that look usable.

rolandwalker:
## Description

 * calculate frecency for tokens in recent history, simply tokenizing and ignoring literals
 * recalculate frecency in a background thread every N queries at the REPL
 * sort completion candidates by frecency

A good followup could be making the sort-tuple returned by `completion_sort_key()` configurable.

Fixes #351.

## Checklist

    ```bash
    ```

### DIFF
--- a/mycli/main_modes/repl.py
+++ b/mycli/main_modes/repl.py
@@ -74,7 +74,7 @@ from mycli.packages.polars_transform import (
     prepare_polars_transform,
     run_polars_transform,
 )
-from mycli.packages.ptoolkit.history import FileHistoryWithTimestamp
+from mycli.packages.ptoolkit.history import FRECENCY_HISTORY_ENTRIES, FRECENCY_REFRESH_INTERVAL, FileHistoryWithTimestamp
 from mycli.packages.special.utils import format_uptime, get_ssl_version, get_uptime, get_warning_count
 from mycli.packages.sql_utils import (
     extract_new_password,
@@ -146,7 +146,13 @@ def complete_while_typing_filter() -> bool:
 def _create_history(mycli: 'MyCli') -> FileHistoryWithTimestamp | None:
     history_file = os.path.expanduser(os.environ.get('MYCLI_HISTFILE', mycli.config['main'].get('history_file', '~/.mycli-history')))
     if dir_path_exists(history_file):
-        return FileHistoryWithTimestamp(history_file)
+        frecency_history_entries = int(mycli.config['main'].get('frecency_history_entries', FRECENCY_HISTORY_ENTRIES) or 0)
+        frecency_refresh_interval = int(mycli.config['main'].get('frecency_refresh_interval', FRECENCY_REFRESH_INTERVAL) or 0)
+        return FileHistoryWithTimestamp(
+            history_file,
+            frecency_history_entries=frecency_history_entries,
+            frecency_refresh_interval=frecency_refresh_interval,
+        )
 
     mycli.echo(
         f'Error: Unable to open the history file "{history_file}". Your query history will not be saved.',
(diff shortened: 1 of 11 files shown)


mycli/main_modes/repl.py
```diff
@@ -146,7 +146,14 @@ def complete_while_typing_filter() -> bool:
 def _create_history(mycli: 'MyCli') -> FileHistoryWithTimestamp | None:
     history_file = os.path.expanduser(os.environ.get('MYCLI_HISTFILE', mycli.config['main'].get('history_file', '~/.mycli-history')))
     if dir_path_exists(history_file):
-        return FileHistoryWithTimestamp(history_file)
+        configured_history_entries = mycli.config['main'].get('frecency_history_entries', FRECENCY_HISTORY_ENTRIES)
+        frecency_history_entries = int(configured_history_entries) if str(configured_history_entries).strip() else 0
+        frecency_refresh_interval = int(mycli.config['main'].get('frecency_refresh_interval', FRECENCY_REFRESH_INTERVAL))
```
scottnemes: Should this have similar if/else logic as `frecency_history_entries` above?

mycli/packages/ptoolkit/history.py
```diff
@@ -1,28 +1,152 @@
...
+            tokens = tokenize(entry, dialect='mysql')
+        except TokenError:
+            continue
+        for token in tokens:
+            if normalized := _normalize_frecency_token(token):
+                frecency[normalized] += weight
+    return dict(frecency)
+
+
+def frecency_score(text: str, frecency: Mapping[str, float]) -> float:
```
scottnemes: Potential for caching/memoization if you want to reduce load on larger schemas:

```
(mycli) arthlo@CAMELOT-2025:~/git/official-mycli$ git diff
diff --git a/mycli/packages/ptoolkit/history.py b/mycli/packages/ptoolkit/history.py
index ba14d38..78b2ba0 100644
--- a/mycli/packages/ptoolkit/history.py
+++ b/mycli/packages/ptoolkit/history.py
@@ -1,5 +1,6 @@
 from collections import defaultdict
 from collections.abc import Iterable, Mapping
+from functools import lru_cache
 from itertools import islice
 import logging
 import os
@@ -56,12 +57,18 @@ def _calculate_frecency(
     return dict(frecency)


-def frecency_score(text: str, frecency: Mapping[str, float]) -> float:
+@lru_cache(maxsize=8192)
+def _frecency_tokens(text: str) -> tuple[str, ...]:
+    """Tokenize a completion candidate. Cached: candidates repeat on every keystroke."""
     try:
         tokens = tokenize(text, dialect='mysql')
     except TokenError:
-        return 0.0
-    normalized_tokens = [normalized for token in tokens if (normalized := _normalize_frecency_token(token))]
+        return ()
+    return tuple(normalized for token in tokens if (normalized := _normalize_frecency_token(token)))
+
+
+def frecency_score(text: str, frecency: Mapping[str, float]) -> float:
+    normalized_tokens = _frecency_tokens(text)
     if not normalized_tokens:
         return 0.0
     return sum(frecency.get(token, 0.0) for token in normalized_tokens) / len(normalized_tokens)
```

scottnemes (approved):
Added two comments for potential improvements if desired

mycli/main_modes/repl.py
rolandwalker: Good point!  I made them the same, but without the if/else.

mycli/packages/ptoolkit/history.py
rolandwalker: Thanks, made the change.

PR #2184, opened 2026-08-31
WIP: trailing dot on database-name completions

rolandwalker:
## Description
Distinguish database names from table names in completion menus by a trailing dot on database names (aka schema names).

The dot is informational only, and is not inserted when the completion candidate is selected.

One reason this is a WIP is that the dot still appears on a `use` statement.  It should only appear when database names are mixed with table names.

xref: #1876 

Example:

last image

## Checklist

- [ ] To lint and format the code, I ran
    ```bash
    ```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,11 @@
+Upcoming (TBD)
+==============
+
+Features
+--------
+* Show a trailing dot on database names in completion menus.
+
+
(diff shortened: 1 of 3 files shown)


rolandwalker:
Closing without further work.  The dot feels like visual clutter.

PR #2192, opened 2026-09-02
Upgrade WSL OS to Ubuntu 26.04 in CI

rolandwalker:
## Description
Upgrade WSL OS to Ubuntu 26.04 in CI, from Ubuntu 24.04.

Motivation: more users are probably on the latest OS.

## Checklist

    ```bash
    ```


rolandwalker:
It turns out 26.04 is not supported yet per

* https://github.com/Vampire/setup-wsl#inputs

PR #2208, opened 2026-09-06
fix zsh completion test harness when PREFIX is exported

chrisburr:
## Description

When packaging for conda-forge `$PREFIX` is set as an environment variable. This ends up causing `test_zsh_completion_lists_dsn_aliases_in_supported_contexts` to fail due ot zsh's `typeset NAME` printing `NAME=value` when the parameter already exists, so with PREFIX exported the harness writes an extra line to stdout and the captured completion output no longer matches.

Setting it to `''` avoids this issue and makes things consistent.

## Checklist

    ```bash
    ```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,11 @@ Bug Fixes
 * Raise Boundary tunnel stabilization pause to 0.2 sec.
 
 
+Internal
+---------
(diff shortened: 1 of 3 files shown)


rolandwalker:
Thank you @chrisburr !

